PR #24, opened 2017-12-14
Adding VCF output capability; plus filtering to non-ref variants only after subsetting samples

gtiao:
VCF output capability has been requested from collaborators and is necessary for Mike to run the final step in the seqr upload pipeline to make the samples available in the UI. The pipeline currently needs to match sample IDs in the VCF with what's in seqr; and after remapping sample IDs in the VDS (and, by extension, the elasticsearch database) we need to provide that script with a VCF with the remapped sample IDs.

Filtering to non-ref variants will reduce the elasticsearch table sizes.

### DIFF
--- a/entire_vds_pipeline.py
+++ b/entire_vds_pipeline.py
@@ -488,7 +488,7 @@ if not args.skip_vep:
 
     input_path = vep_output_vds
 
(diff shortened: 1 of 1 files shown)


bw2:
Thanks. Looks great.

PR #39, opened 2018-07-10
Use project argument in run script scripts

nawatts:
Currently, run_script.py, run_script_GRCh37.py, and run_script_GRCh38.py all accept a project argument but ignore it and pass "seqr-project" as the project argument to their create cluster and submit scripts.


bw2:
Nice, thanks.

PR #42, opened 2018-08-02
Add Hail 0.2 version of computed fields utils and ClinVar script

nawatts:


hail_v02_scripts/utils/elasticsearch/client.py
```diff
@@ -0,0 +1,253 @@
...
+            export_globals_to_index_meta (bool): whether to add table.globals object to the index _meta field:
+                (see https://www.elastic.co/guide/en/elasticsearch/reference/current/mapping-meta-field.html)
+            verbose (bool): whether to print schema and stats
+        """
+
+        # see https://www.elastic.co/guide/en/elasticsearch/hadoop/master/spark.html#spark-sql-write
+        # "By default, elasticsearch-hadoop will ignore null values in favor of not writing any field at all.
+        # Since a DataFrame is meant to be treated as structured tabular data, you can enable writing nulls as null
+        # valued fields for DataFrame Objects only by toggling the es.spark.dataframe.write.null setting to true."
+        elasticsearch_config = {"es.spark.dataframe.write.null": "true"}
```
nawatts: Not sure if this should be on by default, configurable, or we should just ignore it. I added it here to be consistent with the Hail 0.1 scripts.

https://www.elastic.co/guide/en/elasticsearch/hadoop/master/spark.html#spark-sql-write
> By default, elasticsearch-hadoop will ignore null values in favor of not writing any field at all. Since a DataFrame is meant to be treated as structured tabular data, you can enable writing nulls as null valued fields for DataFrame Objects only by toggling the es.spark.dataframe.write.null setting to true.

For example, for `{foo: 'value', bar: null}`, Hail 0.1 would write `{foo: 'value', bar: null}` to Elasticsearch. Without this configuration setting, Hail 0.2 writes `{foo: 'value'}`.

hail_v02_scripts/utils/elasticsearch/client.py
```diff
@@ -0,0 +1,253 @@
...
+            host (str): Elasticsearch server host
+            port (str): Elasticsearch server port
+        """
+
+        self._host = host
+        self._port = port
+
+        self.es = elasticsearch.Elasticsearch(host, port=port)
+
+    def create_or_update_mapping(self, index_name, index_type_name, elasticsearch_schema={}, num_shards=1, _meta=None):
```
nawatts: Some code, like this method and the functions to encode/decode field names are duplicated between here and `hail_scripts`. We could organize `hail_scripts` into something like:
```
hail_scripts/
  v01/
  v02/
  shared/
```
if we wanted to avoid the duplication.

hail_v02_scripts/utils/elasticsearch/client.py
```diff
@@ -0,0 +1,253 @@
...
+            index_name, index_type_name, elasticsearch_schema, num_shards=num_shards, _meta=_meta
+        )
+
+        logger.info(
+            "==> exporting data to elasticsearch. Write mode: %s, blocksize: %s"
+            % (elasticsearch_write_operation, block_size)
+        )
+
+        # FIXME: https://github.com/hail-is/hail/issues/4063
+        elasticsearch_config = None
```
nawatts: For now, the following arguments won't work because they use this configuration.
* elasticsearch_write_operation
* ignore_elasticsearch_write_errors
* elasticsearch_mapping_id
* disable_doc_values_for_fields
* disable_index_for_fields

hail_v02_scripts/load_clinvar_to_es.py
```diff
@@ -0,0 +1,106 @@
...
+    )
+)
+
+review_status_str = hl.delimit(hl.sorted(hl.array(hl.set(mt.info.CLNREVSTAT)), key=lambda s: s.replace("^_", "z")))
+
+mt = mt.select_rows(
+    allele_id=mt.info.ALLELEID,
+    alt=get_expr_for_alt_allele(mt),
+    chrom=get_expr_for_contig(mt),
+    clinical_significance=hl.delimit(hl.set(mt.info.CLNSIG)),
```
nawatts: Should these be sorted before being joined into a string? Or should they be joined at all?

hail_v02_scripts/utils/elasticsearch/client.py
nawatts: This Hail bug has been fixed, so this can be removed now.

hail_v02_scripts/utils/elasticsearch/client.py
nawatts: 👍. If that's the way we want to go, I can submit a separate PR to reorganize the existing code before adding this new code. There's quite a few path names and imports to change 😭.

bw2: Whatever you guys decide to do with @mattsolo1 ..

hail_v02_scripts/utils/elasticsearch/client.py
bw2: I know it wouldn't delete the fields from the mapping (since ES only lets you add fields to an existing index, but not remove). My guess is it wouldn't touch the values either.

nawatts:
Rebased after #44 was merged and addressed @bw2's comments.

PR #47, opened 2018-08-13
Load clinvar locally

bw2:


hail_scripts/v01/utils/add_clinvar.py
```diff
@@ -126,6 +126,11 @@ def download_and_import_latest_clinvar_vcf(hail_context, genome_version):
     # import vcf
     vds = hail_context.import_vcf(clinvar_vcf_hdfs_path, force_bgz=True, min_partitions=10000, drop_samples=True) #.filter_intervals(hail.Interval.parse("1-MT"))
 
+    if subset:
+        import hail
```
nawatts: Minor, but I think it's preferable to put all imports at the top unless there's something like a circular dependency preventing it.

PR #57, opened 2018-09-05
convert Array[Stuct{..}] to nested object in ES mapping

bw2:


hail_scripts/v01/utils/elasticsearch_utils.py
```diff
@@ -174,17 +186,25 @@ def parse_vds_schema(vds_variant_schema_fields, current_parent=()):
...
+                nested_schema = parse_vds_schema(field.typ.element_type.fields, [])
+                field_path_to_field_type_map[tuple(current_parent + [field_name])] = nested_schema
+            elif field_type.startswith("Struct"):
+                struct_schema = parse_vds_schema(field.typ.fields, current_parent + [field_name])
+                field_path_to_field_type_map.update(struct_schema)
+            else:
+                field_path_to_field_type_map[tuple(current_parent + [field_name])] = field_type
+        except Exception as e:
+            traceback.print_exc()
+            raise ValueError("Error processing field: %s[%s]: %s" % (field_name, field_type, e))
```
nawatts: Maybe include `current_parent` with `field_name` in case a nested field has the same name as a top-level field.

hail_scripts/v01/utils/elasticsearch_utils.py
```diff
@@ -155,7 +167,7 @@ def generate_vds_make_table_arg(field_path_to_field_type_map, is_split_vds=True)
         # drop the 'v', 'va' root from key-table key names
         key = _field_path_to_elasticsearch_field_name(field_path)
         expr = "%s = %s" % (key, ".".join(field_path))
-        if is_split_vds and field_type.startswith("Array"):
+        if is_split_vds and isinstance(field_type, basestring) and field_type.startswith("Array"):
```
nawatts: Out of curiosity, why is the `isinstance(field_type, basestring)` check needed now?

bw2: for nested fields, field_type is dict which breaks field_type.startswith("Array")

nawatts: Doh, I should have realized from the `if isinstance(field_type, dict)` above.

The docstring for the `field_path_to_field_type_map` argument claims that values are strings.

bw2: oops, fixed

PR #59, opened 2018-09-05
Include SIFT and PolyPhen annotations in sortedTranscriptConsequences

nawatts:
The gnomAD browser's variant page uses SIFT and PolyPhen annotations and thus we need to store them in Elasticsearch when loading gnomAD data.

https://github.com/macarthur-lab/gnomad_browser/blob/master/templates/variant_details.html

Closes https://github.com/macarthur-lab/gnomadjs/issues/173

### DIFF
--- a/hail_scripts/v02/utils/computed_fields/vep.py
+++ b/hail_scripts/v02/utils/computed_fields/vep.py
@@ -125,7 +125,18 @@ def get_expr_for_vep_sorted_transcript_consequences_array(vep_root, include_codi
     ]
 
     if include_coding_annotations:
-        selected_annotations.extend(["amino_acids", "lof", "lof_filter", "lof_flags", "lof_info", "protein_id"])
+        selected_annotations.extend(
+            [
+                "amino_acids",
+                "lof",
+                "lof_filter",
+                "lof_flags",
+                "lof_info",
+                "polyphen_prediction",
+                "protein_id",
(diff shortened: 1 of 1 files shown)


hail_scripts/v02/utils/computed_fields/vep.py
```diff
@@ -125,7 +125,20 @@ def get_expr_for_vep_sorted_transcript_consequences_array(vep_root, include_codi
     ]
 
     if include_coding_annotations:
-        selected_annotations.extend(["amino_acids", "lof", "lof_filter", "lof_flags", "lof_info", "protein_id"])
+        selected_annotations.extend(
+            [
```
bw2: are you sure you want both pred. and score fields? seqr currently only shows the predictions:

image

nawatts: You're right, it is only the prediction fields that are needed.

bw2: The prediction values are also a little weird. IIRC they have ;-separated values that are repeated multiple times (for different proteins off of the same transcript). 
You might want to pre-process them to a single value per transcript.

nawatts: I checked the predictions on ClinVar variants and didn't see anything that looked like it had multiple values concatenated into one string.

PR #64, opened 2018-09-13
Adding new LoFTEE version, but modified to remove features that use ERP and PhyloCSF since these are only available on GRCh37

bw2:


bw2:
Putting LoFTEE code here temporarily until @konradjk has an official version that supports GRCh38.

bw2:
Merging to proceed with build steps that depend on this. Feedback appreciated post-merge.

PR #67, opened 2018-09-17
Misc improvements

bw2:


hail_scripts/v01/utils/shell_utils.py
```diff
@@ -72,13 +74,28 @@ def run(command,
...
+    previous_is_slash_r = False
+    while True:
+        out = p.stdout.read(1)
+        if out == '' and p.poll() is not None:
+            break
+        if out != '':
+            log_buffer.write(out)
+            if verbose:
+                line_buffer.write(out)
+                if out.endswith('\r') or (out.endswith('\n') and previous_is_slash_r):
```
nawatts: The `universal_newlines` argument to `subprocess.Popen` might simplify this.

https://docs.python.org/2/library/subprocess.html#frequently-used-arguments

.travis.yml
```diff
@@ -0,0 +1,15 @@
...
+
+install:
+- pip install -r requirements.txt
+
+script:
+- python -m unittest discover -p '*test*.py'
+
+#notifications:
+#  slack:
+#    secure: tYNwi4kff+8FWbv+nWqKZt28Qez0Dv77rJG6nvdiaO7QgPZmaiG5l8NqYdxV63CMgNzHwTFAz8ca9uOoh06o7+HuiFNNiqXEAu8umLvlnvkKT6itSDyImxxD7ypitUwcBEKkNy1fxhrLRPQVLjxz37885kA/VtWpm19A5vQ8A7w=
```
nawatts: Was this key intended to be committed?

.travis.yml
```diff
@@ -0,0 +1,15 @@
+language: python
+
+python:
+- '2.7'
```
nawatts: Hail 0.2 runs in Python 3.6. But for now, there are no tests for Hail 0.2 code.

hail_scripts/v01/utils/shell_utils.py
bw2: I tried using universal_newlines but it didn't work correctly for either progress bars like tqdm or  regular line text in some cases.

.travis.yml
bw2: yeah, this is encrypted based on https://docs.travis-ci.com/user/encryption-keys/

PR #70, opened 2018-09-24
Iteration1 load into temp elasticsearch cluster

bw2:
- combines load_GRCh37_dataset.py and load_GRCh38_dataset.py into one load_dataset.py script with a --genome-version arg
- adds a new `--use-temp-es-cluster` option. Instead of having the hail pipeline load data into an existing elasticsearch cluster, it creates temporary data nodes with local SSDs, then uses elasticsearch shard allocation filtering to make the pipeline load data only into those nodes. When loading completes, it transfers the shards to the original persistent nodes and deletes the temp nodes. 
- move some of the ./servctl deployment utilities from seqr into this repo and use them to make the k8s elasticsearch configs into Jinja templates so parameters like number of data nodes can be set on the command line. Make this repo a submodule of the seqr repo, and then have seqr use the sharded elasticsearch cluster configs in this repo to avoid duplication. 

Remaining TODOs:
- figure out a way to allow 2 or more load_dataset.py jobs to run in parallel.


gcloud_dataproc/utils/seqr_api.py
```diff
@@ -1,94 +1,59 @@
...
-        data = {
-            'username_or_email': seqr_username,
-            'password': seqr_password,
-            'csrfmiddlewaretoken': csrf_token,
-        }
-        response = s.post(seqr_login_url, data=data)    # submit login
+    response.raise_for_status()
 
-        response.raise_for_status()
+    response = s.get(seqr_individual_api_url.format(**locals()))   # get json from API
```
hanars: `seqr_individual_api_url` has already had `format` called on it

gcloud_dataproc/utils/seqr_api.py
```diff
@@ -1,94 +1,59 @@
...
+        if not line or line.startswith("#"):
+            continue
+        if not fam_file_rows and "affected" in line.lower() and "indiv" in line.lower() and "family" in line.lower():
+            continue  # skip header row
 
-        # upload subset-samples to same place as VCF
-        if not is_subset_samples_file_specified:
-            with open(subset_samples_filename, "w") as f:
-                f.write("\n".join([row for row in subset_samples_rows]))
+        fields = line.split("\t")
```
hanars: I think this would be more resilient and easier to read if instead of using a file download we use the `/api/project/<project_guid>/details` endpoint and then get the data from the `individualsByGuid` key of the response. That way we are actually getting the fields we want explicitly instead of relying on the columns of this download to never change

bw2: The current version isn't ideal but the tsv is slightly less brittle than it looks because it's following .fam file format. I agree json is much better, but I'd vote for staying with the current api and just adding a format=json option - that way it's only retrieving individuals and isn't affected by changes to the project page.

PR #71, opened 2018-09-25
Add helpers to precompute browser LoF flags

nawatts:
To avoid having to retrieve all transcript consequences from Elasticsearch, precompute flags which are shown in the browser.

Matching this behavior from gnomad_browser:
https://github.com/macarthur-lab/gnomad_browser/blob/b4ff2705fd540037acc31f234f660b26162f5207/utils.py#L95-L109


nawatts:
There's some additional code in gnomad_browser that impacts the value of these flags.

https://github.com/macarthur-lab/gnomad_browser/blob/b4ff2705fd540037acc31f234f660b26162f5207/utils.py#L79-L84

PR #77, opened 2018-10-28
Store genotypes as child docs

bw2:


hail_scripts/v01/utils/elasticsearch_client.py
```diff
@@ -270,12 +294,24 @@ def export_kt_to_elasticsearch(
...
-        elasticsearch_schema = generate_elasticsearch_schema(
+        index_schema = generate_elasticsearch_schema(
             field_path_to_field_type_map,
             field_name_to_elasticsearch_type_map=field_name_to_elasticsearch_type_map,
             disable_doc_values_for_fields=disable_doc_values_for_fields,
             disable_index_for_fields=disable_index_for_fields)
 
+        if child_kt is not None:
+            # see https://www.elastic.co/guide/en/elasticsearch/reference/current/parent-join.html
+            index_schema["join_field"] = {  # since an index can only have 1 "join" field, it's ok to use generic names.
```
hanars: It is okay to use generic names, but why not use a descriptive name since we know what this field represents?

bw2: `export_kt_to_elasticsearch(..)` is meant to be generic to tables and doesn't assume them to be variants or genotypes.

PR #82, opened 2018-11-19
converting seqr reference data pipeline to hail v0.2 

bw2:
- added reference data loading scripts 
- rearranged files


download_and_create_reference_datasets/v02/create_ht__cadd.py
```diff
@@ -0,0 +1,8 @@
+#!/usr/bin/env python
```
nawatts: This is inconsistent with the other v02 scripts, which use python3.

hail_scripts/v02/utils/computed_fields/variant_id.py
```diff
@@ -20,14 +20,20 @@ def get_expr_for_contig_number(table):
     )
 
 
-def get_expr_for_original_alt_alleles_set(split_variant):
-    """Compute variant id for each original alt allele in a variant split with hl.split_multi"""
-    return hl.set(
-        split_variant.old_alleles[1:].map(
-            lambda alt: get_expr_for_variant_id(hl.min_rep(split_variant.old_locus, [split_variant.old_alleles[0], alt]))
-        )
-    )
```
nawatts: I believe this breaks gnomAD's pipeline for loading ExAC.

download_and_create_reference_datasets/v02/create_ht__cadd.py
bw2: oops

hail_scripts/v02/utils/computed_fields/variant_id.py
nawatts: I had been thinking about changing these functions to accept individual field expressions rather than the full table, since passing the table requires a specific table structure. I like `get_expr_for_variant_ids` more than `get_expr_for_orig_alt_alleles` since it's more general. If `get_expr_for_variant_ids` took a locus and alleles argument, then the ExAC pipeline could pass old_locus and old_alleles.

hail_scripts/v02/utils/elasticsearch_client.py
```diff
@@ -79,8 +80,10 @@ def export_table_to_elasticsearch(
                 this string in all field names. This replacement is not reversible (or atleast not
                 unambiguously in the general case) Set this to None to disable replacement, and fall back
                 on an encoding that's uglier, but reversible (eg. "." will be converted to "_$dot$_")
+            func_to_run_after_index_exists (function): optional function to run after creating the index, but before exporting any data.
```
nawatts: Would it be better to split out the index creation into another function? `export_table_to_elasticsearch` could by default call it before exporting any data, but in the case where you need to do something in between, you could call them separately.

hail_scripts/v02/utils/hail_utils.py
```diff
@@ -0,0 +1,123 @@
...
+
+    # ensure that va.wasSplit and va.aIndex are defined before calling split_multi() since split_multi() doesn't define these if all variants are bi-allelic
+    mt = mt.annotate_rows(was_split=False, a_index=1, original_alt_alleles=get_expr_for_orig_alt_alleles(mt))
+    mt = hl.split_multi_hts(mt)
+    mt = mt.key_rows_by(**hl.min_rep(mt.locus, mt.alleles))
+
+    return mt
+
+
+def read_mt(mt_path: str):
```
nawatts: To me, these functions that wrap one Hail function obfuscate the code. Someone familiar with Hail knows what `hl.read_matrix_table` does, but they likely have to look up what `hail_scripts/v02/utils/hail_utils.read_mt` does.

hail_scripts/v02/utils/hail_utils.py
```diff
@@ -0,0 +1,123 @@
...
+        reference_genome=f"GRCh{genome_version}",
+        contig_recoding=contig_recoding,
+        min_partitions=min_partitions,
+        force_bgz=force_bgz,
+        drop_samples=drop_samples,
+        skip_invalid_loci=skip_invalid_loci)
+
+    mt = mt.annotate_globals(sourceFilePath=vcf_path, genomeVersion=genome_version)
+
+    # ensure that va.wasSplit and va.aIndex are defined before calling split_multi() since split_multi() doesn't define these if all variants are bi-allelic
```
nawatts: Is this still the case with Hail 0.2?

hail_scripts/v02/utils/hail_utils.py
```diff
@@ -0,0 +1,123 @@
+import hail as hl
```
nawatts: I think this module could use a better name. `hail_utils` nested in `hail_scripts.utils` is a bit redundant. Also maybe splitting up... `run_vep` is definitely unlike the other functions here, which are all import/export related.

download_and_create_reference_datasets/v02/create_ht__cadd.py
bw2: fixed in  https://github.com/macarthur-lab/hail-elasticsearch-pipelines/commit/08e22d61390457c02929d8372cf80eebf041bc3d

hail_scripts/v02/utils/computed_fields/variant_id.py
bw2: I like that idea.

hail_scripts/v02/utils/elasticsearch_client.py
bw2: I don't have a strong opinion on this. These methods did get too big in v01. 
I need to make one more change to the v01 method now (to allow parent/child relations to have custom names). Beyond that though, if there's a particular way you want these methods (as well as the computed_fields ones ) to be structured, do you think you could port them (while I finish up the reference data scripts)?

hail_scripts/v02/utils/hail_utils.py
bw2: I added the import_vcf wrapper to do the several things (like split_multi) that I always do after importing, and then added the other ones for completeness. I see your point though and am fine with dropping these.

download_and_create_reference_datasets/v01/create_vds__cadd.py
```diff
@@ -0,0 +1,9 @@
+#!/usr/bin/env python
+
+from kubernetes.shell_utils import simple_run as run
```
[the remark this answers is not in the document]
bw2: agreed. This module is imported by kubectl_utils, but also other things. Also, code in this directory is also used by modules in the seqr repo. What do you think about moving `shell_utils.py` and `network_utils.py` to the top level dir, or to some top-level `shared_utils` dir, or renaming the `kubernetes` dir to something else?

hail_scripts/v02/utils/hail_utils.py
bw2: sure .. happy to go with your preferences on the dir structure.

hail_scripts/v02/utils/hail_utils.py
bw2: good q. I need to double check.

download_and_create_reference_datasets/v01/create_vds__cadd.py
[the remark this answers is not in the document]
nawatts: +1 for moving those two files to the top level.

hail_scripts/v02/utils/hail_utils.py
```diff
@@ -0,0 +1,123 @@
...
+    :rtype: MatrixTable
+    """
+    if genome_version == "37":
+        mt = mt.annotate_global_expr('global.gencodeVersion = "19"')  # see
+    elif genome_version == "38":
+        mt = mt.annotate_global_expr('global.gencodeVersion = "25"')  # see gs://hail-common/vep/vep/homo_sapiens/85_GRCh38/info.txt
+    else:
+        raise  ValueError(f"Invalid genome version: {genome_version}")
+
+    mt = hl.methods.vep(mt, config="/vep/vep-gcloud-grch{}.properties".format(genome_version), name= name, block_size=block_size)
```
nawatts: VEP now requires a JSON configuration file. https://hail.is/docs/0.2/methods/genetics.html#hail.methods.vep

cloudtools uses this init script to setup VEP: gs://hail-common/vep/vep/vep85-loftee-init-docker.sh

PR #83, opened 2018-11-19
Update ref data in existing indices

bw2:


hail_scripts/v01/utils/add_clinvar.py
```diff
@@ -50,17 +50,24 @@
 )"""
 
 
+CLINVAR_VDS_CACHE = { "37": None, "38": None }
+
```
hanars: I think the keys in this are tuples once its loaded so its a bit confusing that its initialized with these string keys, I think it would be better just tp start with an empty dict

hail_scripts/v01/utils/add_hgmd.py
```diff
@@ -1,19 +1,46 @@
...
+HGMD_PATH_GRCh38 = 'gs://seqr-reference-data-private/GRCh38/HGMD/HGMD_PRO_2018.1_hg38_without_DB_field.vds'
+
+def reset_hgmd_fields_in_vds(hail_context, vds, genome_version, root="va.hgmd", subset=None, verbose=True):
+    hgmd_vds = read_hgmd_vds(hail_context, genome_version, subset=subset)
+
+    # %(root)s.review_status = NA:String,
+    vds = vds.annotate_variants_vds(hgmd_vds, expr="""
+        %(root)s.accession = NA:String,
+        %(root)s.class = NA:String
+    """ % locals())
```
hanars: In general, using `locals` is really fragile because renaming variables can completely break things and its hard to know where if at all variables are being used. I know its all over this code base but I think we should try not to add any more

bw2: This pattern seems equivalent to python 3.6  f-strings (https://www.python.org/dev/peps/pep-0498/), and it seems to me finding references to vars is ~ the same inside or outside these strings.

hail_scripts/v01/utils/add_clinvar.py
bw2: oops.. good point. I'll just do = {}.

hail_scripts/v01/utils/add_hgmd.py
hanars: I have fixed bugs caused by this multiple times in seqr. Maybe in theory its the same level of difficulty to find references to these variables but in practice this has broken on production

PR #87, opened 2018-11-26
Remove upstream downstream variant annotations

hanars:
Also includes small fixes for child doc genotypes

### DIFF
--- a/hail_scripts/v01/utils/computed_fields.py
+++ b/hail_scripts/v01/utils/computed_fields.py
@@ -26,8 +26,6 @@ CONSEQUENCE_TERMS = [
     "NMD_transcript_variant",
     "non_coding_transcript_variant",
     "nc_transcript_variant",  # deprecated
-    "upstream_gene_variant",
-    "downstream_gene_variant",
     "TFBS_ablation",
     "TFBS_amplification",
     "TF_binding_site_variant",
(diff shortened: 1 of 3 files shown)


nawatts:
In gnomAD, we show these consequences. If this change gets made in the Hail 0.2 version as part of porting Seqr pipelines (#82), then we'd need an argument to make the filter optional.

hanars:
@nawatts that makes sense. I think I would rather leave this without the argument in 0.1 pipeline, since we will probably want to update the arguments in the 0.2 pipeline anyways

hail_scripts/v01/utils/computed_fields.py
```diff
@@ -315,7 +327,7 @@ def get_expr_for_vep_consequence_terms_set(vep_transcript_consequences_root="va.
 def get_expr_for_vep_protein_domains_set(vep_transcript_consequences_root="va.vep.transcript_consequences"):
     return """%(vep_transcript_consequences_root)s
         .filter( x => isDefined(x.domains) && x.domains.length > 0 )
-        .map( x => x.domains.map( domain => domain.db+":"+domain.name ) )
+        .map( x => x.domains )
```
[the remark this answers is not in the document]
hanars: because this is being called on `sortedTranscriptConsequences` instead of the root consequences, and this transformation has already been done on that field

hail_scripts/v01/utils/computed_fields.py
```diff
@@ -26,8 +26,6 @@
     "NMD_transcript_variant",
     "non_coding_transcript_variant",
     "nc_transcript_variant",  # deprecated
-    "upstream_gene_variant",
-    "downstream_gene_variant",
```
nawatts: This breaks https://github.com/macarthur-lab/hail-elasticsearch-pipelines/blob/92700c1b494fd702487a3f52c2bc4a3a9f1d332d/download_and_create_reference_datasets/v01/hail_scripts/write_dataset_validation_kt.py#L71-L72

hanars: Hmm could we use intronic variants instead?

nawatts: Either
```
... major_consequence_rank.toInt > {}.get("nc_transcript_variant").toInt ...
```
or
```
... major_consequence_rank.toInt >= {}.get("TFBS_ablation").toInt ...
```
would preserve the same behavior as before.

But I don't understand why when splitting into coding and non-coding variants, the 11 VEP consequences between synonymous_variant and downstream_gene_variant are omitted.

hanars: I think @bw2 wrote this and might have some insight into that

PR #89, opened 2018-12-03
Fix identification of synonymous variants for formatting hgvs

nawatts:
Currently, for purposes of formatting the hgvs field, synonymous variants are identified based on the presence of a URL-encoded equals sign in their vep.transcript_consequences.hgvsp field.

This change makes the URL-encoding optional, so that the hgvs field is formatted if hgvsp contains "=" or "%3D".


nawatts:
Relevant to macarthur-lab/gnomadjs#225. This worked with ExAC, since those variants are loaded from a VCF file. gnomAD is loaded directly from a Hail table though, so the "=" aren't URL-encoded.

One could argue that this should only accept "=" and the ExAC loading script should decode fields before calling these utilities, but it's nice to have this work for both cases.

PR #90, opened 2018-12-05
Add functions to compute genes where variants should be flagged

nawatts:
Currently, in the gnomAD browser, the "LC LoF" and "LoF Flag" flags are shown for a variant only if **all** the variant's transcript consequences are not HC or flagged, respectively.

@konradjk pointed out that on the gene and transcript pages, these flags should be based only on the variant's consequences in that gene or transcript.
https://macarthurlab.slack.com/archives/C02A348ML/p1542398082164000

Based on a variant's sorted transcript consequences, these functions compute the set of gene IDs where the variant should be flagged.

### DIFF
--- a/hail_scripts/v02/utils/computed_fields/flags.py
+++ b/hail_scripts/v02/utils/computed_fields/flags.py
@@ -9,9 +9,43 @@ def get_expr_for_lc_lof_flag(sortedTranscriptConsequences):
     )
 
 
+def get_expr_for_genes_with_lc_lof_flag(sorted_transcript_consequences):
+    """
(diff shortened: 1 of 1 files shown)


hanars:
It looks like you added a couple functions but don't call them anywhere. is that on purpose?

nawatts:
Yes, it was intentional. They'll be called in the gnomAD browser's variant loading pipeline.

https://github.com/macarthur-lab/gnomadjs/blob/master/projects/gnomad/data/shape_gnomad_r2_1_for_browser.py

PR #93, opened 2019-01-10
Nested docs for sample update

hanars:
Upload genotypes as an object keyed by sample id


hanars:
Turns out we can probably use existing nested gentoypes, so closing this

PR #97, opened 2019-01-22
Remove upstream/downstream gene consequence terms

nawatts:
#87 applied to the Hail 0.2 version of computed fields.

Resolves macarthur-lab/gnomadjs#384.

### DIFF
--- a/hail_scripts/v02/utils/computed_fields/vep.py
+++ b/hail_scripts/v02/utils/computed_fields/vep.py
@@ -151,7 +151,6 @@ def get_expr_for_vep_sorted_transcript_consequences_array(vep_root, include_codi
         "cdna_start",
         "cdna_end",
         "codons",
-        "consequence_terms",
         "gene_id",
         "gene_symbol",
         "hgvsc",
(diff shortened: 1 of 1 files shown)


hanars:
Is there any reason not to remove these consequences from the `CONSEQUENCE_TERMS` list completely? I feel like keeping them there is confusing/ might mess up sorting

nawatts:
I kind of like having them there since then you can use `CONSEQUENCE_TERM_RANK_LOOKUP` on VEP output, not just `sortedTranscriptConsequences`. But I don't know that that is actually useful.

hanars:
Maybe move them to the bottom of the list? I just find it odd that we are ranking them higher than other transcripts but also filtering them out

nawatts:
I don't think we should change the order since it is is defined by Ensembl's estimate of severity. Which is another reason I favor leaving these two terms in the list... that way, we have the complete ranking from https://ensembl.org/info/genome/variation/prediction/predicted_data.html (though I just noticed we're missing `start_retained_variant`).

hanars:
okay thats fine. Maybe put a comment in the code indicating where that ranking comes from then?

PR #103, opened 2019-02-04
Temp loading cluster fixes

bw2:
fixes create-cluster and create-persistent-es-node steps


gcloud_dataproc/load_dataset.py
```diff
@@ -90,16 +103,83 @@ def _process_kubernetes_configs(action, config_paths, settings):
...
+def _make_disks(settings, es_disk_snapshots=None):
+    """
+
+    Args:
+        es_disk_snapshots (list): optional list of snapshot names
+    """
+
+    # create disks from snapshots
+    created_disks = []
+    if es_disk_snapshots:
```
hanars: Why would we ever create disks from snapshots during data loading?

gcloud_dataproc/load_dataset.py
```diff
@@ -290,50 +369,60 @@ def main():
...
-        _enable_cluster_routing_rebalance(False, args.cluster_name, args.host, args.port)
-
         # create temp es nodes
         settings = {
             "DEPLOY_TO": args.k8s_cluster_name,
             "CLUSTER_NAME": args.k8s_cluster_name,
             "ES_CLUSTER_NAME": args.k8s_cluster_name,
             "NAMESPACE": args.k8s_cluster_name,  # kubernetes namespace
             "IMAGE_PULL_POLICY": "Always",
+            #"GCLOUD_ZONE": "central",
```
hanars: why is this commented out?

gcloud_dataproc/load_dataset.py
bw2: This can be used to create a completely new ES cluster that has both persistent and temp nodes, and load existing data (similar to restoring postgres backups)  https://github.com/macarthur-lab/hail-elasticsearch-pipelines/pull/103/files/28fc0b74de306c994ff4581d7c36e9a36a8fe2b8#diff-9eed0c0fd4f99bedc78484994f835391R203

gcloud_dataproc/load_dataset.py
bw2: it's typically already configured in someone's local gcloud config.

hanars: but then why not just remove it?

gcloud_dataproc/load_dataset.py
hanars: I don't think that functionality should be a flag on a script designed to load a dataset, I think thats incredibly confusing. I agree that thats a useful functionality to have, but I think it should be done by a different script which just creates the new cluster and doesn't do any of the other loading pipeline stuff. Also, my intuition is that that script would make more sense in the seqr repo instead of this repo, as I assume the use case for it is mostly related to getting test instances/ local instances for seqr up. But I don't feel so strongly about that

mattsolo1: This seems like it fits into a larger design discussion to be had about the seqr data loading process. Do we want to continue with a small number of complex scripts that are configurable through command line args? Or break it out into a smaller number of scripts with single responsibilities (but could still be composed into something larger)? I'm favor of the later option. Thoughts, @knguyen142?

hanars: I agree that ultimately it makes more sense to move toward an architecture of smaller composeable scripts than the giant monolithic one we have now. That said, I think its out of the scope of getting this branch merged to entirely refactor the giant script we already have. My preference would be that this new functionality get broken out into its own script, but I don't think we should blow up the scope of this by refactoring whats already there

knguyen142: Yeah a bunch of smaller scripts, or one script with a bunch of subcommands and options that can apply to individual or shared subcommands makes sense e.g. gcloud.

bw2: That's a great point, these scripts have gotten way too big.

hanars: To clarify, I want this functionality broken out into its own script before I feel comfortable approving this

bw2: The functionality of creating a cluster with both persistent and temp nodes existed here before this PR. I think it makes sense to split things as part of the general refactoring.

hanars: the es-disk-snapshots flag did not used to exist, that is all entirely new functionality

bw2: true.

hanars:
In you latest commit, did you forget to add the new file?

hanars:
Is there a reason this hasn't been merged?

bw2:
I put the separate script in this PR https://github.com/macarthur-lab/hail-elasticsearch-pipelines/pull/106/files

PR #104, opened 2019-02-05
working step0?

knguyen142:
Part of v02 migration. The strategy is to create a new v02 directory and go step by step and change only functions that make calls to Hail. 

This is the step0 change (except for validating, remapping/subset sampling, and vep). Only functionality that involves the load function for step0 are added.

The goal is to do multiple PRs step by step so the diffs are manageable. 

Most of this code is copied and pasted from their equivalent v01 file (we intentionally want to introduce as little change as possible for this port), so it makes sense to look at the equivalent v01 when reviewing.

Will send info on sanity checks/testing tomorrow.

### DIFF
--- a/hail_scripts/v02/load_dataset_to_es.py
+++ b/hail_scripts/v02/load_dataset_to_es.py
@@ -1,0 +1,298 @@
+#!/usr/bin/env python
+
+import os
+
+# os.system("pip install elasticsearch")
+
+import argparse
+import hail as hl
+import logging
+from pprint import pprint, pformat
+import time
+import sys
+
+from hail_scripts.shared.elasticsearch_utils import ELASTICSEARCH_INDEX, \
+    ELASTICSEARCH_UPDATE, ELASTICSEARCH_UPSERT
+from hail_scripts.v01.utils.add_hgmd import add_hgmd_to_vds
+from hail_scripts.v01.utils.add_eigen import add_eigen_to_vds
+from hail_scripts.v01.utils.gcloud_utils import delete_gcloud_file
+from hail_scripts.v02.utils.vds_utils import read_in_dataset
+from hail_scripts.v01.utils.computed_fields import get_expr_for_variant_id, \
+    get_expr_for_vep_gene_ids_set, get_expr_for_vep_transcript_ids_set, \
+    get_expr_for_vep_consequence_terms_set, \
+    get_expr_for_vep_sorted_transcript_consequences_array, \
+    get_expr_for_worst_transcript_consequence_annotations_struct, get_expr_for_end_pos, \
+    get_expr_for_xpos, get_expr_for_contig, get_expr_for_start_pos, get_expr_for_alt_allele, \
+    get_expr_for_ref_allele, get_expr_for_vep_protein_domains_set, get_expr_for_variant_type, \
+    get_expr_for_filtering_allele_frequency
+from hail_scripts.v01.utils.elasticsearch_utils import VARIANT_GENOTYPE_FIELDS_TO_EXPORT, \
+    VARIANT_GENOTYPE_FIELD_TO_ELASTICSEARCH_TYPE_MAP, \
+    SV_GENOTYPE_FIELDS_TO_EXPORT, SV_GENOTYPE_FIELD_TO_ELASTICSEARCH_TYPE_MAP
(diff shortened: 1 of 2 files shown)


hail_scripts/v02/load_dataset_to_es.py
```diff
@@ -0,0 +1,298 @@
...
+
+    # Store step0_output_vds as the cached version of the dataset in google buckets, and also set it as the global.sourceFilePath
+    # because
+    # 1) vep is the most time-consuming step (other than exporting to elasticsearch), so it makes sense to cache results
+    # 2) at this stage, all subsetting and remapping has already been applied, so the samples in the dataset are only the ones exported to elasticsearch
+    # 3) annotations may be updated / added more often than vep versions.
+    mt = mt.annotate_globals(sourceFilePath = args.step0_output_mt)
+    mt = mt.annotate_globals(genomeVersion = args.genome_version)
+    mt = mt.annotate_globals(sampleType = args.sample_type)
+    mt = mt.annotate_globals(datasetType = args.dataset_type)
```
nawatts: These could all be combined into one call to `annotate_globals`.

hail_scripts/v02/load_dataset_to_es.py
```diff
@@ -0,0 +1,298 @@
...
+    logger.info("Command args: \n" + " ".join(sys.argv[:1]) + ((" --index " + args.index) if "--index" not in sys.argv else ""))
+
+    logger.info("Parsed args: \n" + pformat(args.__dict__))
+
+    return args
+
+def compute_index_name(args):
+    """Returns elasticsearch index name computed based on command-line args"""
+
+    # generate the index name as:  <project>_<WGS_WES>_<family?>_<VARIANTS or SVs>_<YYYYMMDD>_<batch>
```
nawatts: The format in this comment doesn't look accurate.

hail_scripts/v02/load_dataset_to_es.py
```diff
@@ -0,0 +1,298 @@
...
+from hail_scripts.v01.utils.fam_file_utils import MAX_SAMPLES_PER_INDEX, compute_sample_groups_from_fam_file
+from hail_scripts.v01.utils.vds_schema_string_utils import convert_vds_schema_string_to_annotate_variants_expr
+from hail_scripts.v01.utils.add_1kg_phase3 import add_1kg_phase3_to_vds
+from hail_scripts.v01.utils.add_cadd import add_cadd_to_vds
+from hail_scripts.v01.utils.add_dbnsfp import add_dbnsfp_to_vds
+from hail_scripts.v01.utils.add_clinvar import add_clinvar_to_vds
+from hail_scripts.v01.utils.add_exac import add_exac_to_vds
+from hail_scripts.v01.utils.add_gnomad import add_gnomad_to_vds
+from hail_scripts.v01.utils.add_topmed import add_topmed_to_vds
+from hail_scripts.v01.utils.add_mpc import add_mpc_to_vds
```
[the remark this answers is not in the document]
knguyen142: Essentially, yes.

So the context is we decided to just change the minimal possible (only the hail diff), so copied and pasted everything from v01 to this. But instead of doing a PR with the whole file (since we can't see the PR diff changes in a new file),I decided to take everything out of that are the later steps so the PR is more manageable.

I left it here so as I add the other steps I won't have to worry about imports, and then when it's all done I'll remove the unused ones.

tldr; a middle ground between a way of sanity for me so I don't have to look at all the code at once, and a way for PR reviewers to not have to look at all of the non-relevant v01 code at once.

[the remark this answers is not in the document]
knguyen142: Now that I'm thinking about it, it might make sense for me to get a copied and pasted version of v01 to v02 merged to master, and all subsequent diffs on v02 will only see v02 changes against a v01 copy.

knguyen142:
New strategy, will get a base copy of v02 (copied from v01) merged to master, and then do a PR again so we can just view the diffs I made in an easy way.
@nawatts will address your PR comments in new PR.

hail_scripts/v02/utils/vds_utils.py
```diff
@@ -0,0 +1,75 @@
+import hail as hl
```
knguyen142: And why did I name this file vds_utils? Haha.

PR #110, opened 2019-02-15
eigen, 1kg, primate_ai

knguyen142:
Eigen, 1kg, and Primate AI according to @bw2 are simple load VCF and export to table and that's why they don't have a corresponding v01 file. 

What I did was use https://github.com/macarthur-lab/hail-elasticsearch-pipelines/blob/master/download_and_create_reference_datasets/v02/create_ht__mpc.py which is a standard load and dump on the mpc data as template, and copied over the corresponding reference dataset.

# Testing
None since it's so simple and easier to test on the live data. Will need to learn how to do this.

### DIFF
--- a/download_and_create_reference_datasets/v02/create_ht__1kg.py
+++ b/download_and_create_reference_datasets/v02/create_ht__1kg.py
@@ -1,0 +1,14 @@
+#!/usr/bin/env python3
+
+from kubernetes.shell_utils import simple_run as run
+
+for genome_version, vcf_path in [
+    ("37", "gs://seqr-reference-data/GRCh37/1kg/1kg.wgs.phase3.20130502.GRCh37_sites.vds"),
+    ("38", "gs://seqr-reference-data/GRCh38/1kg/1kg.wgs.phase3.20170504.GRCh38_sites.vds"),
+]:
+    run(("python3 gcloud_dataproc/v02/run_script.py "
+        "--cluster create-ht-mpc "
+        "hail_scripts/v02/convert_vcf_to_hail.py "
+        "--output-sites-only-ht "
+        f"--genome-version {genome_version} "
+        f"{vcf_path}"))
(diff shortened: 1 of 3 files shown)


hanars:
This is a lot of nearly identical copy pasted code. I'm wondering if theres a way to make it more general? Maybe having a single `create_reference_data_ht` script that takes the reference data source as an arg?

knguyen142:
Yeah I was thinking of doing that somehow, but it's hard because I don't have the setup to actually run these (which is why I avoided creating equivalent run scripts in my other PRs).

knguyen142:
In fact, I think all of the scripts can be consolidated into one? I assumed we separated it for a reason, but a consolidation can happen in another PR. In general, I think [code boy scouting ](https://deviq.com/boy-scout-rule/)and features should be separate unless small.

hanars:
So as far as I know (might be worth confirming with @bw2) the reason they are separated out is that when we have to update a reference data source, its usually only one source at a time. So we need the ability to update just that data source, and thus there is a script for each one instead of a big global script that does all of them. However, I think one script that has a mandatory argument for which data source to update works fine functionally and is a lot cleaner

hanars:
And in general I agree that mixing features and code clean up is a bad idea. However, since you are writing these files from scratch my instinct would be just to write them cleanly the first time, rather than duplicating badly structured code and then having to rewrite it almost immediately. But if you feel more comfortable with that latter approach, thats your call and I'm happy to approve this as is

knguyen142:
Yeah I think if it's only isolated in the code I'm writing, happy to do it.
But I think the overall run script can be restructured, I just don't know
if it'll cause any problems. I actually don't think setting a global config
is a bad idea.

knguyen142:
But also, if it's code that's reliability, only run once in a blue moon,
and does not need to be maintained, a refactor might be overkill.

hanars:
It is code that is only run once in a while, but it certainly needs to be maintained. Another thing to keep in mind is we add new reference data sources with some regularity and every time we do that, we would be copy-pasting these files and modifying them slightly. A shared script would make that process much easier, and to me would justify a refactor. 

In terms of restructuring this causing problems, these scripts aren't run regularly and I think have only ever been run by Ben, so its not like a change to that would surprise/ confuse anyone. These will all need to be run before the rest of the v2 pipeline can be run, and my guess is you will be the one to do that the first time, so its really up to you which you would prefer

PR #112, opened 2019-02-27
crash submit.py if `gcloud dataproc jobs submit` exit code != 0

bw2:
Quick fix for delete-temp-nodes issue:

`gcloud dataproc jobs submit` returns a non-zero exit code if the dataproc pipeline crashes. 
This fix then also crashes  `submit.py` so that wrapper scripts (like `load_dataset.py`) can exit without running cleanup steps like delete-temp-loading-nodes and leaving intermediate data for reruns.

### DIFF
--- a/gcloud_dataproc/submit.py
+++ b/gcloud_dataproc/submit.py
@@ -95,4 +95,7 @@ else:
     """ % locals()
 
 print(command)
-os.system(command)
+return_code = os.system(command)
+
+if return_code != 0:
+    raise ValueError("'gcloud dataproc jobs submit ..' command exited with non-zero return code: {}".format(return_code))
--- a/gcloud_dataproc/submit.py
+++ b/gcloud_dataproc/submit.py
@@ -5,6 +5,7 @@ import getpass
 import multiprocessing
 import os
 import socket
+import subprocess
 
 p = argparse.ArgumentParser()
 p.add_argument("-c", "--cluster", default="no-vep")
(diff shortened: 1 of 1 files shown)


gcloud_dataproc/submit.py
```diff
@@ -95,4 +95,7 @@
     """ % locals()
 
 print(command)
-os.system(command)
+return_code = os.system(command)
```
nawatts: I think [`subprocess.check_call`](https://docs.python.org/2/library/subprocess.html#subprocess.check_call) would be preferable for this.

gcloud_dataproc/submit.py
```diff
@@ -95,4 +95,7 @@
...
+
+if return_code != 0:
+    raise ValueError("'gcloud dataproc jobs submit ..' command exited with non-zero return code: {}".format(return_code))
```
nawatts: Nitpicking, but this case doesn't really fit the description of ValueError.
> Raised when an operation or function receives an argument that has the right type but an inappropriate value, and the situation is not described by a more precise exception such as IndexError.
> -- https://docs.python.org/2/library/exceptions.html#exceptions.ValueError

hanars:
We decided to go with a separate clean up script for deleting temp nodes

bw2:
Thanks @nawatts . Re-opening because I think it's worth adding the error even though we don't need it anymore for deleting temp nodes.

PR #113, opened 2019-02-27
Change contig/xpos functions to accept a locus instead of a table

nawatts:
Currently, the functions for contig, contig number, and xpos all accept a Table/MatrixTable and return a value derived from the `locus` field of that table. However, for some datasets (such as a table of regions with start/stop positions) it would be useful to calculate xpos for fields other than `locus`.

This changes the functions for contig, contig number, and xpos to accept a LocusExpression instead of a Table so that they can be used on any field.

It also changes `get_expr_for_contig` to remove the "chr" prefix from contigs so that `get_expr_for_contig_number` and `get_expr_for_xpos` will work with GRCh38 loci.


knguyen142:
Makes sense to me, I'll defer to someone else for the approval, but it was helpful to see this.

hanars:
Looks good to me

PR #114, opened 2019-02-27
Delete temp nodes script

hanars:
I haven't actually been able to test this yet, but wanted to get it up in case it was helpful


gcloud_dataproc/load_dataset.py
```diff
@@ -216,6 +213,30 @@ def _create_es_nodes(settings, create_persistent_es_nodes=False):
...
+
+        "CLUSTER_MACHINE_TYPE": "n1-highmem-4",
+        "ELASTICSEARCH_VERSION": "6.3.2",
+        "ELASTICSEARCH_JVM_MEMORY": "13g",
+        "ELASTICSEARCH_DISK_SIZE": "100Gi",
+        "ELASTICSEARCH_DISK_SNAPSHOTS": None,
+
+        "KIBANA_SERVICE_PORT": 5601,
+
+        "ES_NUM_PERSISTENT_NODES": num_persistent_nodes,
```
bw2: would num_persistent_nodes be undefined?

gcloud_dataproc/delete_temp_loading_nodes.py
```diff
@@ -0,0 +1,32 @@
...
+
+p = argparse.ArgumentParser()
+p.add_argument("--num-persistent-nodes", type=int, help="For use with --num-persistent-nodes. Number of persistent data nodes to create.", default=3)
+p.add_argument("--host", help="Elastisearch host", default=os.environ.get("ELASTICSEARCH_SERVICE_HOSTNAME", "localhost"))
+p.add_argument("--port", help="Elastisearch port", default="9200")
+p.add_argument("k8s-cluster-name", help="Specifies the kubernetes cluster name that hosts elasticsearch.", required=True)
+args = p.parse_args()
+
+client = ElasticsearchClient(args.host, args.port)
+wait_for_loading_shards_transfer(client)
```
bw2: nice feature

gcloud_dataproc/delete_temp_loading_nodes.py
```diff
@@ -0,0 +1,32 @@
...
+import os
+
+from kubernetes.shell_utils import run
+from load_dataset import _get_es_node_settings, _process_kubernetes_configs, _wait_for_data_nodes_state, \
+    _compute_firewall_rule_name
+from hail_scripts.v01.load_dataset_to_es import wait_for_loading_shards_transfer
+from hail_scripts.v01.utils.elasticsearch_client import ElasticsearchClient
+
+p = argparse.ArgumentParser()
+p.add_argument("--num-persistent-nodes", type=int, help="For use with --num-persistent-nodes. Number of persistent data nodes to create.", default=3)
```
bw2: would  remove this arg and just use some dummy value

gcloud_dataproc/delete_temp_loading_nodes.py
```diff
@@ -0,0 +1,32 @@
...
+p.add_argument("--host", help="Elastisearch host", default=os.environ.get("ELASTICSEARCH_SERVICE_HOSTNAME", "localhost"))
+p.add_argument("--port", help="Elastisearch port", default="9200")
+p.add_argument("k8s-cluster-name", help="Specifies the kubernetes cluster name that hosts elasticsearch.", required=True)
+args = p.parse_args()
+
+client = ElasticsearchClient(args.host, args.port)
+wait_for_loading_shards_transfer(client)
+
+settings = _get_es_node_settings(args.k8s_cluster_name, args.num_temp_loading_nodes)
```
bw2: may be worth calling 
`_set_k8s_context(settings)` like it does in
https://github.com/macarthur-lab/hail-elasticsearch-pipelines/pull/114/files#diff-9eed0c0fd4f99bedc78484994f835391L167

so people don't accidentally delete loading nodes in the wrong cluster (for example if they're switching between a  test cluster and the production one)

gcloud_dataproc/load_dataset.py
hanars: good catch

PR #116, opened 2019-02-27
v02 rewrite base

knguyen142:
I wanted to do many small PRs in a separate directory, so people are not overwhelmed with big diffs. This PR sets up the base structure for the project using Luigi. Let me know what you guys think.

## Overview
- `luigi_pipeline/lib/hail_tasks.py`: has 2 classes, one for moving a VCF to Hail MT, and another one for MT to ElasticSearch (ES logic not spec'd out). Write now it does the bare minimal loading and dumping, but in the future hail utils will be here.
- `luigi_pipeline/seqr_loading.py`: Will be the seqr loading specific logic. The classes here inherit from the hail ones, and the dependencies are established here. All higher level seqr logic will go here.
- Configs: I think using Luigi's config files are a better alternative to passing in a bunch of arguments.

## Improvements
- Luigi provided checking for inputs and outputs. This is super helpful, as we now won't need to manually worry about what steps are done and which ones need to be executed.
- Program arguments. With a config file, we in theory can have a bunch of arguments and handle this in a clean way. Arguments are scoped to their specific task (and defined in the config file in the same way).
- Modularity. Tasks can now be independent of each other (similar to having multiple scripts that we had suggested). 
- Maintenance and improvements. New tasks can be chained onto existing ones and all of the input/output are handled.

## Testing
- Tests created! (I can't promise we can continue this coverage.)
- Ran locally (you can too with the documented steps)
- Ran on Dataproc https://console.cloud.google.com/dataproc/jobs/700e962f41504fa096d75767cc91d124?project=seqr-project&organizationId=548622027621&supportedpurview=project&region=global&graph=GCE_CPU&duration=PT1H


luigi_pipeline/lib/hail_tasks.py
```diff
@@ -0,0 +1,62 @@
...
+    return target(filename)
+
+
+class VcfFile(luigi.Task):
+    filename = luigi.Parameter()
+    def output(self):
+       return GCSorLocalTarget(self.filename)
+
+
+class HailMatrixTableTask(luigi.Task):
```
bw2: I think having a HailMatrixTableTask base class is confusing. This might just be about the name - it's a MatrixTableTask but the code only reads from VCF files (instead of reading in a .mt for example) - and a .vcf only gets read in once at the very beginning in the v01 pipeline.  In general I'd vote for using utility functions or composition instead of inheritance for sharing functionality. 
Also, it might be better to implement the full pipeline first to see what abstractions are most useful.

luigi_pipeline/requirements.txt
```diff
@@ -0,0 +1,3 @@
+elasticsearch==6.2.0
```
bw2: elasticsearch>=6.2.0  ?

luigi_pipeline/tests/test_hail_tasks.py
```diff
@@ -0,0 +1,52 @@
...
+        temp_dest_path = os.path.join(self.test_dir, 'temp_test_load.mt')
+        task = HailMatrixTableTask(source_paths=[TEST_DATA_MT_1KG], dest_path=temp_dest_path)
+
+        mt = task.import_vcf()
+        self.assertTrue(mt.count() == (30, 16))
+
+    def test_hail_matrix_table_run(self):
+        temp_dest_path = os.path.join(self.test_dir, 'temp_test_run.mt')
+
+        task = HailMatrixTableTask(source_paths=[TEST_DATA_MT_1KG], dest_path=temp_dest_path)
```
bw2: cool that you can have step-level tests

luigi_pipeline/README.md
```diff
@@ -0,0 +1,31 @@
...
+
+## Local
+### Install
+```
+$ pip install -r requirements.txt
+```
+
+### Tests
+```
+$ pip install nose
```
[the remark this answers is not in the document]
knguyen142: It's just the one I've always used, but I believe you should be able to run it with others.

luigi_pipeline/lib/hail_tasks.py
knguyen142: Yeah so I thought about the mt functionality, and figured that until it is needed it shouldn't be implemented. But when it is needed, it's not hard to add. The source paths can include `.mt` and it can do logic from that, or we can create another class and rename this to `HailMatrixTableFromVCFTask`.

I think a hail module can added to supplement when there's code that needs to be provided for non-luigi functionality, but this task and subclassing from this is a nice decomposition so we don't have unnecessary luigi-specific housekeeping functionality that bloat up the seqr tasks, and can easily chain them (in case we have another task that reads from the output mt). E.g. if we need to read from mt and we add that functionality to the task, we can create a seqr sublcass and we automatically inherit all that functionality. Furthermore, it's just python convention and specifically luigi convention (e.g. https://github.com/spotify/luigi/blob/master/luigi/contrib/dataproc.py#L43). Python is a great for scripts, but it's an object oriented language and even we should be using that more (e.g. instead of just naming util functions, we should have classes that encapsulate specific functionality).

luigi_pipeline/requirements.txt
knguyen142: Just copied and pasted from https://github.com/macarthur-lab/hail-elasticsearch-pipelines/blob/master/requirements.txt so not sure if we needed to pin it.

luigi_pipeline/configs/luigi.cfg
```diff
@@ -0,0 +1,7 @@
+[SeqrVCFToMTTask]
+source_paths = ["gs://seqr-kev/seqr-loading-test/1kg_30variants.vcf.bgz"]
+dest_path = gs://seqr-kev/seqr-loading-test/1kg.mt
+reference_mt_path = TODO
+
+[SeqrMTToESTask]
+dest_file = gs://seqr-kev/seqr-loading-test/_SUCCESS
```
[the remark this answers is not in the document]
knguyen142: Actually I meant the `_SUCCESS` that's generated from Hail. We can generate our own, but figured if they already did that specifies an atomic success, we should use that.

luigi_pipeline/README.md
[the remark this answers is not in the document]
bw2: since you're only importing `unittest` in your tests, I would remove the nose dependency unless there's a major reason to use it. Want to avoid some tests casually becoming dependent on it.

luigi_pipeline/lib/hail_tasks.py
bw2: I'm not against using classes sometimes, especially when something is naturally noun-like and has multiple data fields (for example elasticsearch_client.py). The hail code base I think has a nice balance with a few core classes and [many utilities](https://hail.is/docs/0.2/methods/index.html). I'm generally predisposed [against inheritance](https://javarevisited.blogspot.com/2013/06/why-favor-composition-over-inheritance-java-oops-design.html) except where it's clearly better than utilities/composition of objects. 
How about we revisit in subsequent pipeline PRs?

knguyen142: Yeah in the next ones as we have more functionality we'll see whether the separation makes code look cleaner and reusable.

luigi_pipeline/requirements.txt
nawatts: There's also several scripts that call `os.system("pip install elasticsearch")` without _any_ version requirement.

luigi_pipeline/README.md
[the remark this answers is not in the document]
knguyen142: Yeah I can change it to whatever if it's a small change, but just wanted to point out that nose knows.

luigi_pipeline/lib/hail_tasks.py
knguyen142: Sorry, what class should be renamed `ImportVcfTask`? Do you mean `HailMatrixTableTask`, which currently has what you suggested:

>  I could see a base HailMatrixTableTask class that implements dest_path and output and leaves run for subclasses.

luigi_pipeline/lib/hail_tasks.py
```diff
@@ -0,0 +1,62 @@
...
+
+
+class HailElasticSearchTask(luigi.Task):
+    """
+    Loads a MT to ES (TODO).
+    """
+    source_path = luigi.OptionalParameter(default=None)
+
+    def requires(self):
+        return [VcfFile(filename=self.source_path)]
```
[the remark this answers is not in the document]
knguyen142: Good point, `VcfFile` should be named `GCSorLocalFile`.

luigi_pipeline/requirements.txt
knguyen142: How about I just merge this requirements file with the other one and remove it here so we don't have redundancy?

nawatts: That sounds good. Though there may also be cause for distinguishing what packages are needed locally to submit jobs vs what packages are needed on the Spark cluster.

luigi_pipeline/lib/hail_tasks.py
nawatts: > Sorry, what class should be renamed ImportVcfTask? Do you mean HailMatrixTableTask.

Yes. I think that either this class's implementation is too specific for a base class or its name is too generic for a concrete class.

> which currently has what you suggested

I should have said "that *only* implements dest_path and output". If `HailMatrixTableTask` is intended as a base class for all tasks that output a MatrixTable, I think a more abstract implementation would be preferable, like having `run` raise NotImplementedError. For example, would a task that imports an MT from a TSV file inherit from `HailMatrixTableTask`? If so, with the current implementation, then it would have  unused `import_vcf` method.

knguyen142: Yeah so the case you suggested in similar to Ben's with `mt` files. I'm of the opinion that implementation should happen when needed but we should structure it so it can be easily modified (e.g. for your case it's very easy to support `tsv` in the existing implementation, or separate it when needed). 

The alternatives would be to
- not subclass, which is what Ben's suggesting, but a clear abstraction is needed right now so we don't have a bunch of unnecessary logic in the seqr tasks and can easily add more seqr classes.
- Make  `HailMatrixTableTask` naming super specific like `HailMatrixTableFromVcfTask`, but I think this is premature.

I think having `run raise NotImplementedError` is a good idea however. Keep in mind that these base classes are there to consolidate hail logic from seqr logic, and both are going to change a lot. So the naming is fuzzy to support the changes as needed, but separating them early on is because of luigi and python convention.

@mike-w-wilson I can show you how to `NotImplementedError` works and help you implement it.

luigi_pipeline/requirements.txt
bw2: @knguyen142  merging this into the top-level requirements file is a good idea.
@nawatts that's a good point. The os.system("pip install elasticsearch") should only be present in scripts that run on Spark - not sure if there's a better way to install dependencies there. I don't have a preference for `>=` vs. `==` for lib. versions. Do you guys?

nawatts: A cluster start script could parse a `requirements.txt` file and call `cluster start --packages=...`

PR #122, opened 2019-03-13
ref data exac

knguyen142:
Exac data computation. This is already run.

### DIFF
--- a/download_and_create_reference_datasets/v02/hail_scripts/write_exac.py
+++ b/download_and_create_reference_datasets/v02/hail_scripts/write_exac.py
@@ -1,0 +1,50 @@
+import logging
+
+import hail as hl
+
+from hail_scripts.v02.utils.hail_utils import import_vcf
+
+CONFIG= {
+    '37': 'gs://seqr-reference-data/GRCh37/gnomad/ExAC.r1.sites.vep.vcf.bgz',
+    '38': 'gs://seqr-reference-data/GRCh38/gnomad/ExAC.r1.sites.liftover.b38.vcf.gz',
+}
+
+logger = logging.getLogger('v02.hail_scripts.create_exac_ht')
+
+
+def vcf_to_mt(path, genome_version):
+    '''
+    Converts exac vcf to mt and adds info.AF_POPMAX annotation.
+
+    :param path: vcf path
+    :param genome_version: genome version
+    :return:
+    '''
+    mt = import_vcf(path,
+                    genome_version=genome_version,
+                    min_partitions=1000)
+
+    # Annotate rows with popmax af if AC_POPMAX is not 'NA'. Value =
+    # AC/AN, and null otherwise.
+    cond = (mt.info.AC_POPMAX[mt.a_index-1] != "NA")
+    value = (hl.float((hl.int(mt.info.AC_POPMAX[mt.a_index - 1]) /
+                       hl.int(mt.info.AN_POPMAX[mt.a_index - 1]))))
+    alternative = hl.null(hl.tfloat)
+    mt = mt.annotate_rows(info=mt.info.annotate(AF_POPMAX=hl.cond(cond, value, alternative)))
+
+    return mt
+
+
+def run():
+   for genome_version, path in CONFIG.items():
+       logger.info('reading from input path: %s' % path)
+       mt = vcf_to_mt(path, genome_version)
+
+       mt.describe()
+
+       output_path = path.replace(".vcf", "").replace(".gz", "").replace(".bgz", "")\
+                         .replace(".*", "").replace("*", "") + ".ht"
+       logger.info('writing to output path: %s' % output_path)
+       mt.rows().write(output_path)
+
+run()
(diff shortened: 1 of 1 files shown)


download_and_create_reference_datasets/v02/hail_scripts/write_exac.py
```diff
@@ -0,0 +1,50 @@
...
+    :param genome_version: genome version
+    :return:
+    '''
+    mt = import_vcf(path,
+                    genome_version=genome_version,
+                    min_partitions=1000)
+
+    # Annotate rows with popmax af if AC_POPMAX is not 'NA'. Value =
+    # AC/AN, and null otherwise.
+    cond = (mt.info.AC_POPMAX[mt.a_index-1] != "NA")
```
bw2: Comparing to https://github.com/macarthur-lab/hail-elasticsearch-pipelines/blob/master/hail_scripts/v01/utils/add_exac.py#L155
why does this check that `AC_POPMAX != 'NA'` but not AN_POPMAX?

download_and_create_reference_datasets/v02/hail_scripts/write_exac.py
```diff
@@ -0,0 +1,50 @@
...
+    value = (hl.float((hl.int(mt.info.AC_POPMAX[mt.a_index - 1]) /
+                       hl.int(mt.info.AN_POPMAX[mt.a_index - 1]))))
+    alternative = hl.null(hl.tfloat)
+    mt = mt.annotate_rows(info=mt.info.annotate(AF_POPMAX=hl.cond(cond, value, alternative)))
```
bw2: Would also discard unused fields as the 1st step 
https://github.com/macarthur-lab/hail-elasticsearch-pipelines/blob/master/hail_scripts/v01/utils/add_exac.py#L101

download_and_create_reference_datasets/v02/hail_scripts/write_exac.py
knguyen142: So I tried adding that with `and hl.int(exac_mt.info.AN_POPMAX[exac_mt.a_index-1]))` but had problems. Whenever I appended any boolean to it I would get `TypeError: 'Expression' objects have no static length: use 'hl.len' for the length of collections` and tried googling it but couldn't get an answer.

I did a query for ones with AN = 'NA' and AC <= 0 and that returned nothing. 
```
mt1 = mt.filter_rows(mt.info.AN_POPMAX[mt.a_index - 1] != 'NA')
mt2 = mt1.filter_rows(hl.int(mt1.info.AC_POPMAX[mt1.a_index - 1])<=0
mt2.count()
```
Note I had to separate the queries into 2 because appending them for filter rows returned the same error.

download_and_create_reference_datasets/v02/hail_scripts/write_exac.py
knguyen142: I didn't for this ref data for 2 reasons:
1. It does not conform to our convention of load and dump (e.g. load_from_vcf)
2. We might need those in the future.
The selection will happen when combining the data sets. I believe filtering at this level will reduce the size but not the performance time of combining the dataset, since selecting fields should be pretty fast.

download_and_create_reference_datasets/v02/hail_scripts/write_exac.py
nawatts: For Hail expressions, the `==`, `!=`, `>`, `<` operators return a Hail [BooleanExpression](https://hail.is/docs/0.2/expressions.html#hail.expr.expressions.BooleanExpression), not a Python bool. To do boolean logic with Hail, use `&`, `|`, and `~` instead of `and`, `or`, and `not`. Beware of the priority of operations though. The bitwise operators have higher priority than comparison operators, so parentheses are usually necessary.

download_and_create_reference_datasets/v02/hail_scripts/write_exac.py
```diff
@@ -0,0 +1,50 @@
...
+
+def run():
+   for genome_version, path in CONFIG.items():
+       logger.info('reading from input path: %s' % path)
+       mt = vcf_to_mt(path, genome_version)
+
+       mt.describe()
+
+       output_path = path.replace(".vcf", "").replace(".gz", "").replace(".bgz", "")\
+                         .replace(".*", "").replace("*", "") + ".ht"
```
nawatts: The last two replace calls for `.*` and `*` seem unnecessary.

download_and_create_reference_datasets/v02/hail_scripts/write_exac.py
knguyen142: So I read about Hail booleans yesterday and believe I tried '&' and '&&', with errors. I'll try again, but I think I tried Hail Boolean Expressions with and mixed with other python booleans and they worked. I might've made a mistake in analyzing the results/errors, but tried almost all combinations.

bw2:
is this PR still relevant?

bw2:
If nobody objects, I can delete this PR next week.

PR #124, opened 2019-03-19
seqr write validation ht

knguyen142:
Another day, another data set. This is the validation data set used in seqr to validate the incoming variants https://github.com/macarthur-lab/hail-elasticsearch-pipelines/blob/master/hail_scripts/v01/utils/validate_vds.py#L43.

It wasn't a lot of code and isn't run often, so I decided to do an almost 1 to 1 port of https://github.com/macarthur-lab/hail-elasticsearch-pipelines/blob/master/download_and_create_reference_datasets/v01/hail_scripts/write_dataset_validation_kt.py.

# Testing:
## For non-coding file:
v01: 
```
> hc.read_table('gs://seqr-reference-data/GRCh37/validate_vds/common_noncoding_variants.grch37.kt').count()
- 1880L

```

v02:
```
> hl.read_table('gs://seqr-kev/combined-test/validation-noncoding.ht').count()
- 2243
```

## For coding variants:
v01: 
```
> hc.read_table('gs://seqr-reference-data/GRCh37/validate_vds/common_coding_variants.grch37.kt').count()
- 354L
```

v02:
```
> hl.read_table('gs://seqr-kev/combined-test/validation-noncoding.ht').count()
- 359

```

On cursory glance, there seems to be overlap between the 2.

### DIFF
--- a/download_and_create_reference_datasets/v02/hail_scripts/write_dataset_validation_ht.py
+++ b/download_and_create_reference_datasets/v02/hail_scripts/write_dataset_validation_ht.py
@@ -5,6 +5,13 @@ import hail as hl
 from hail_scripts.v02.utils.computed_fields.vep import get_expr_for_vep_sorted_transcript_consequences_array, \
     get_expr_for_worst_transcript_consequence_annotations_struct, CONSEQUENCE_TERM_RANK_LOOKUP
 
+VALIDATION_KEYTABLE_PATHS = {
+    'coding_37': 'gs://seqr-reference-data/GRCh37/validate_ht/common_coding_variants.grch37.ht',
+    'coding_38': 'gs://seqr-reference-data/GRCh38/validate_ht/common_coding_variants.grch38.ht',
+    'noncoding_37': 'gs://seqr-reference-data/GRCh37/validate_ht/common_noncoding_variants.grch37.ht',
+    'noncoding_38': 'gs://seqr-reference-data/GRCh38/validate_ht/common_noncoding_variants.grch38.ht',
+}
+
 logging.basicConfig(format='%(asctime)s %(levelname)-8s %(message)s')
 logger = logging.getLogger()
 logger.setLevel(logging.INFO)
@@ -16,10 +23,12 @@ args = p.parse_args()
 
 def read_gnomad_subset(genome_version):
     logger.info("==> Read gnomAD subset")
+    filtered_contig = '1' if genome_version == '37' else 'chr1'
 
     # select a subset of gnomad genomes variants that are in > 90% of samples
-    ht = hl.split_multi(hl.read_table('gs://gnomad-public/release/2.1.1/ht/genomes/gnomad.genomes.r2.1.1.sites.ht'))
-    ht = hl.filter_intervals(ht, [hl.parse_locus_interval('1', reference_genome='GRCh%s'%genome_version)])
+    ht = hl.read_table('gs://gnomad-public/release/2.1.1/ht/genomes/gnomad.genomes.r2.1.1.sites.ht')
+    ht = hl.filter_intervals(ht, [hl.parse_locus_interval(filtered_contig,
+                                                          reference_genome='GRCh%s'%genome_version)])
     ht = ht.filter(ht.freq[0].AF > 0.90)
 
     ht = ht.annotate(sorted_transaction_consequences=(
@@ -37,7 +46,7 @@ def read_gnomad_subset(genome_version):
 def write_out_ht(ht, output_path):
     ht = ht.select()
     print(ht.count())
-    # ht.write(output_path, overwrite=True)
+    ht.write(output_path, overwrite=True)
 
 
 ht = read_gnomad_subset(args.genome_version)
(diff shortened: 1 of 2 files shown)


download_and_create_reference_datasets/v02/hail_scripts/write_dataset_validation_ht.py
```diff
@@ -0,0 +1,53 @@
...
+p.add_argument("-g", "--genome-version", help="Genome build: 37 or 38", choices=["37", "38"], default="37")
+args = p.parse_args()
+
+
+def read_gnomad_subset(genome_version):
+    logger.info("==> Read gnomAD subset")
+
+    # select a subset of gnomad genomes variants that are in > 90% of samples
+    ht = hl.split_multi(hl.read_table('gs://gnomad-public/release/2.1.1/ht/genomes/gnomad.genomes.r2.1.1.sites.ht'))
+    ht = hl.filter_intervals(ht, [hl.parse_locus_interval('1', reference_genome='GRCh%s'%genome_version)])
```
nawatts: Hail 0.2 validates loci, and GRCh38 contigs are prefixed with "chr". So for GRCh38, this would need to be `chr1` instead of `1`.

download_and_create_reference_datasets/v02/hail_scripts/write_dataset_validation_ht.py
```diff
@@ -0,0 +1,53 @@
...
+p = ap.ArgumentParser()
+p.add_argument("-g", "--genome-version", help="Genome build: 37 or 38", choices=["37", "38"], default="37")
+args = p.parse_args()
+
+
+def read_gnomad_subset(genome_version):
+    logger.info("==> Read gnomAD subset")
+
+    # select a subset of gnomad genomes variants that are in > 90% of samples
+    ht = hl.split_multi(hl.read_table('gs://gnomad-public/release/2.1.1/ht/genomes/gnomad.genomes.r2.1.1.sites.ht'))
```
nawatts: Multi-allelic variants are already split in the gnomAD 2.1+ Hail table and VCF.

See "Split all the multi-allelic variants" in https://macarthurlab.org/2018/10/17/gnomad-v2-1/

download_and_create_reference_datasets/v02/hail_scripts/write_dataset_validation_ht.py
```diff
@@ -0,0 +1,53 @@
...
+    ht = ht.select()
+    print(ht.count())
+    # ht.write(output_path, overwrite=True)
+
+
+ht = read_gnomad_subset(args.genome_version)
+ht.persist()
+
+coding_ht = ht.filter(
+    hl.int(ht.main_transcript.major_consequence_rank) <= hl.int(CONSEQUENCE_TERM_RANK_LOOKUP.get('synonymous_variant')))
```
nawatts: Is the cast to int necessary here? Both `main_transcript.major_consequence_rank` and `CONSEQUENCE_TERM_RANK_LOOKUP.get(`...`)` should already be ints.

download_and_create_reference_datasets/v02/hail_scripts/write_dataset_validation_ht.py
```diff
@@ -0,0 +1,53 @@
...
+write_out_ht(coding_ht, 'gs://seqr-kev/combined-test/validation-coding.ht')
```
nawatts: Is the plan to switch to the `seqr-reference-data` bucket once the full pipeline is done?

hail_scripts/v02/utils/computed_fields/vep.py
```diff
@@ -176,7 +183,7 @@ def get_expr_for_vep_sorted_transcript_consequences_array(vep_root, include_codi
             ]
         )
 
-    omit_consequence_terms = hl.set(["upstream_gene_variant", "downstream_gene_variant"])
+    omit_consequence_terms = hl.set(omit_consequences) if len(omit_consequences) > 0 else hl.empty_set(hl.tstr)
```
nawatts: Nitpicking, but `if omit_consequences else hl.empty_set(hl.tstr)` would be slightly more idiomatic.

download_and_create_reference_datasets/v02/hail_scripts/write_dataset_validation_ht.py
```diff
@@ -0,0 +1,53 @@
...
+
+    ht.describe()
+
+    return ht
+
+
+def write_out_ht(ht, output_path):
+    ht = ht.select()
+    print(ht.count())
+    # ht.write(output_path, overwrite=True)
```
nawatts: Looks like this may be commented out from testing.

download_and_create_reference_datasets/v02/hail_scripts/write_dataset_validation_ht.py
nawatts: Since this Hail table is based on GRCh37, should `read_gnomad_subset` raise an exception if GRCh38 is requested?

download_and_create_reference_datasets/v02/hail_scripts/write_dataset_validation_ht.py
knguyen142: Ahh that's right. I couldn't run it on 38 because it's not available yet so didn't test.

download_and_create_reference_datasets/v02/hail_scripts/write_dataset_validation_ht.py
knguyen142: Good to know about splitting, I didn't want to assume.
So there are a lot of cases not considered in this script (38 being one of them). The way I figured it, it's a script that's gonna be run probably once and most likely by me. So this is more of a good it done so we can make good progress.

download_and_create_reference_datasets/v02/hail_scripts/write_dataset_validation_ht.py
knguyen142: It ended up being from the port, but I kept it there after realizing this because it doesn't hurt to be sure.

download_and_create_reference_datasets/v02/hail_scripts/write_dataset_validation_ht.py
knguyen142: Yes, forgot to remove this.

knguyen142:
Thinking about this now, it seems like with the new additional variants as candidates for validation might mess with the threshold? We check if a certain percentage is in this set to classify it as being in that set, does this threshold need to go down now? @bw2

download_and_create_reference_datasets/v02/hail_scripts/write_dataset_validation_ht.py
knguyen142: Also, when 38 is ready the path would probably include a 37 and 38, right? I'll change it when it happens.

PR #125, opened 2019-03-20
Kev/v02 ref data combine

knguyen142:
Combining all seqr reference data into one combined dataset. This is a rewrite of  https://github.com/macarthur-lab/hail-elasticsearch-pipelines/blob/master/download_and_create_reference_datasets/v01/hail_scripts/combine_all_variant_level_reference_data.py.

### Improvements
- Everything done in one step and no intermediary files.
- With outer joins we don't have to do a 2 step process of generating the union of variants from all datasets and annotating them one by one
- Instead of having individual functions for each dataset that fetched its own fields, there's a config with a generic syntax for selecting annotations
- Removed exclude individual dataset logic (and instead, select on specific datasets to run)

# Testing
Since we used new gnomad datasets (2.1.1 vs. 2.0.2), the combined data is not identical and an exact count sanity check cannot be done. However, the counts are close and are reasonable considering the added variants from gnomad https://docs.google.com/spreadsheets/d/1qHyQhZMhViT0McQ3TyB0EdVjpLUqJq4wWAfm_ackhhw/edit#gid=0.

V01
```

         Samples: 0
        Variants: 8657527985
       Call Rate: nan
         Contigs: ['X', '12', '8', '19', '4', '15', '11', '9', 'Y', '22', '13', '16', '5', '10', '21', '6', '1', '17', '14', '20', '2', '18', '7', '3']
   Multiallelics: 0
            SNPs: 8576020113
            MNPs: 2609239
      Insertions: 29242938
       Deletions: 46614438
 Complex Alleles: 3041257
    Star Alleles: 0
     Max Alleles: 2
```

V02
```
'==============================
Number of variants: 8657820153
==============================
Alleles per variant
-------------------
  2 alleles: 8657820153 variants
==============================
Variants per contig
-------------------
   1: 682247951 variants
   2: 721367831 variants
   3: 589856942 variants
   4: 568340246 variants
   5: 538050957 variants
   6: 506999424 variants
   7: 470723279 variants
   8: 432696742 variants
   9: 363827693 variants
  10: 397888458 variants
  11: 397105121 variants
  12: 395320276 variants
  13: 289543190 variants
  14: 267455272 variants
  15: 247503880 variants
  16: 239162095 variants
  17: 235919484 variants
  18: 226106399 variants
  19: 169554775 variants
  20: 180243501 variants
  21: 106424435 variants
  22: 105822167 variants
   X: 456668087 variants
   Y: 68991948 variants
==============================
Allele type distribution
------------------------
        SNP: 8575982091 alternate alleles
   Deletion: 46820464 alternate alleles
  Insertion: 29367097 alternate alleles
    Complex: 3041262 alternate alleles
        MNP: 2609239 alternate alleles
=============================='
```

Also, hemi count > 0 is close (`4,804,694` vs. `4,854,252`) which is important since this is not provided by new gnomad and has to be computed.

I did a quick sanity check by manually comparing 4 variants from both and the fields are the same. Not exhaustive, but validating this has not been pain-free.

### DIFF
--- a/download_and_create_reference_datasets/v02/hail_scripts/write_1kg_ht.py
+++ b/download_and_create_reference_datasets/v02/hail_scripts/write_1kg_ht.py
@@ -46,14 +46,22 @@ def vcf_to_mt(path, genome_version):
 
     return all_mt
 
+def annotate_mt(mt):
+    # Annotate POPMAX_AF, which is max of respective fields using a_index for multi-allelics.
+    return mt.annotate_rows(POPMAX_AF=hl.max(mt.info.AFR_AF[mt.a_index-1],
+                                             mt.info.AMR_AF[mt.a_index - 1],
+                                             mt.info.EAS_AF[mt.a_index - 1],
+                                             mt.info.EUR_AF[mt.a_index - 1],
+                                             mt.info.SAS_AF[mt.a_index - 1]))
 
 def run():
    for genome_version, path in CONFIG.items():
        logger.info('reading from input path: %s' % path)
+
        mt = vcf_to_mt(path, genome_version)
+       mt = annotate_mt(mt)
 
        mt.describe()
-       mt.rows().show()
 
        output_path = path.replace(".vcf", "").replace(".gz", "").replace(".bgz", "")\
                          .replace(".*", "").replace("*", "") + ".ht"
--- a/download_and_create_reference_datasets/v02/hail_scripts/write_combined_reference_data_ht.py
+++ b/download_and_create_reference_datasets/v02/hail_scripts/write_combined_reference_data_ht.py
@@ -1,272 +1,167 @@
-import argparse
-import hail
-import logging
-from pprint import pprint
-import time
-
-#from hail_scripts.v01.utils.add_1kg_phase3 import add_1kg_phase3_to_vds, read_1kg_phase3_vds
-#from hail_scripts.v01.utils.add_cadd import add_cadd_to_vds, read_cadd_vds
-#from hail_scripts.v01.utils.add_dbnsfp import add_dbnsfp_to_vds, read_dbnsfp_vds
-#from hail_scripts.v01.utils.add_eigen import add_eigen_to_vds, read_eigen_vds
-#from hail_scripts.v01.utils.add_exac import add_exac_to_vds, read_exac_vds
-#from hail_scripts.v01.utils.add_gnomad import add_gnomad_to_vds, read_gnomad_vds
-#from hail_scripts.v01.utils.add_gnomad_coverage import add_gnomad_exome_coverage_to_vds, add_gnomad_genome_coverage_to_vds
-#from hail_scripts.v01.utils.add_mpc import add_mpc_to_vds, read_mpc_vds
-#from hail_scripts.v01.utils.add_primate_ai import add_primate_ai_to_vds, read_primate_ai_vds
-#from hail_scripts.v01.utils.add_topmed import add_topmed_to_vds, read_topmed_vds
-#from hail_scripts.v01.utils.gcloud_utils import delete_gcloud_file
-#from hail_scripts.v01.utils.hail_utils import create_hail_context
-#from hail_scripts.v01.utils.vds_utils import write_vds, read_vds
-
-logging.basicConfig(format='%(asctime)s %(levelname)-8s %(message)s')
-logger = logging.getLogger()
-logger.setLevel(logging.INFO)
-
-
-p = argparse.ArgumentParser()
-p.add_argument("-g", "--genome-version", help="Genome build: 37 or 38", choices=["37", "38"], required=True)
-
-p.add_argument("--output-vds", help="Output vds path",
-               default="gs://seqr-reference-data/GRCh{genome_version}/all_reference_data/combined_reference_data_grch{genome_version}.ht")
-
-p.add_argument("--exclude-topmed", action="store_true", help="Don't add TopMed AFs. Intended for testing.")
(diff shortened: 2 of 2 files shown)


download_and_create_reference_datasets/v02/hail_scripts/write_combined_reference_data_ht.py
```diff
@@ -1,272 +1,254 @@
...
+    selects = {}
+    global_idx = hl.eval(ht.globals.freq_index_dict['gnomad'])
+    selects['AF'] = ht.freq[global_idx].AF
+    selects['AN'] = ht.freq[global_idx].AN
+    selects['AC'] = ht.freq[global_idx].AC
+    selects['hom'] = ht.freq[global_idx].homozygote_count
+
+    selects['POPMAX_AF'] = ht.popmax[ht.globals.popmax_index_dict['gnomad']].AF
+    selects['FAF_AF'] = ht.faf[ht.globals.popmax_index_dict['gnomad']].faf95
+    selects['hemi'] = hl.cond(ht.locus.in_autosome_or_par(),
```
knguyen142: @lfrancioli I worked with you on this query, can you confirm that it's right?

download_and_create_reference_datasets/v02/hail_scripts/write_1kg_ht.py
```diff
@@ -46,14 +46,22 @@ def vcf_to_mt(path, genome_version):
 
     return all_mt
 
+def annotate_mt(mt):
+    # Annotate POPMAX_AF, which is max of respective fields using a_index for multi-allelics.
+    return mt.annotate_rows(POPMAX_AF=hl.max(mt.info.AFR_AF[mt.a_index-1],
```
bw2: This is only works if the mt is bi-allelic. (A multiallelic mt would silently overwrite popmax for all except 1 of the alelles). Would be great to check/throw an error on multi-allelic mt.

nawatts: Multi-allelic variants are split in `vcf_to_mt` before this is run. The `a_index` field is added by [split_multi](https://hail.is/docs/0.2/methods/genetics.html#hail.methods.split_multi).

bw2: Even though it's correct within the current pipeline, it's a hard to catch bug if the code is reused or upstream code is modified. Having methods that double check args adds overhead but I'd vote for making the assumption explicit in this case.

nawatts: It wouldn't be an obvious error message, but this should fail on non-split tables since the `a_index` field would not exist.

knguyen142: Yeah I think it's important to remember the frequency with which this is run when talking about exception cases. In keeping with the spirit of this repo, a lot of cases are not considered haha. When I'm rewriting, I'm paying more attention to overall maintainability and cleanliness, but also the frequency with which it is used to allocate how much time is spent making it robust.

bw2: I'm thinking of cases where a_index is added without splitting (like https://github.com/macarthur-lab/hail-elasticsearch-pipelines/blob/b8b2c7c5b50697941fe9737c7e12c592b0a74341/download_and_create_reference_datasets/v02/hail_scripts/write_1kg_ht.py#L36) - which was also necessary in v01 and led to this bug happening. 
I agree, it's quite unlikely to happen.

knguyen142: Hmm, maybe I'm not understanding. Can you clarify in what cases it would report an incorrect info and what the incorrect info would be, as well as the check that would catch this?

download_and_create_reference_datasets/v02/hail_scripts/write_combined_reference_data_ht.py
```diff
@@ -1,272 +1,254 @@
...
+        field_name: hl.struct(**select_fields)
+    }
+
+    print(select_fields)
+    return base_ht.select(**select_query).distinct()
+
+def join_hts(datasets, coverage_datasets=[], reference_genome='37'):
+    # Get a list of hail tables and combine into an outer join.
+    hts = [get_ht(dataset, reference_genome) for dataset in datasets]
+    joined_ht = reduce((lambda joined_ht, ht: joined_ht.join(ht, 'outer')), hts)
```
bw2: cool way to implement the join

download_and_create_reference_datasets/v02/hail_scripts/write_1kg_ht.py
bw2: The current code is correct.. but if the upstream code ever broke in a way that ran this line:
` biallelic_mt = biallelic_mt.annotate_rows(a_index=1, was_split=False) `
on multiallelic variants, annotate_mt(mt) would become incorrect.
I guess this is just a reminder to watch out if/when modifying the splitting logic, since 
` biallelic_mt.annotate_rows(a_index=1, was_split=False) `
is still necessary in places. 
Checking for it explicitly would require filtering for  `hl.len(mt.alleles) > 2` and making sure the result is 0.

download_and_create_reference_datasets/v02/hail_scripts/write_combined_reference_data_ht.py
knguyen142: Thanks!

PR #128, opened 2019-03-22
seqr loading luigi pipeline validate

knguyen142:
## Summary
Validation of the dataset by comparing the matches against common non-coding and coding variants (generated from https://github.com/macarthur-lab/hail-elasticsearch-pipelines/blob/master/download_and_create_reference_datasets/v02/hail_scripts/write_dataset_validation_ht.py). Rewrite of https://github.com/macarthur-lab/hail-elasticsearch-pipelines/blob/master/hail_scripts/v01/utils/validate_vds.py#L29-L75. 

## Design
Compute the percentage of matches from the dataset against non-coding and coding variants to impute if genome or exome. If above a threshold, we infer that it includes the category region.

## Implementation
- separate `hl_mt_impute_sample_type` in matrix table class that calculates matches
- `validate_mt` in seqr task that uses the above to check if valid.
- Valid conditions:
  - 'WES' reported and matches against coding variants but not non-coding variants
  - 'WGS' reported that matches against both coding and non-coding variants

## Other Goodies
- Added small hail tables for the validation set (~1k and ~350 variants respectively) for local testing and local runs
- added a GlobalConfig mechanism to put in configurable parameters accessible throughout runtime

## Testing
- (some) Tests included! (@mike-w-wilson you can take a look at these for some patterns--and probably some anti-patterns hehe)
- Not tested extensively against a real dataset because I don't have one


luigi_pipeline/lib/hail_tasks.py
```diff
@@ -25,6 +27,33 @@ class HailMatrixTableTask(luigi.Task):
...
+        :param genome_version: reference genome version
+        :param threshold: if the matched percentage is over this threshold, we classify as match
+        :return: a dict of coding/non-coding to dict with 'matched_count', 'total_count' and 'match' boolean.
+        """
+        stats = {}
+        types_to_ht_path = {
+            'noncoding': GlobalConfig().param_kwargs['validation_%s_noncoding_ht' % genome_version],
+            'coding': GlobalConfig().param_kwargs['validation_%s_coding_ht' % genome_version]
+        }
+        for type, ht_path in types_to_ht_path.items():
```
nawatts: Might be nice to use a different variable name to avoid shadowing the built-in `type`.

luigi_pipeline/seqr_loading.py
```diff
@@ -1,16 +1,76 @@
...
 
+class SeqrValidationError(Exception):
+    pass
+
 class SeqrVCFToMTTask(HailMatrixTableTask):
     """
     Inherits from a Hail MT Class to get helper function logic. Main logic to do annotations here.
     """
     reference_mt_path = luigi.Parameter(description='Path to the matrix table storing the reference variants.')
+    sample_type = luigi.Parameter(description='Sample type, WGS or WES')
```
nawatts: If this is supposed to be either "WGS" or "WES", should it use [luigi.ChoiceParameter](https://luigi.readthedocs.io/en/stable/api/luigi.parameter.html#luigi.parameter.ChoiceParameter)?

luigi_pipeline/seqr_loading.py
```diff
@@ -1,16 +1,76 @@
...
+        if not has_coding and not has_noncoding:
+            # No common variants detected.
+            raise SeqrValidationError(
+                'Genome version validation error: dataset specified as GRCh{genome_version} but doesn\'t contain '
+                'the expected number of common GRCh{genome_version} variants'.format(genome_version=genome_version)
+            )
+        elif has_noncoding and not has_coding:
+            # Non coding only.
+            raise SeqrValidationError(
+                'Dataset is contains noncoding variants but is missing common coding variants for'
```
nawatts: Should this be "Dataset contains" instead of "Dataset is contains"? Also, this error message is not consistent with the others. It does not say "___ validation error" and it is not clear why this is a problem or how it can be corrected. The others have a definite path to resolution: specify the correct reference genome version or sample type.

luigi_pipeline/tests/test_hail_tasks.py
```diff
@@ -17,30 +18,69 @@ def tearDown(self):
...
+        return HailMatrixTableTask(source_paths=[TEST_DATA_MT_1KG],
+                                   dest_path=temp_dest_path,
+                                   genome_version='37')
+
+    def _set_validation_configs(self):
+        global_config = GlobalConfig()
+        global_config.param_kwargs['validation_37_coding_ht'] = global_config.validation_37_coding_ht = \
+            'tests/data/validation_37_coding.ht'
+        global_config.param_kwargs['validation_37_noncoding_ht'] = global_config.validation_37_noncoding_ht = \
+            'tests/data/validation_37_noncoding.ht'
```
nawatts: Is there a Luigi configuration file for tests that these could be read from?

luigi_pipeline/seqr_loading.py
knguyen142: Nice, forgot about that when I was looking for ways to limit a Luigi Parameter.

luigi_pipeline/seqr_loading.py
knguyen142: Hmm, I just modified https://github.com/macarthur-lab/hail-elasticsearch-pipelines/blob/master/hail_scripts/v01/utils/validate_vds.py#L61. I think this is just an error on the dataset, but can be more descriptive ('Please verify the dataset is valid.') But I can't think of much else besides that?

luigi_pipeline/tests/test_hail_tasks.py
knguyen142: Yeah I briefly looked into specifying a config file in code, which would be a nice to have for generic changes, but couldn't find it in the allocated time :(

nawatts: Ok. It's a little unfortunate that tests have to mock both the global_config attribute and a value in global_config.param_kwargs.

luigi_pipeline/seqr_loading.py
nawatts: Fair enough. I was thinking that, depending on how errors are reported, this could look like an information message after which the program stops.

knguyen142:
> The test data Hail tables could be [repartitioned](https://hail.is/docs/0.2/hail.Table.html#hail.Table.repartition) to avoid adding so many files to git.

Sweet, I didn't know partition was related to the number of files. Changed now.

luigi_pipeline/lib/hail_tasks.py
```diff
@@ -25,6 +27,33 @@ class HailMatrixTableTask(luigi.Task):
     """
     source_paths = luigi.ListParameter(description='List of paths to VCFs to be loaded.')
     dest_path = luigi.Parameter(description='Path to write the matrix table.')
+    genome_version = luigi.Parameter(description='Reference Genome Version (37 or 38)')
+
+    @staticmethod
+    def hl_mt_impute_sample_type(mt, genome_version, threshold=0.3):
+        """
+        Impute the sample type by chcking against a list of common coding and non-coding variants.
```
bw2: Sorry took a while to get to this.. the logic LGTM.  Re. method boundaries and names,
- it's confusing that the "impute-sample-type" method returns counts rather than the sample type, and the actual decision about sample type happens elsewhere. I'd vote for putting the code from line 39-61 (https://github.com/macarthur-lab/hail-elasticsearch-pipelines/pull/128/files#diff-7087d398dbeca709f67c00788f5802deR39) inside this method, and have the return value be "WES" or "WSG". 

- also would call this `impute_sample_type` instead of `hl_mt_impute_sample_type` since it's already in a HailMatrixTableTask class.

knguyen142: Good point on the naming. I thought it was weird as well, and considered having the function return a type ('WGS', 'WES', 'unknown'). However, I don't think the validation logic (throwing errors) should be included there, since this is seqr specific. I'll just rename it `sample_type_stats`.
And I remembered having a reason for the `hl_mt` prefix, but can't remember it now. I'll change it and see if I remember later.

PR #129, opened 2019-03-28
naming refactor

knguyen142:
Naming refactor according to https://github.com/macarthur-lab/hail-elasticsearch-pipelines/pull/128

fyi @bw2


bw2:
I like this better, but the sample type imputation isn't seqr-specific, so I think it is valid to have a generic method that returns "WES" or "WGS" + throws an error if `not has_coding` (or two different errors depending on whether `has_non_coding` is also True/False).

PR #130, opened 2019-04-02
adding subset and remap functionality

mike-w-wilson:
Remapping and subsetting tests were also added. Files needed for both are created during the tests.

### DIFF
--- a/luigi_pipeline/lib/hail_tasks.py
+++ b/luigi_pipeline/lib/hail_tasks.py
@@ -1,17 +1,26 @@
 """
 Tasks for Hail.
 """
+import logging
+
 import hail as hl
 import luigi
 from luigi.contrib import gcs
 
 from lib.global_config import GlobalConfig
 
+logger = logging.getLogger(__name__)
+
+class MatrixTableSampleSetError(Exception):
+    def __init__(self, message, missing_samples):
+
+        super().__init__(message)
+        self.missing_samples = missing_samples
+        pass
 
 def GCSorLocalTarget(filename):
     target = gcs.GCSTarget if filename.startswith('gs://') else luigi.LocalTarget
     return target(filename)
-
 
 class VcfFile(luigi.Task):
     filename = luigi.Parameter()
@@ -49,7 +58,7 @@ class HailMatrixTableTask(luigi.Task):
     @staticmethod
     def sample_type_stats(mt, genome_version, threshold=0.3):
         """
-        Calculate stats for sample type by chcking against a list of common coding and non-coding variants.
+        Calculate stats for sample type by checking against a list of common coding and non-coding variants.
         If the match for each respective type is over the threshold, we return a match.
 
         :param mt: Matrix Table to check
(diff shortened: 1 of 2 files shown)


luigi_pipeline/lib/hail_tasks.py
```diff
@@ -72,6 +81,63 @@ def sample_type_stats(mt, genome_version, threshold=0.3):
...
 
+    @staticmethod
+    def subset_samples(mt, subset_path):
+        """
+        Subset the MatrixTable to the provided list of samples
+        :param mt: MatrixTable from VCF
+        :param subset_path: Path to a file with a single column 's'
+        :return: MatrixTable subsetted to list of samples
+        """
+        subset_ht = hl.import_table(subset_path, no_header=False).key_by('s')
```
nawatts: `import_table` accepts a `key` parameter. https://hail.is/docs/0.2/methods/impex.html#hail.methods.import_table

Same comment for `remap_sample_ids`.

luigi_pipeline/tests/test_hail_tasks.py
```diff
@@ -90,3 +91,53 @@ def requires(self):
...
+        mt = hl.import_vcf(TEST_DATA_MT_1KG)
+        self.assertRaises(MatrixTableSampleSetError, HailMatrixTableTask.subset_samples, mt, self.create_temp_sample_subset_file(mt, 1, True))
+
+    def test_hail_matrix_table_subset_wrong_sample_id_correct(self):
+        task = HailMatrixTableTask(source_paths=[TEST_DATA_MT_1KG], dest_path="empty", genome_version='37')
+        mt = hl.import_vcf(TEST_DATA_MT_1KG)
+        try:
+            task.subset_samples(mt, self.create_temp_sample_subset_file(mt, 1, True))
+        except MatrixTableSampleSetError as e:
+            if e.missing_samples != ['wrong_sample']:
```
nawatts: Maybe `assertEqual` instead of calling `fail` based on a condition?

luigi_pipeline/tests/test_hail_tasks.py
```diff
@@ -90,3 +91,53 @@ def requires(self):
...
+
+    def test_hail_matrix_table_subset_raise_e(self):
+        mt = hl.import_vcf(TEST_DATA_MT_1KG)
+        self.assertRaises(MatrixTableSampleSetError, HailMatrixTableTask.subset_samples, mt, self.create_temp_sample_subset_file(mt, 1, True))
+
+    def test_hail_matrix_table_subset_wrong_sample_id_correct(self):
+        task = HailMatrixTableTask(source_paths=[TEST_DATA_MT_1KG], dest_path="empty", genome_version='37')
+        mt = hl.import_vcf(TEST_DATA_MT_1KG)
+        try:
+            task.subset_samples(mt, self.create_temp_sample_subset_file(mt, 1, True))
```
nawatts: This test could incorrectly pass if `task.subset_samples` does not raise an exception.

luigi_pipeline/tests/test_hail_tasks.py
```diff
@@ -90,3 +91,53 @@ def requires(self):
...
+            subset = subset.annotate(test='wrong_sample').key_by('test')
+            subset = subset.drop('s').rename({'test': 's'})
+        subset.export(temp_input_file, header=True)
+        return temp_input_file
+
+    def test_hail_matrix_table_remap_0(self):
+        task = HailMatrixTableTask(source_paths=[TEST_DATA_MT_1KG], dest_path="empty", genome_version='37')
+        mt = hl.import_vcf(TEST_DATA_MT_1KG)
+        remap_mt = task.remap_sample_ids(mt, self.create_temp_sample_remap_file(mt, 0))
+        self.assertEquals(remap_mt.anti_join_cols(mt.cols()).count_cols(), 0)
```
nawatts: `assertEqual` would be preferable here since `assertEquals` is a deprecated alias.

https://docs.python.org/3/library/unittest.html#deprecated-aliases

luigi_pipeline/lib/hail_tasks.py
```diff
@@ -72,6 +81,63 @@ def sample_type_stats(mt, genome_version, threshold=0.3):
...
+        anti_join_ht = subset_ht.anti_join(mt.cols())
+        anti_join_ht_count = anti_join_ht.count()
+        mt_sample_count = mt.cols().count()
+        missing_samples = anti_join_ht.s.collect()
```
nawatts: Can this line be moved under the `if anti_join_ht_count != 0` condition so that it is only run if `missing_samples` is used?

Same comment for `remap_sample_ids`.

luigi_pipeline/tests/test_hail_tasks.py
```diff
@@ -90,3 +91,53 @@ def requires(self):
...
+        task = HailMatrixTableTask(source_paths=[TEST_DATA_MT_1KG], dest_path="empty", genome_version='37')
+        mt = hl.import_vcf(TEST_DATA_MT_1KG)
+        try:
+            task.subset_samples(mt, self.create_temp_sample_subset_file(mt, 1, True))
+        except MatrixTableSampleSetError as e:
+            if e.missing_samples != ['wrong_sample']:
+                self.fail()
+
+
+
```
knguyen142: Nit: Can we remove the extra new lines so there's just one? My editor automatically changes it to one new line, so sometimes it'll show up as diff on a new PR. https://stackoverflow.com/questions/36043061/automatically-add-newline-on-save-in-pycharm

luigi_pipeline/tests/test_hail_tasks.py
knguyen142: Awesome, good to see it worked!

luigi_pipeline/lib/hail_tasks.py
knguyen142: 👍 Also a good idea to keep vars declaration closest to where they're used to limit scope.

luigi_pipeline/tests/test_hail_tasks.py
knguyen142: Looks like you can capture the exception with https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertRaises. Sorry, I didn't think about it yesterday when we were going over it in person.

luigi_pipeline/lib/hail_tasks.py
knguyen142: Another thing you can do is skip creating `anti_join_ht` var. Go directly to `missing_samples`.

```
missing_samples = subset_ht.anti_join(mt.cols()).collect()
```
Then whenever you want the count, use `len(missing_samples)`, which is constant time. When there is no anti-join set, the `collect` call should be minimal. When there is an anti-join set, we save on calling `count()` and `collect()`, and it's more succint.

knguyen142:
@mike-w-wilson Did you test this locally (doing a local run) and on dataproc? I believe the configs have to change, and for dataproc, we would either have to upload the additional remap and subset files to gs, or include them to the --files.

luigi_pipeline/lib/hail_tasks.py
mike-w-wilson: Didn't know that, thanks!

luigi_pipeline/tests/test_hail_tasks.py
mike-w-wilson: Yup, that sounds better.

luigi_pipeline/tests/test_hail_tasks.py
mike-w-wilson: Makes sense, I'll make this change

luigi_pipeline/lib/hail_tasks.py
```diff
@@ -72,6 +81,63 @@ def sample_type_stats(mt, genome_version, threshold=0.3):
             ht_stats['match'] = (ht_stats['matched_count']/ht_stats['total_count']) >= threshold
         return stats
 
+    @staticmethod
+    def subset_samples(mt, subset_path):
+        """
+        Subset the MatrixTable to the provided list of samples
```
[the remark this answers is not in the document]
mike-w-wilson: I think subsetting samples implies subsetting to variants that only those samples have but happy to make the change. subset_samples_and_variants work?

luigi_pipeline/lib/hail_tasks.py
```diff
@@ -72,6 +81,63 @@ def sample_type_stats(mt, genome_version, threshold=0.3):
...
+            raise MatrixTableSampleSetError(
+                f'Only {remap_ht.semi_join(mt.cols()).count()} out of {remap_count} '
+                'remap IDs matched IDs in the variant callset.\n'
+                f'IDs that aren\'t in the callset: {missing_samples}\n'
+                f'All callset sample IDs:{mt.s.collect()}', missing_samples
+            )
+
+        mt = mt.annotate_cols(**remap_ht[mt.s])
+        remap_expr = hl.cond(hl.is_missing(mt.seqr_id), mt.s, mt.seqr_id)
+        mt = mt.annotate_cols(seqr_id=remap_expr).key_cols_by('seqr_id')
```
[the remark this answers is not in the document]
mike-w-wilson: I also want to preserve the original vcf_id for right now so I created a new annotation `vcf_id=mt.s`. Hail doesn't like the `key_by_cols(s=mt.seqr_id)` on the same line as the remap and vcf_id annotation so I've moved the keying to it's own line.

PR #131, opened 2019-04-02
luigi v02 vep and derived fields

knguyen142:
## Summary
V02 VEP and derived fields, done kind of in a rush so please forgive. 

## Features
- ported `get_expr_for_variant_type` for hail utils v02 (https://github.com/macarthur-lab/hail-elasticsearch-pipelines/blob/master/hail_scripts/v01/load_dataset_to_es.py#L461) (not used because it looks like it isn't selected, but added anyway).
- ported `get_expr_for_end_pos` for hail utils v02
- fixed run_vep? Not sure how it was working before.
- fixed get_expr_for_vep_protein_domains_set
- Ported over https://github.com/macarthur-lab/hail-elasticsearch-pipelines/blob/master/hail_scripts/v01/load_dataset_to_es.py#L447-L630 to 40 lines
- Added notion of vep runner so we can specify using the hail default one, or a dummy one (for local).  I intended to have this be the interface for a vep runner with cache class, but this might not be necessary since there might be a more elegant way of memoizing. 

## Testing
- One test for derived fields, scant but mostly sanity checking
- Verified local run and dataproc run work


hail_scripts/v02/utils/computed_fields/vep.py
```diff
@@ -78,7 +78,7 @@ def get_expr_for_vep_gene_ids_set(vep_transcript_consequences_root, only_coding_
 
 def get_expr_for_vep_protein_domains_set(vep_transcript_consequences_root):
     return hl.set(
-        vep_transcript_consequences_root.flatmap(lambda c: c.domains.map(lambda domain: domain.db + ":" + domain.name))
```
knguyen142: Ran into an error, looks like https://github.com/macarthur-lab/hail-elasticsearch-pipelines/blob/6cdd8a66607c02de6fc860290e26fcff190fcf0d/hail_scripts/v02/utils/computed_fields/vep.py#L145 converted it to a string.

luigi_pipeline/tests/test_seqr_loading_tasks.py
```diff
@@ -49,3 +51,21 @@ def test_seqr_loading_validate_wgs_mismatch(self, mock_sample_type_stats):
...
+        rsid = 'rs35471880'
+
+        mt = hl.import_vcf(TEST_DATA_MT_1KG)
+        mt = hl.split_multi(mt.filter_rows(mt.rsid == rsid))
+        mt = mt.annotate_rows(**VEP_DATA[rsid])
+        mt = SeqrVCFToMTTask.derive_fields(mt, 'VARIANTS')
+
+        obj = mt.rows().collect()[0]
+
+        # Cannot do a nested compare because of nested objects, so do one by one.
```
nawatts: I thought [assertEqual](https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertEqual) could handle that? Is it because the objects are Hail structs?

hail_scripts/v02/utils/computed_fields/vep.py
knguyen142: What name do you propose? And would that mean updating the v01 equivalent https://github.com/macarthur-lab/hail-elasticsearch-pipelines/blob/94597bf14272e6ee022edb49c003c21bbc0208ec/hail_scripts/v01/utils/computed_fields.py#L327, as well as the default `vep_transcript_consequences_root` to sorted instead of original?

nawatts: > What name do you propose?

`vep_sorted_transcript_consequences_root` instead of `vep_transcript_consequences_root`, similar to [get_expr_for_vep_gene_id_to_consequence_map](https://github.com/macarthur-lab/hail-elasticsearch-pipelines/blob/6bbc5c6b0a00ab83ed3415d3a6a0a5e5bc57f319/hail_scripts/v02/utils/computed_fields/vep.py#L252) and [get_expr_for_worst_transcript_consequence_annotations_struct](https://github.com/macarthur-lab/hail-elasticsearch-pipelines/blob/6bbc5c6b0a00ab83ed3415d3a6a0a5e5bc57f319/hail_scripts/v02/utils/computed_fields/vep.py#L283).

I think the Hail 0.1 version can be changed separately.

luigi_pipeline/seqr_loading.py
```diff
@@ -18,14 +20,61 @@ class SeqrVCFToMTTask(HailMatrixTableTask):
...
+        """
+        mt = mt.annotate_rows(
+            sortedTranscriptConsequences=vep.get_expr_for_vep_sorted_transcript_consequences_array(mt.vep))
+        mt = mt.annotate_rows(
+            docId=variant_id.get_expr_for_variant_id(mt, 512),
+            variantId=variant_id.get_expr_for_variant_id(mt),
+            contig=variant_id.get_expr_for_contig(mt.locus),
+            pos=variant_id.get_expr_for_start_pos(mt),
+            start=variant_id.get_expr_for_start_pos(mt),
+            end=variant_id.get_expr_for_end_pos(mt),
```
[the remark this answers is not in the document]
knguyen142: Right, this one is supposed to be end, not xstop, right?

luigi_pipeline/seqr_loading.py
```diff
@@ -18,14 +20,61 @@ class SeqrVCFToMTTask(HailMatrixTableTask):
...
+            xstop=variant_id.get_expr_for_xpos(mt.locus) + hl.len(variant_id.get_expr_for_ref_allele(mt))
+        )
+        common_selects = ['docId', 'variantId', 'contig', 'start', 'end', 'ref', 'alt', 'xpos', 'xstart',
+                          'xstop', 'rsid', 'filters', 'a_index', 'geneIds', 'transcriptIds', 'codingGeneIds',
+                          'domains', 'transcriptConsequenceTerms', 'sortedTranscriptConsequences',
+                          'mainTranscript']
+        if dataset_type == 'VARIANTS':
+            mt = mt.select_rows(*common_selects, AC=mt.info.AC, AF=mt.info.AF, AN=mt.info.AN)
+        elif dataset_type == 'SV':
+            mt = mt.select_rows(*common_selects, AC=mt.info.AC, AF=mt.info.AF, AN=mt.info.AN)
```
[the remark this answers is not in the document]
knguyen142: Good catch, I forgot to extract the SV info variants. https://github.com/macarthur-lab/hail-elasticsearch-pipelines/blob/81df07887fced60ad55baef58312223b43f17e5c/hail_scripts/v01/load_dataset_to_es.py#L606

luigi_pipeline/README.md
```diff
@@ -22,10 +22,10 @@ $ LUIGI_CONFIG_PATH=configs/seqr-loading-local.cfg python seqr_loading.py SeqrMT
...
 - Install https://github.com/Nealelab/cloudtools
 ```
-$ cluster start seqr-loading-dev --num-workers 2 --pkgs luigi,google-api-python-client --max-idle 60m
+$ cluster start seqr-loading-dev --num-workers 2 --pkgs luigi,google-api-python-client --max-idle 60m --vep
 ```
 
 ### Run
 ```
-$ cluster submit seqr-loading-dev seqr_loading.py --pyfiles lib --files configs/luigi.cfg --args "SeqrMTToESTask --local-scheduler"
+$ cluster submit seqr-loading-dev seqr_loading.py --pyfiles lib,../hail_scripts --files configs/luigi.cfg --args "SeqrMTToESTask --local-scheduler"
```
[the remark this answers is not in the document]
knguyen142: So the luigi readme is in the luigi_pipeline directory so I think it's safe to assume everything is relative to working directory.

luigi_pipeline/seqr_loading.py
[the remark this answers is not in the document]
nawatts: Then again, I guess there's nothing to be done about this. Since these names are from the Hail 0.1 version, changing them could affect the Seqr application.

luigi_pipeline/tests/test_seqr_loading_tasks.py
knguyen142: Yes, I should put in nested hail objects. It looks like the hail comparator isn't set. I tried getting around that by comparing against converted json sorted on field names, as well as converting to dicts recursively, but didn't get it to work in the time box.

hail_scripts/v02/utils/computed_fields/vep.py
knguyen142: K how about I revert the change and keep the method unchanged, and create a new one with that name for this?

nawatts: Ok. It looks like this change would break load_clinvar_to_es.py, since that calls it with the original VEP consequences.

https://github.com/macarthur-lab/hail-elasticsearch-pipelines/blob/6bbc5c6b0a00ab83ed3415d3a6a0a5e5bc57f319/hail_scripts/v02/load_clinvar_to_es.py#L73

luigi_pipeline/lib/hail_vep_runners.py
```diff
@@ -0,0 +1,557 @@
...
+
+class HailVEPRunner(HailVEPRunnerBase):
+
+    def run(self, mt, genome_version):
+        return hail_utils.run_vep(mt, genome_version)
+
+
+class HailVEPDummyRunner(HailVEPRunnerBase):
+
+    MOCK_VEP_DATA = hl.struct(
```
knguyen142: This is copied from one of the vep annotations in the 1kg variants. Ideally, this would be generic data and labeld as `mock` somehow, but it involves a lot of genetics domain knowledge that I don't know. I'll add to comments.

hail_scripts/v02/utils/computed_fields/vep.py
knguyen142: So create two methods good?

knguyen142: I guess it's also possible to do a check on the type before, I just hate making changes without being able to test against older ones haha.

luigi_pipeline/seqr_loading.py
```diff
@@ -18,14 +20,61 @@ class SeqrVCFToMTTask(HailMatrixTableTask):
...
+            docId=variant_id.get_expr_for_variant_id(mt, 512),
+            variantId=variant_id.get_expr_for_variant_id(mt),
+            contig=variant_id.get_expr_for_contig(mt.locus),
+            pos=variant_id.get_expr_for_start_pos(mt),
+            start=variant_id.get_expr_for_start_pos(mt),
+            end=variant_id.get_expr_for_end_pos(mt),
+            ref=variant_id.get_expr_for_ref_allele(mt),
+            alt=variant_id.get_expr_for_alt_allele(mt),
+            xpos=variant_id.get_expr_for_xpos(mt.locus),
+            xstart=variant_id.get_expr_for_start_pos(mt),
```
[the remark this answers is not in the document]
knguyen142: Good catch, I didn't notice that xpos and pos were different. @mike-w-wilson can you look through this and the original to see if I missed anything else?

luigi_pipeline/seqr_loading.py
```diff
@@ -18,14 +20,61 @@ class SeqrVCFToMTTask(HailMatrixTableTask):
...
+        """
+        Transforms a mt with fields derived from `vep`, `locus`, `alleles`, `position`, etc.
+        :param mt: matrix table with required fields (e.g. `vep`, `locus`, etc.)
+        :param dataset_type: either `VARIANTS` or `SV`
+        :return: mt with selected derived fields
+        """
+        mt = mt.annotate_rows(
+            sortedTranscriptConsequences=vep.get_expr_for_vep_sorted_transcript_consequences_array(mt.vep))
+        mt = mt.annotate_rows(
+            docId=variant_id.get_expr_for_variant_id(mt, 512),
```
mike-w-wilson: Just curious, how was 512 decided? Do we want that hardcoded?

luigi_pipeline/seqr_loading.py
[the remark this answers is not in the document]
mike-w-wilson: `get_expr_for_filtering_allele_frequency` is used in v01/load_dataset_to_es.py. Not sure if you were waiting to define that elsewhere. Otherwise it looks good to me.

luigi_pipeline/seqr_loading.py
knguyen142: So this is copied over, and I assume they chose that based on some ES search string limit? 
My general heuristic is if it's used only once and pretty obvious, it can be a magic number.

luigi_pipeline/seqr_loading.py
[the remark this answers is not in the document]
knguyen142: Yes, good catch. I remembered I left it out because I was confused since I remembered FAF was something in the gnomad fields now. Let's remember this and revisit it when we do the ref data.

PR #134, opened 2019-04-18
v02 genotypes

knguyen142:
# Features
- bug fix for row_annotations decorator (save some state to instance and not class)
- removed SV schema for now until needed
- added multi_annotation boolean to row_annotations decorator so we can support annotating num_samples_gq_i_to_n more easily (I thought of writing each out, which I actually prefer so it's defined, but there are just so many...)

# Testing
- Tested against v01 implementation, where we found discrepancies in missing genotypes discussion https://discuss.hail.is/t/hail-v01-to-v02-missing-genotypes/939 and https://hail.zulipchat.com/. Pending consensus, but for non-missing genotypes, computations match. 
- Wrote tests cases off of one variant in the 1kg30

It's a lot of changes, so I'll only tag people in parts that they might be familiar with.

### DIFF
--- a/luigi_pipeline/lib/model/seqr_mt_schema.py
+++ b/luigi_pipeline/lib/model/seqr_mt_schema.py
@@ -101,21 +101,56 @@ class SeqrVariantSchema(SeqrSchema):
     def an(self):
         return self.mt.info.AN
 
+    @row_annotation()
+    def genotypes(self):
+        return hl.agg.collect(hl.struct(**self._genotype_fields()))
 
-class SeqrSVSchema(SeqrSchema):
+    @row_annotation(fn_require=genotypes)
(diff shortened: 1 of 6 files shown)


luigi_pipeline/lib/model/seqr_mt_schema.py
```diff
@@ -101,21 +101,56 @@ def af(self):
...
-        return self.mt.info.SVTYPE
-
-    @row_annotation(name='SVLEN')
-    def svlen(self):
-        return self.mt.info.SVLEN
-
-    @row_annotation(name='END')
-    def end(self):
-        return self.mt.info.END
+    @row_annotation()
```
knguyen142: @hanars @bw2 can you guys take a look through this and confirm function parity?

PR #139, opened 2019-04-30
v02 loading

knguyen142:
Working wrapper script to load dataproc and ES nodes.

# Changes
- Ported over `load_dataset.py` to `load_dataset_v02.py` with minor changes so we can make progress. Will evaluate whether it's worth it to re-write. (changes usually have `v02` in name of function or in comment. 
  - different functions to spin up and submit to dataproc using cloudtools
  - in v02 all args sent to `load_dataset.py` is sent to `load_dataset_to_es.py`, but with config files in v02, we only send what's needed (es_host and es_port)
- Minor changes
  - added es_host and es_port to es task
  - minor bug fixes

FYI @bw2 @hanars @mike-w-wilson

### DIFF
--- a/luigi_pipeline/README.md
+++ b/luigi_pipeline/README.md
@@ -29,3 +29,10 @@ $ cluster start seqr-loading-dev --num-workers 2 --pkgs luigi,google-api-python-
 ```
 $ cluster submit seqr-loading-dev seqr_loading.py --pyfiles lib,../hail_scripts --files configs/luigi.cfg --args "SeqrMTToESTask --local-scheduler"
 ```
+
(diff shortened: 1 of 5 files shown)


knguyen142:
@bw2 I tried downsizing the instance by changing https://github.com/macarthur-lab/hail-elasticsearch-pipelines/blob/889bd9cbc96d4a95db56972c96904e09e4c489d6/luigi_pipeline/load_dataset_v02.py#L271 to `n1-highmem-2' and `n1-standard-1` but the script got stuck at `es-data[1].is_running = None`. My guess is that it's configured somewhere to only be deployed to highmem-4 or bigger?

PR #141, opened 2019-05-03
Join mega ref, clinvar, and hgmd tables into seqr mt schema

mike-w-wilson:
## Summary 
-Added reference data annotations to seqr loading pipeline which should allow for end to end runs. Tested full runs of SeqrVCFToMTTask locally and on the cloud. Spot checked results but will complete more thorough testing once reference schema is created. 
## Changes
-Added the gold_star annotations to the write_clinvar_ht script and ran it. https://console.cloud.google.com/dataproc/jobs/a85d98ed6a334547959e3a51760a9c98?project=seqr-project&organizationId=548622027621&supportedpurview=project&region=global&duration=P4D 
-Updated ES port parameter from a string to an int
-Removed unnecessary 'AD' field from gnomad genome coverage reference field selection
-Added all reference data annotations to seqr_mt_schema includes: cadd, dbnsfp, eigen, gnomad_exomes, gnomad_exomes_coverage, gnomad_genomes, gnomad_genomes_coverage, g1k, mpc, primate_ai, splice_ai, topmed, hgmd, clinvar
## To Do
-Create unit tests for new schema additions 
-Create reference schema and validate

### DIFF
--- a/luigi_pipeline/seqr_loading.py
+++ b/luigi_pipeline/seqr_loading.py
@@ -18,12 +18,17 @@ class SeqrVCFToMTTask(HailMatrixTableTask):
     """
     Inherits from a Hail MT Class to get helper function logic. Main logic to do annotations here.
     """
-    reference_mt_path = luigi.Parameter(description='Path to the matrix table storing the reference variants.')
+    reference_ht_path = luigi.Parameter(description='Path to the Hail table storing the reference variants.')
+    clinvar_ht_path = luigi.Parameter(description='Path to the Hail table storing the clinvar variants.')
+    hgmd_ht_path = luigi.Parameter(description='Path to the Hail table storing the hgmd variants.')
     sample_type = luigi.ChoiceParameter(choices=['WGS', 'WES'], description='Sample type, WGS or WES', var_type=str)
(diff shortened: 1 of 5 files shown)


luigi_pipeline/seqr_loading.py
```diff
@@ -36,58 +41,63 @@ def run(self):
...
-        sample_type_stats = HailMatrixTableTask.sample_type_stats(mt, genome_version)
-
-        for name, stat in sample_type_stats.items():
-            logger.info('Table contains %i out of %i common %s variants.' %
-                        (stat['matched_count'], stat['total_count'], name))
-
-        has_coding = sample_type_stats['coding']['match']
-        has_noncoding = sample_type_stats['noncoding']['match']
+@staticmethod
+def validate_mt(mt, genome_version, sample_type):
```
knguyen142: Why is this unindented? Looks like it's removed from the class, I don't think `self.validate_mt` above would work, and makes sense why it wouldn't fail since we disable validation?

mike-w-wilson: Sorry, this was a mistake. Fixed.

mike-w-wilson:
Thanks, I've added the ref tests to Asana. All fields match the VDS output.

PR #144, opened 2019-06-04
Test ref data for 1kg_30variants dataset

mike-w-wilson:
clinvar.GRCh37.ht and combined_reference_data_grch37.ht subsetted to variants only in 1kg_30variants for local testing


knguyen142:
I skipped through this, sorry! Can you modify the local luigi configs and check if it works?

https://github.com/macarthur-lab/hail-elasticsearch-pipelines/blob/master/luigi_pipeline/configs/seqr-loading-local.cfg#L10

https://github.com/macarthur-lab/hail-elasticsearch-pipelines/blob/master/luigi_pipeline/configs/optimized_configs/seqr-loading-local.cfg (NOTE: I'm about to remove the reference_ht paths here in a PR so they'll have to be re-added.

bw2:
is this PR still relevant?

PR #145, opened 2019-06-19
Add gnomad 38 ref data

mike-w-wilson:
# Summary
-Added  geno2mp to 37 and 38 ref datasets to config
-Added gnomad 38 liftover and gnomad coverage to 38 configs

### DIFF
--- a/download_and_create_reference_datasets/v02/hail_scripts/write_combined_reference_data_ht.py
+++ b/download_and_create_reference_datasets/v02/hail_scripts/write_combined_reference_data_ht.py
@@ -1,3 +1,4 @@
+import argparse
 from datetime import datetime
 from functools import reduce
 import logging
@@ -30,7 +31,7 @@ CONFIG = {
             'field_name': 'g1k',
         },
         '38': {
-            'path': 'gs://seqr-reference-data/GRCh38/1kg/1kg.wgs.phase3.20170504.GRCh38_sites.ht',
+            'path': 'gs://seqr-reference-data/GRCh38/1kg/1kg.wgs.phase3.20170504.GRCh38_sites_test.ht',
             'select': {'AC': 'info.AC#', 'AF': 'info.AF#', 'AN': 'info.AN', 'POPMAX_AF': 'POPMAX_AF'},
             'field_name': 'g1k',
         },
@@ -133,7 +134,7 @@ CONFIG = {
             'custom_select': 'custom_gnomad_select'
         },
         '38': {
-            'path': 'gs://gnomad-public/release/2.1.1/liftover_grch38/ht/exomes/gnomad.exomes.r2.1.1.sites.liftover_grch38.ht',
+            'path': 'gs://seqr-reference-data/GRCh38/gnomad/gnomad.exomes.r2.1.1.sites.liftover_grch38.ht',
             'custom_select': 'custom_gnomad_select'
         }
     },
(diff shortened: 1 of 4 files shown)


knguyen142:
@mike-w-wilson can you provide details on what tables were generated and which still need to be done?

mike-w-wilson:
Sorry @knguyen142 , the datasets generated were:
- Geno2MP for 37 and 38
- gnomAD coverage for 38, exomes and genomes
- gnomAD 38, exomes and genomes
We still need to generate 1kg for 38. The keys are out of order, which prevents write_combined_reference_data_ht.py from running successfully. I am creating that table now but that will be in another PR.

knguyen142:
Ahh, does it make sense to wait and confirm that the combined ref data can be written before merging that file?

mike-w-wilson:
That 1kg file referenced here is was already merged back in February but shouldn't have been as it is out of order. The file name won't change in the rewrite of 1kg. This PR only focuses on gnomAD and Geno2MP but if you want to wait, I can change the summary of this ticket to include 1kg corrections. The write_1kg_ht.py script will also need to be updated.

download_and_create_reference_datasets/v02/hail_scripts/write_combined_reference_data_ht.py
```diff
@@ -110,34 +110,65 @@
...
+            'path': 'gs:/gnomad-public/release/2.1.1/liftover_grch38/ht/genomes/gnomad.genomes.r2.1.1.sites.liftover_grch38.ht',
+            'custom_select': 'custom_gnomad_select'
+        }
     },
     'exac': {
         '37': {
             'path': 'gs://seqr-reference-data/GRCh37/gnomad/ExAC.r1.sites.vep.ht',
             'select': {'AF_POPMAX': 'info.AF_POPMAX', 'AF': 'info.AF#', 'AC_Adj': 'info.AC_Adj#', 'AC_Het': 'info.AC_Het#',
-                       'AC_Hom': 'info.AC_Hom#', 'AC_Hemi': 'info.AC_Hemi#', 'AN_Adj': 'info.AN_Adj',},
+                       'AC_Hom': 'info.AC_Hom#', 'AC_Hemi': 'info.AC_Hemi#', 'AN_Adj': 'info.AN_Adj'},
```
knguyen142: No biggie but sometimes you'll see trailing comma https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Trailing_commas

download_and_create_reference_datasets/v02/hail_scripts/write_combined_reference_data_ht.py
```diff
@@ -241,11 +272,14 @@ def join_hts(datasets, coverage_datasets=[], reference_genome='37'):
     joined_ht.write(os.path.join(output_path))
 
 def run():
-    # TODO: '38' when gnomad liftover is done.
```
knguyen142: Might make sense to have the user specify which genome version. Usually we use args, but we can consider reading in input, depends on what you think is better ux https://www.geeksforgeeks.org/taking-input-in-python/.

download_and_create_reference_datasets/v02/hail_scripts/write_combined_reference_data_ht.py
mike-w-wilson: That is good to know and makes sense.

download_and_create_reference_datasets/v02/hail_scripts/write_combined_reference_data_ht.py
mike-w-wilson: I'm good with using argparse if you are? I like the idea of kicking it off and not worrying about it. Plus, I think user input into dataproc is tricky.

mike-w-wilson:
@knguyen142 , I added in the option of choosing builds in the combine_ref scripts using argparse.

PR #147, opened 2019-07-08
Updated docId and null write fields in ES index

mike-w-wilson:
## Summary
-Added docId as ES mapping ID to get rid of duplicate docs in index
-ES will skip writing  null fields so added write_null_es_fields arg to write out doc fields that have null values
-Updated error message for subsetting. Originally was printing out the number of samples in the vcf instead of the number of subset samples. This has been fixed. 
-Addressed glob patterns for importing vcfs so luigi requires will not fail

### DIFF
--- a/hail_scripts/v02/utils/elasticsearch_client.py
+++ b/hail_scripts/v02/utils/elasticsearch_client.py
@@ -29,7 +29,7 @@ class ElasticsearchClient(BaseElasticsearchClient):
         delete_index_before_exporting :bool = True,
         elasticsearch_write_operation :str = ELASTICSEARCH_INDEX,
         ignore_elasticsearch_write_errors :bool = False,
-        elasticsearch_mapping_id=None,
+        elasticsearch_mapping_id :str = "docId",
         field_name_to_elasticsearch_type_map=None,
         disable_doc_values_for_fields=(),
         disable_index_for_fields=(),
@@ -37,6 +37,7 @@ class ElasticsearchClient(BaseElasticsearchClient):
         func_to_run_after_index_exists=None,
         export_globals_to_index_meta=True,
         verbose=True,
+        write_null_es_fields :bool = False,
     ):
         """Create a new elasticsearch index to store the records in this table, and then export all records to it.
 
@@ -85,6 +86,7 @@ class ElasticsearchClient(BaseElasticsearchClient):
                 (see https://www.elastic.co/guide/en/elasticsearch/reference/current/mapping-meta-field.html)
             child_table (Table): if not None, records in this Table will be exported as children of records in the main Table.
             verbose (bool): whether to print schema and stats
+            write_null_es_fields (bool): whether to write fields that are null to the index
         """
 
         elasticsearch_config = {}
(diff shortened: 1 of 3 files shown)


hail_scripts/v02/utils/elasticsearch_client.py
```diff
@@ -29,14 +29,15 @@ def export_table_to_elasticsearch(
         delete_index_before_exporting :bool = True,
         elasticsearch_write_operation :str = ELASTICSEARCH_INDEX,
         ignore_elasticsearch_write_errors :bool = False,
-        elasticsearch_mapping_id=None,
+        elasticsearch_mapping_id :str = "docId",
```
nawatts: This should be set in Seqr loading code instead of changing the default. This is also used for loading gnomAD datasets, none of which have a `docId` field.

hail_scripts/v02/utils/elasticsearch_client.py
```diff
@@ -29,14 +29,15 @@ def export_table_to_elasticsearch(
...
         field_name_to_elasticsearch_type_map=None,
         disable_doc_values_for_fields=(),
         disable_index_for_fields=(),
         field_names_replace_dot_with="_",
         func_to_run_after_index_exists=None,
         export_globals_to_index_meta=True,
         verbose=True,
+        write_null_es_fields :bool = False,
```
nawatts: Nit, but I think `write_null_values` would be a bit more descriptive.

hail_scripts/v02/utils/elasticsearch_client.py
knguyen142: Beat me to it haha.

luigi_pipeline/lib/hail_tasks.py
```diff
@@ -41,7 +41,8 @@ class HailMatrixTableTask(luigi.Task):
                                                                                             'to annotate vep.')
 
     def requires(self):
-        return [VcfFile(filename=s) for s in self.source_paths]
+        # We need to exclude globs in source path since they aren't files.
+        return [VcfFile(filename=s) for s in self.source_paths if '*' not in s]
```
nawatts: When does a glob get passed as a source path? If globs won't work, it seems better to raise an error or let the task fail by creating the VcfFile instead of silently ignoring them.

knguyen142: Globs do work, the requires is used to check whether the files exist. So the param is source_paths and we only check that non globs exist, but globs are still used later down.
Sorry the comment was mine, it's probably confusing. @mike-w-wilson can you change it to specify that it's excluded only for checking that it exists?

mike-w-wilson: Some callsets get delivered sharded such as 

> gs://seqr-datasets/GRCh38/RDG_WGS_Broad_Internal/v6/sharded_vcf/RDG_WGS_Broad_Internal_v5.filtered.*.vcf.gz 

Where the glob the shard number. @Kevin what do you think of Nick's suggestion?

knguyen142: @mike-w-wilson I think @nawatts misunderstood that the globs were not being used. We changed https://github.com/macarthur-lab/hail-elasticsearch-pipelines/pull/147/files/f693d335c21b552eafd9595f6057312f0ce6eb2a#diff-9e4438316786f665992553279456402cL57 here instead of using `input()` (which is derived from `requires()` to go straight to the source_paths which has the globs.
Thus, the logic is source_paths can have globs, but they will not be in the requires because they cannot be checked to exist. I mean, if we're sure *'s are only used at the file level, we can extract the directory and make sure that it exists, but I'm not sure if *'s can be used in dirs as well.

mike-w-wilson: I see. Updated the comment.

PR #150, opened 2019-07-15
Kev/variant table optimization

knguyen142:
This adds the optimized run. 

## Unoptimized
**For each project**
VCF --> Project MT with variant annotations and genotype annotations --> ES

Since 1 VCF has multiple projects, this means that variants are annotated multiple times.

## Optimized
**Once for the whole VCF**
VCF --> HT with variant annotations
**For each project** 
VCF --> project MT with only genotype annotations 
Join project genotypes MT with variant HT and export to ES.

We were thinking of doing optimizations where we would only do variant annotations on new variants added since the last VCF, but surprisingly this anti-join and union takes longer than just doing it from scratch (https://docs.google.com/spreadsheets/d/1mIPDBrpB-_-24oMC_oLby6gRTqoaC4hVEMAfaqm1NX8/edit?usp=drive_web&ouid=101134808925953712603). Will evaluate this in the future, but let's do it one step at a time.

# Changes
1. Separate configs for optimized
2. Separated the schemas so we have variants only and genotypes only. The unoptimized uses a schema class that inherits from both of these, effectively unioning them.
3. Created a new script for the optimized version that leverages the classes before it so it's more of separating and re-ordering steps. The power of decomposition!

# Testing
This was tested in the cloud (new config changes not tested) and locally. It does not break any old unit tests, but at the same time does not add any new ones yet :(

Run as before, except with new script and task names e.g. `...seqr_loading_optimized.py... SeqrMTToESOptiimzedTask --local-scheduler --es-host 10.12.1.3 --es-port 9200`


luigi_pipeline/seqr_loading_optimized.py
```diff
@@ -0,0 +1,86 @@
...
+    def requires(self):
+        return [SeqrVCFToVariantHTTask(), SeqrVCFToGenotypesMTTask()]
+
+    def run(self):
+        genotypes_mt = hl.read_matrix_table(self.input()[1].path)
+        variants_ht = hl.read_table(self.input()[0].path)
+        row_ht = genotypes_mt.rows().join(variants_ht)
+
+        row_ht = row_ht.drop('vep').flatten()
+        row_ht = row_ht.drop(row_ht.locus, row_ht.alleles).annotate(type='optimized')
```
nawatts: `SeqrVariantSchema.elasticsearch_row` also drops vep, flattens, and drops locus and alleles. Why does it need to be done both here and there?

luigi_pipeline/lib/model/seqr_mt_schema.py
```diff
@@ -169,6 +170,27 @@ def af(self):
...
+
+        TODO:
+        - Call validate
+        - when all annotations are here, whitelist fields to send instead of blacklisting.
+        :return:
+        """
+        # Converts nested structs into one field, e.g. {a: {b: 1}} => a.b: 1
+        table = self.mt.rows().drop('vep').flatten()
+        # When flattening, the table is unkeyed, which causes problems because our locus and alleles should not
+        # be normal fields. We can also re-key, but I believe this is computational?
```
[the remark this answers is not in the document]
knguyen142: Yeah I think that's why.

luigi_pipeline/seqr_loading_optimized.py
knguyen142: Because of bad organization and rushing on my part. That logic should move somewhere that is shared. I think it should probably go to `SeqrVariantsAndGenotypesSchema` and then we can do `SeqrVariantsAndGenotypesSchema(mt).elasticsearch_row()`.

luigi_pipeline/seqr_loading_optimized.py
```diff
@@ -0,0 +1,86 @@
...
+import re
+
+import luigi
+import hail as hl
+
+from hail_scripts.v02.utils.elasticsearch_client import ElasticsearchClient
+from lib.hail_tasks import HailMatrixTableTask, HailElasticSearchTask, GCSorLocalTarget, MatrixTableSampleSetError
+from lib.model.seqr_mt_schema import SeqrSchema, SeqrVariantSchema, SeqrGenotypesSchema
+import seqr_loading
+# from gnomad_hail.utils import vep_or_lookup_vep
```
nawatts: This looks like it's left over from development.

luigi_pipeline/seqr_loading_optimized.py
```diff
@@ -0,0 +1,86 @@
...
+        clinvar = hl.read_table(self.clinvar_ht_path)
+        hgmd = hl.read_table(self.hgmd_ht_path)
+
+        ht = SeqrVariantSchema(mt, ref_data=ref_data, clinvar_data=clinvar, hgmd_data=hgmd).annotate_all(
+            overwrite=True).select_annotated_mt().rows()
+
+        ht.describe()
+
+        ht.write(self.output().path, stage_locally=True)
+        hl.copy_log(self.output().path + '-logs.log')
```
nawatts: Is this always wanted? Or should it be behind some kind of debug/verbose flag?

knguyen142: So this is something that I haven't come to a conclusion on. It's actually nice to have these logs because in the past when we've had bad run times, the hail team has asked for logs and we have to re-run. What I'm not sure about is the convention, it seems like if we tie it to the bucket that the job creates, that'd be the ideal location. But I don't think it's readily accessible from the code.

nawatts: 👍maybe worth creating a bucket dedicated to logs so that data buckets don't get cluttered.

luigi_pipeline/seqr_loading.py
```diff
@@ -45,11 +44,13 @@ def run(self):
         clinvar = hl.read_table(self.clinvar_ht_path)
         hgmd = hl.read_table(self.hgmd_ht_path)
 
-        mt = SeqrVariantSchema(mt, ref_data=ref_data, clinvar_data=clinvar, hgmd_data=hgmd).annotate_all(
+        mt = SeqrVariantsAndGenotypesSchema(mt, ref_data=ref_data, clinvar_data=clinvar, hgmd_data=hgmd).annotate_all(
             overwrite=True).select_annotated_mt()
 
+        mt = SeqrGenotypesSchema(mt).annotate_all(overwrite=True).mt
```
knguyen142: @knguyen142 this should not be needed.

luigi_pipeline/seqr_loading_optimized.py
```diff
@@ -0,0 +1,84 @@
...
+import hail as hl
+
+from lib.hail_tasks import HailMatrixTableTask, HailElasticSearchTask, GCSorLocalTarget, MatrixTableSampleSetError
+from lib.model.seqr_mt_schema import SeqrSchema, SeqrVariantSchema, SeqrGenotypesSchema, SeqrVariantsAndGenotypesSchema
+import seqr_loading
+
+logger = logging.getLogger(__name__)
+
+
+class SeqrValidationError(Exception):
```
knguyen142: Kev remove this.

luigi_pipeline/seqr_loading_optimized.py
```diff
@@ -0,0 +1,86 @@
...
+        ht.write(self.output().path, stage_locally=True)
+        hl.copy_log(self.output().path + '-logs.log')
+
+class SeqrVCFToGenotypesMTTask(HailMatrixTableTask):
+    remap_path = luigi.OptionalParameter(default=None,
+                                         description="Path to a tsv file with two columns: s and seqr_id.")
+    subset_path = luigi.OptionalParameter(default=None,
+                                          description="Path to a tsv file with one column of sample IDs: s.")
+    def run(self):
+        mt = self.import_vcf()
```
mike-w-wilson: Should we run this off the MT from the SeqrVCFToVariantHTTask instead of importing and splitting the vcf again?

PR #154, opened 2019-08-02
missing annotations

knguyen142:
The long lost forgotten annotations, thanks to @mike-w-wilson validation. 

The important thing to note is `split_multi_hts` drops `old_locus` and `old_alleles` so we might see if we can remove that https://github.com/hail-is/hail/blob/e40fb8901ff40f3e29660cf4ef1193cd2be0f116/hail/python/hail/methods/statgen.py#L2251. In the mean time, we preannotate before splitting to generate the original alt alleles.

### DIFF
--- a/luigi_pipeline/lib/model/seqr_mt_schema.py
+++ b/luigi_pipeline/lib/model/seqr_mt_schema.py
@@ -18,6 +18,24 @@ class SeqrSchema(BaseMTSchema):
     def vep(self):
         return self.mt.vep
 
+    @row_annotation()
+    def rsid(self):
+        return self.mt.rsid
+
+    @row_annotation()
+    def filters(self):
+        return self.mt.filters
+
+    @row_annotation()
+    def aIndex(self):
+        return self.mt.a_index
+
+    @row_annotation()
+    def originalAltAlleles(self):
+        # TODO: This assumes we annotate `locus_old` in this code because `split_multi_hts` drops the proper `old_locus`.
+        # If we can get it to not drop it, we should revert this to `old_locus`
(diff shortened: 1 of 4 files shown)


luigi_pipeline/lib/model/seqr_mt_schema.py
```diff
@@ -18,6 +18,24 @@ def __init__(self, *args, ref_data, clinvar_data, hgmd_data, **kwargs):
...
+    @row_annotation()
+    def rsid(self):
+        return self.mt.rsid
+
+    @row_annotation()
+    def filters(self):
+        return self.mt.filters
+
+    @row_annotation()
+    def aIndex(self):
```
knguyen142: @hanars named aIndex for conformity, but it is unusual since everything else is snake case. Do you use this?

hanars: This is not used, we can just get rid of it

knguyen142: Sweet, we'll make note of that, but will keep it here for now in case we do any strict comparison validations.

knguyen142:
@nawatts  I also edited an incorrect calculation I did earlier here https://github.com/macarthur-lab/hail-elasticsearch-pipelines/pull/154/commits/52fc0f7c522fce597898bce85f9e56294019fe49 which is in v01 https://github.com/macarthur-lab/hail-elasticsearch-pipelines/blob/29051791d36b4cb7c54f066f6923959813f7f773/hail_scripts/v01/utils/computed_fields.py#L422. Thanks for catching Mike.

PR #157, opened 2019-08-07
Change arg for persistent node creation to separate script

hanars:


gcloud_dataproc/create_persistent_es_nodes.py
```diff
@@ -0,0 +1,14 @@
+import argparse
+
+from load_dataset import _get_es_node_settings, _create_persistent_es_nodes
```
knguyen142: I think https://github.com/macarthur-lab/hail-elasticsearch-pipelines/blob/0e576f71547d5ba6243a5ba0768537d47b87869a/gcloud_dataproc/load_dataset.py#L108-L147 should actually be moved to this file. The shorter we can make the load_data.py files the better.

PR #159, opened 2019-08-16
exit status code

knguyen142:
Solved the problem where unsuccessful runs were returning a 0 (successful) exit code so dataproc thinks it's a success. 

e.g. https://console.cloud.google.com/dataproc/jobs/pyspark_export_project_test_proj_20190815_7e8d25a1?organizationId=548622027621&project=seqr-project&supportedpurview=project&region=global

```
===== Luigi Execution Summary =====

Scheduled 3 tasks of which:
* 2 complete ones were encountered:
    - 1 SeqrVCFToGenotypesMTTask(...)
    - 1 SeqrVCFToVariantMTTask(...)
* 1 failed:
    - 1 SeqrMTToESOptimizedTask(source_path=, use_temp_loading_nodes=True, es_host=10.36.1.2, es_port=9200, es_index=test_proj__WES__grch37__variants__vcfv1__20190815)

This progress looks :( because there were failed tasks

===== Luigi Execution Summary =====
```

Had to dig deep into the code base but turns out `run()` returns False on Failure https://github.com/spotify/luigi/blob/0819a5fee065b6e9661a550ef288d4bf00cf3cd1/luigi/interface.py#L195, so this exits with a non-zero exit code on failure.

## Testing
https://console.cloud.google.com/dataproc/jobs/60b4aae302074e7699f54a2b13c065c3?organizationId=548622027621&project=seqr-project&supportedpurview=project&region=global&duration=PT6H isn't it beautiful?

### DIFF
--- a/luigi_pipeline/seqr_loading.py
+++ b/luigi_pipeline/seqr_loading.py
@@ -1,4 +1,5 @@
 import logging
+import sys
 
 import luigi
 import hail as hl
@@ -130,4 +131,5 @@ class SeqrMTToESTask(HailElasticSearchTask):
 
 
 if __name__ == '__main__':
-    luigi.run()
+    # If run does not succeed, exit with 1 status code.
+    luigi.run() or sys.exit(1)
--- a/luigi_pipeline/seqr_loading_optimized.py
+++ b/luigi_pipeline/seqr_loading_optimized.py
@@ -1,5 +1,6 @@
 import logging
 import os
+import sys
 import re
 
 import luigi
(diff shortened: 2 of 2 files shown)


luigi_pipeline/seqr_loading.py
```diff
@@ -130,4 +130,5 @@ def run(self):
 
 
 if __name__ == '__main__':
-    luigi.run()
+    # If run does not succeed, exit with 1 status code.
+    luigi.run() or exit(1)
```
nawatts: I think `sys.exit` would be preferable here.

https://docs.python.org/3/library/constants.html#constants-added-by-the-site-module

knguyen142: Ahh yes I had `from sys import exit` but it was removed when I reset the branch and forgot to add it. Good catch.

PR #160, opened 2019-08-19
disable splicing predictions

bw2:
- switch to using latest loftee version from https://github.com/konradjk/loftee 
- add `run_splice_predictions:0` to significantly decrease VEP runtime


hanars:
Do we not use VEP's splice predictors?

bw2:
They might show up occasionally among other LOFTEE flags. Do you want to check if any users currently rely on LOFTEE splice-related flags? 
Also, the SpliceAI score provides similar info for SNPs .

PR #162, opened 2019-08-21
introduce row annotation skip and make hgmd optional

knguyen142:
# Description
We want to make hgmd optional to run locally per @bw2's request.

# Implementation
- Introduces a RowAnnotationOmit exception (inspired by Airflow's https://github.com/apache/airflow/blob/master/airflow/exceptions.py#L73) that can be raised within an annotations function to skip an annotation altogether. 
- We omit the hgmd annotation altogether if no hgmd dataset is provided

### DIFF
--- a/luigi_pipeline/lib/model/base_mt_schema.py
+++ b/luigi_pipeline/lib/model/base_mt_schema.py
@@ -6,7 +6,7 @@ import logging
 logger = logging.getLogger(__name__)
 
 
-class RowAnnotationSkip(Exception):
+class RowAnnotationOmit(Exception):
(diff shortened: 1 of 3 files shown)


luigi_pipeline/lib/model/base_mt_schema.py
```diff
@@ -6,6 +6,10 @@
 logger = logging.getLogger(__name__)
 
 
+class RowAnnotationSkip(Exception):
```
knguyen142: Should this be `RowAnnotationOmit`?

PR #167, opened 2019-09-03
min num shards

knguyen142:
# Description
Added logic for the number of shards: greater of user specified min_num_shards and a calculated value based on `(variants*samples)/2147483519` translated to `int((mt.count_rows() * mt.count_cols())/2147483519)`. 

NOTE: I don't know where `2147483519` comes from, it was just given to me, it's slightly less than 2^31. Should I use `(2^31 - 2^10) = 2147482624 ` just to make sure we have a larger num shards? 

# Testing
I tested this locally on the 1kg30 dataset (which as expected calculated the shards to be 0 since the data is so small), but have not run that formula on actual data like RGP. @mike-w-wilson we should do that to make sure that formula gives us what we expect. 

# Add on
Piggybacked on this PR to add the hail version to the global annotations --> ES metadata.

# TODO
Optional, but make the min_num_shards configurable in Airflow (I seriously doubt it will be used much though if the formula is accurate).

### DIFF
--- a/luigi_pipeline/lib/hail_tasks.py
+++ b/luigi_pipeline/lib/hail_tasks.py
@@ -169,6 +169,9 @@ class HailElasticSearchTask(luigi.Task):
     es_host = luigi.Parameter(description='ElasticSearch host.', default='localhost')
     es_port = luigi.IntParameter(description='ElasticSearch port.', default=9200)
     es_index = luigi.Parameter(description='ElasticSearch index.', default='data')
+    es_index_min_num_shards = luigi.IntParameter(default=6,
+                                                 description='Number of shards for the index will be the greater of '
+                                                             'this value and a calculated value based on the matrix.')
 
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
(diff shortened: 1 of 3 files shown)


mike-w-wilson:
2147483519 is the ES limit on docs per index. We get that number from ES error messages. I found this online. 

> This is due to the shard exceeding the Lucene limitation on max # of documents per Lucene index. Lucene has a max # of documents per Lucene index limitation. As of LUCENE-5843 94, the limit is 2,147,483,519 (= Integer.MAX_VALUE - 128).

luigi_pipeline/lib/hail_tasks.py
```diff
@@ -223,3 +227,9 @@ def route_index_to_temp_es_cluster(self, to_temp_loading):
 
         index_arg = "{}*".format(self.es_index)
         self._es.es.indices.put_settings(index=index_arg, body=body)
+
+    def _mt_num_shards(self, mt):
+        # The greater of the user specified min shards and calculated based on the variants and samples
+        denominator = 2147483519
+        calculated_num_shards = int((mt.count_rows() * mt.count_cols())/denominator)
```
knguyen142: TODO: Should always round up, e.g. 10.23 should round up to 11.

knguyen142:
Changed 2 Billion to 1.4 Billion to be safe.

PR #172, opened 2019-10-21
Convert struct fields in table globals to dict

nawatts:
When `export_globals_to_index_meta` is set to true, `export_table_to_elasticsearch` adds the table's globals to the Elasticsearch mapping in the `_meta` field.

However, if one of the global fields contains a struct value, creating the Elasticsearch index  / updating the mapping will fail because the Elasticsearch client can not convert a [hail.utils.Struct](https://hail.is/docs/0.2/utils/index.html#hail.utils.Struct) object to JSON.

This change recurses through the globals struct and converts any struct fields to dicts, which the Elasticsearch client can handle.

### DIFF
--- a/hail_scripts/v02/utils/elasticsearch_client.py
+++ b/hail_scripts/v02/utils/elasticsearch_client.py
@@ -18,6 +18,10 @@ from hail_scripts.v02.utils.elasticsearch_utils import elasticsearch_schema_for_
 logger = logging.getLogger()
 
 
+def struct_to_dict(struct):
+    return {k: dict(struct_to_dict(v)) if isinstance(v, hl.utils.Struct) else v for k, v in struct.items()}
+
+
 class ElasticsearchClient(BaseElasticsearchClient):
     def export_table_to_elasticsearch(
         self,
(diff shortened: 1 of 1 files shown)


hail_scripts/v02/utils/elasticsearch_client.py
```diff
@@ -165,7 +169,7 @@ def export_table_to_elasticsearch(
 
         _meta = None
         if export_globals_to_index_meta:
-            _meta = dict(hl.eval(table.globals))
+            _meta = struct_to_dict(hl.eval(table.globals))
```
knguyen142: What about `json.loads(hl.eval(hl.json(table.globals))))`? I tried to find a native hail to python object but this is the closest I could find. 

This would allow us to use other hail expressions as well.

Tested with
`json.loads(hl.eval(hl.json(hl.struct(x=2))))` --> `{'x': 2}`
and
`json.loads(hl.eval(hl.json(hl.set([2]))))` --> `[2]`

nawatts: One downside to that is how `hl.json` encodes dicts.

```
hl.eval(hl.json({"foo": 1, "bar": 2}))
--> '[{"key":"bar","value":2},{"key":"foo","value":1}]'
```

knguyen142: Write but table.globals will not have dicts, right?

nawatts: The gnomAD variants Hail tables' globals contain dicts.

bw2:
I'm not seeing any downside to this change. Should we merge this?

PR #174, opened 2019-10-28
added gnomad v3 to ref data

knguyen142:
Replaces ref 38 gnomad genome with v3 instead of v2.

- separates selector function `custom_gnomad_select_v2` and `custom_gnomad_select_v3`
- fixes 1kg path `1kg.wgs.phase3.20170504.GRCh38_sites_test.ht` to `1kg.wgs.phase3.20170504.GRCh38_sites.ht`
- writes the ht to a separate directory `v2` and versions the file instead of overwriting
- adds to globals (version as well as adds src paths to the dataset so we can see which version of gnomad is used in that ref data)

Checked that `describe` works, if everyone approves will kick this off. Once this is written, we just need to update the ref data location https://github.com/macarthur-lab/All-The-Envs/blob/master/airflow/dags/seqr-vcf-to-es.py#L56

FYI @mike-w-wilson when you get back

### DIFF
--- a/download_and_create_reference_datasets/v02/hail_scripts/write_combined_reference_data_ht.py
+++ b/download_and_create_reference_datasets/v02/hail_scripts/write_combined_reference_data_ht.py
@@ -6,9 +6,9 @@ import os
 
 import hail as hl
 
-
+VERSION = '2.0.1'
 OUTPUT_TEMPLATE = 'gs://seqr-reference-data/GRCh{genome_version}/' \
-                  'all_reference_data/combined_reference_data_grch{genome_version}.ht'
+                  'all_reference_data/v2/combined_reference_data_grch{genome_version}-{version}.ht'
 
 '''
 Configurations of dataset to combine. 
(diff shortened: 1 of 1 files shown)


download_and_create_reference_datasets/v02/hail_scripts/write_combined_reference_data_ht.py
```diff
@@ -205,6 +206,28 @@ def custom_gnomad_select(ht):
...
+
+def custom_gnomad_select_v3(ht):
+    """
+    Custom select for public gnomad dataset (which we did not generate). Extracts fields like
+    'AF', 'AN', and generates 'hemi'.
+    :param ht: hail table
+    :return: select expression dict
+    """
+    selects = {}
+    global_idx = hl.eval(ht.globals.freq_index_dict['adj'])
```
knguyen142: @nawatts and @lfrancioli  can you check that these are correct with the v2 equivalent https://github.com/macarthur-lab/hail-elasticsearch-pipelines/pull/174/files#diff-35b65d916f1bb6cb19d59b94cde1478dR189

PR #187, opened 2019-12-12
update ES index mechanism

NLSVTN:
First, rather raw adapted scripting for ES index update. Further code clean up may be needed: differentiation between GRCh37 and GRCh38, additional config files, etc. The script can be run the following way:

`PYTHONPATH=/opt/seqr/hail-elasticsearch-pipelines/luigi_pipeline:$PYTHONPATH python -u gcloud_dataproc/submit.py  --cpu-limit 4 --num-executors 3 --driver-memory 2g --executor-memory=4g --hail-version 0.2 --run-locally hail_scripts/v02/update_reference_data_in_existing_index.py --index-name $INDEX_NAME --genome-version 37 --update-hgmd --host $ES_HOST --port $ES_PORT`

### DIFF
--- a/hail_scripts/v02/update_models/update_mt_schema.py
+++ b/hail_scripts/v02/update_models/update_mt_schema.py
@@ -1,0 +1,53 @@
+import hail as hl
+
+from lib.model.base_mt_schema import BaseMTSchema, row_annotation, RowAnnotationOmit
+from hail_scripts.v02.utils.computed_fields import variant_id
+
+class UpdateSchema(BaseMTSchema):
+    @row_annotation(name='docId')
+    def doc_id(self, length=512):
+        return variant_id.get_expr_for_variant_id(self.mt, length)
+
+    @row_annotation()
+    def aIndex(self):
+        return self.mt.a_index
+
+class HGMDSchema(UpdateSchema):
+
+    def __init__(self, *args, hgmd_data, **kwargs):
+        super().__init__(*args, **kwargs)
+        self._hgmd_data = hgmd_data
+
+    @row_annotation()
+    def hgmd(self):
+        if self._hgmd_data is None:
+            raise RowAnnotationOmit
+        return hl.struct(**{'accession': self._hgmd_data[self.mt.row_key].rsid,
+                            'class': self._hgmd_data[self.mt.row_key].info.CLASS})
+
+
+class CLINVARSchema(UpdateSchema):
+
+    def __init__(self, *args, clinvar_data, **kwargs):
+        super().__init__(*args, **kwargs)
+        self._clinvar_data = clinvar_data
+
+    @row_annotation()
+    def clinvar(self):
+        return hl.struct(**{'allele_id': self._clinvar_data[self.mt.row_key].info.ALLELEID,
+                            'clinical_significance': hl.delimit(self._clinvar_data[self.mt.row_key].info.CLNSIG),
+                            'gold_stars': self._clinvar_data[self.mt.row_key].gold_stars})
+
+
+class CIDRSchema(UpdateSchema):
+
+    def __init__(self, *args, cidr_data, **kwargs):
+        super().__init__(*args, **kwargs)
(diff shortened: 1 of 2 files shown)


bw2:
Thanks for the PR!

hail_scripts/v02/update_reference_data_in_existing_index.py
```diff
@@ -0,0 +1,142 @@
...
+
+import hail as hl
+from hail_scripts.v02.update_models.update_mt_schema import HGMDSchema, CLINVARSchema, CIDRSchema
+from hail_scripts.v02.utils.elasticsearch_client import ElasticsearchClient
+from lib.model.seqr_mt_schema import SeqrVariantsAndGenotypesSchema
+
+logger = logging.getLogger(__name__)
+
+# Change the paths to your hadoop/gcloud space
+clinvar_ht_path='/seqr-reference-data/GRCh37/clinvar/clinvar.GRCh37.ht'
```
bw2: Hard-coding the paths is pretty convenient but non-standard. 

@knguyen142 Should we move all seqr-reference-data paths to the luigi config?

https://github.com/macarthur-lab/hail-elasticsearch-pipelines/blob/master/luigi_pipeline/lib/global_config.py

knguyen142: Those can be overridden https://github.com/macarthur-lab/hail-elasticsearch-pipelines/blob/master/luigi_pipeline/configs/seqr-loading-local-GRCh37.cfg#L2-L3

hail_scripts/v02/update_reference_data_in_existing_index.py
```diff
@@ -0,0 +1,142 @@
...
+
+
+args = p.parse_args()
+
+#if args.download_latest_clinvar_vcf:
+#    for genome_version in ["37", "38"]:
+#        vds = download_and_import_latest_clinvar_vcf(hc, genome_version)
+#        write_vds(vds, CLINVAR_VDS_PATH.format(genome_version=genome_version))
+
+if args.index_name and not args.all:
```
knguyen142: If I'm reading this correct, the order of precedence is all and then the individual index name. If so, this is equivalent to

```
if args.all:
  ...
elif args.index_name:
  ...
else:
  ...
```

hail_scripts/v02/update_reference_data_in_existing_index.py
knguyen142: Ahh maybe I misunderstood your suggestion, what are you suggesting?

hail_scripts/v02/update_models/update_mt_schema.py
```diff
@@ -0,0 +1,53 @@
+import hail as hl
+
+from lib.model.base_mt_schema import BaseMTSchema, row_annotation, RowAnnotationOmit
+from hail_scripts.v02.utils.computed_fields import variant_id
+
+class UpdateSchema(BaseMTSchema):
```
knguyen142: If the annotations are as simple as this, I actually don't think the schema is needed. The schema models were made to model complex relationships between the data, but if most of this is a few annotations that are used only in the script, we can just add it straight in the script.

NLSVTN: Yeah, just code reuse seems like a good idea for me. When you see how the whole index load happens and understand it, update index code would look obvious right away.

hail_scripts/v02/update_reference_data_in_existing_index.py
bw2: Never mind, I just noticed [reference_ht_path](https://github.com/macarthur-lab/hail-elasticsearch-pipelines/blob/master/luigi_pipeline/configs/seqr-loading-local-GRCh37.cfg#L11) is also in there and contains all the other ref datasets. Then this is the only script that needs to be updated.

bw2:
If no one has other comments, I'll merge this as is.

PR #191, opened 2020-01-28
Fixed POPMAX_OR_GLOBAL fallback value

mike-w-wilson:
This adds in a fallback to gnomAD global AF for a variant when POPMAX AF is not available. This was tested on the variant chr4-73258424-G-T in gnomAD v3 (4-74124141-G-T in gnomAD v2) where the missing value was discovered.


nawatts:
FYI, you could write this a little more tersely using [hl.or_else](https://hail.is/docs/0.2/functions/core.html#hail.expr.functions.or_else).

knguyen142:
LGTM but I don't really know the gnomad/hail stuff.

PR #200, opened 2020-03-19
Make it easier for local installs to use the hail v02 pipeline

bw2:
gs://hail-common files have been moved to other buckets - which broke the v01 pipeline and provided extra motivation to get v02 working for non-Broad (eg. outside AirFlow).


bw2:
Once this is merged we should be able to delete all the v01 code.

PR #204, opened 2020-06-23
add 37 coordinates

mike-w-wilson:
This adds a `rg37_locus` to 38 callsets. Anything that does not lift is currently missing as you can't assign anything but a locus at this annotation. I could add in a `lifted_under` annotation or something if seqr needs a separate field indicating success.

### DIFF
--- a/luigi_pipeline/lib/hail_tasks.py
+++ b/luigi_pipeline/lib/hail_tasks.py
@@ -170,6 +170,18 @@ class HailMatrixTableTask(luigi.Task):
         logger.info(f'Remapped {remap_count} sample ids...')
         return mt
 
+    @staticmethod
+    def add_37_coordinates(mt):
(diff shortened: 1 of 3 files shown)


hanars:
what format is the locus in in the final output? I'm not super familiar with hail data types. I don't think we need a special `lift_under` annotation - if we have the lifted locus where it succeeds and a null/ falsety value thats enough for seqr

mike-w-wilson:
Hail has a locus data type so  rg37_locus is a locus in the variant matrix table. It looks like the ES export interprets it as 
```
        "rg37_locus": {
          "properties": {
            "contig": {
              "type": "keyword"
            },
            "position": {
              "type": "integer"
            }
          }
        },
```
Would it make sense to cast it to a str?

hanars:
I think thats fine for the export, no need for a string

PR #205, opened 2020-06-30
make it easier to load VCFs with non-standard chromosome names

bw2:
motivated by https://github.com/macarthur-lab/seqr/issues/1343

### DIFF
--- a/luigi_pipeline/lib/hail_tasks.py
+++ b/luigi_pipeline/lib/hail_tasks.py
@@ -75,8 +75,16 @@ class HailMatrixTableTask(luigi.Task):
 
     def import_vcf(self):
         # Import the VCFs from inputs. Set min partitions so that local pipeline execution takes advantage of all CPUs.
+        recode = {}
+        if self.genome_version == "38":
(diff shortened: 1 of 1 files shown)


luigi_pipeline/lib/hail_tasks.py
```diff
@@ -75,8 +75,16 @@ def run(self):
 
     def import_vcf(self):
         # Import the VCFs from inputs. Set min partitions so that local pipeline execution takes advantage of all CPUs.
+        recode = {}
+        if self.genome_version == "38":
+            recode = {f"{i}": f"chr{i}" for i in (list(range(1, 23)) + ['X', 'Y'])}
```
hanars, line 80: Wouldn't we want both build versions to map to the same set of chromosome names?

bw2: That's what you'd think, but the official names for GRCh37 contigs are 1,2,3,4 and GRCh38 is chr1, chr2, chr..

hanars: 🤦

PR #206, opened 2020-07-10
Seqr sv pipeline

hanars:
Here's the code for the seqr SV loading pipeline. I'm not entirely convinced this is the best place for it to live given that it doesn't use hail in any way, but it utilizes all the helper functions for exporting to elasticsearch so it's convenient for it to be here


sv_pipeline/load_data.py
```diff
@@ -0,0 +1,392 @@
...
+    load_file(file_path, _parse_row, columns=IN_SILICO_COLS)
+
+
+def format_sv(sv):
+    sv[TRANSCRIPTS_FIELD] = [{'gene_id': gene} for gene in sv[GENES_FIELD]]
+    sv['transcriptConsequenceTerms'] = [sv[CALL_FIELD]]
+    if sv[SF_FIELD]:
+        sv['sn'] = int(sv[SC_FIELD] / sv[SF_FIELD])
+    sv['pos'] = sv[START_COL]
+    sv['xpos'] = CHROM_TO_XPOS_OFFSET[sv[CHROM_FIELD]] + sv[START_COL]
```
ch-kr, line 252: what is the purpose of this `CHROM_TO_XPOS_OFFSET`? is this a formatting change for export to elasticsearch?

sv_pipeline/load_data.py
```diff
@@ -0,0 +1,392 @@
...
+    sv['xstart'] = sv['xpos']
```
ch-kr, line 253: why does this value need to be stored twice? also, why do you store the start column as `pos` in addition to `xpos` and `xstart`?

sv_pipeline/load_data.py
```diff
@@ -0,0 +1,392 @@
...
+
+def subset_and_group_svs(input_dataset, sample_subset, sample_remap, ignore_missing_samples, write_subsetted_bed=False):
+    parsed_svs_by_name = {}
+    found_samples = set()
+    skipped_samples = set()
+    out_file_path = None
+    if write_subsetted_bed:
+        file_path = input_dataset.split('/')
+        file_path[-1] = 'subset_{}'.format(file_path[-1])
+        out_file_path ='/'.join(file_path)
```
ch-kr, line 205: totally a nitpick -- maybe add a space after `out_file_path`

sv_pipeline/load_data.py
```diff
@@ -0,0 +1,392 @@
...
+        parsed_svs_by_id[variant_id] = get_parsed_column_values(row, header_indices, CORE_COLUMNS)
+        parsed_svs_by_id[variant_id][COL_CONFIGS[VAR_NAME_COL]['field_name']] = variant_id
+        parsed_svs_by_id[variant_id][GENOTYPES_FIELD] = []
+
+    sample_info = get_parsed_column_values(row, header_indices, SAMPLE_COLUMNS)
+    sample_info[SAMPLE_ID_FIELD] = sample_id
+
+    sv = parsed_svs_by_id[variant_id]
+    sv[GENOTYPES_FIELD].append(sample_info)
+    # Use the largest coordinates for the merged SV
```
ch-kr, line 167: maybe I'm misreading this, but wouldn't you want to show the actual coordinates of the sample (not the largest coordinates for overlapping SVs)?

sv_pipeline/load_data.py
```diff
@@ -0,0 +1,392 @@
...
+    for genotype in sv[GENOTYPES_FIELD]:
+        sample_id = genotype['sample_id']
+        sv['samples'].append(sample_id)
+
+        cn_key = 'samples_cn_{}'.format(genotype['cn']) if genotype['cn'] < 4 else 'samples_cn_gte_4'
+        if cn_key not in sv:
+            sv[cn_key] = []
+        sv[cn_key].append(sample_id)
+
+        if genotype[QS_FIELD] > 1000:
```
ch-kr, line 265: I can't remember, is 1000 the cutoff for QS from the Talkowski group where the score is no longer informative?

sv_pipeline/load_data.py
```diff
@@ -0,0 +1,392 @@
...
+    sv['pos'] = sv[START_COL]
+    sv['xpos'] = CHROM_TO_XPOS_OFFSET[sv[CHROM_FIELD]] + sv[START_COL]
+    sv['xstart'] = sv['xpos']
+    sv['xstop'] = CHROM_TO_XPOS_OFFSET[sv[CHROM_FIELD]] + sv[END_COL]
+    sv['samples'] = []
+    for genotype in sv[GENOTYPES_FIELD]:
+        sample_id = genotype['sample_id']
+        sv['samples'].append(sample_id)
+
+        cn_key = 'samples_cn_{}'.format(genotype['cn']) if genotype['cn'] < 4 else 'samples_cn_gte_4'
```
ch-kr, line 260: do we not display copy numbers greater than 4? I wouldn't know what to do with them, just curious

sv_pipeline/load_data.py
```diff
@@ -0,0 +1,392 @@
...
+
+CORE_COLUMNS = [CHR_COL, SC_COL, SF_COL, CALL_COL, GENES_COL]
+SAMPLE_COLUMNS = [START_COL, END_COL, QS_COL, CN_COL,  NUM_EXON_COL, DEFRAGGED_COL, SAMPLE_COL]
+COLUMNS = CORE_COLUMNS + SAMPLE_COLUMNS + [VAR_NAME_COL]
+
+IN_SILICO_COLS = [VAR_NAME_COL, CALL_COL, IN_SILICO_COL]
+
+QS_BIN_SIZE = 10
+
+CHROMOSOMES = [
```
ch-kr, line 84: this is a basic question, do we get chrM SV calls? (it looks like there aren't any in the latest callset)

sv_pipeline/load_data.py
```diff
@@ -0,0 +1,392 @@
...
+    str: 'keyword',
+    bool: 'boolean',
+}
+ES_FIELD_TYPLE_MAP = {
+    'xpos': 'long',
+    'xstart': 'long',
+    'xstop': 'long',
+}
+ES_INDEX_TYPE = 'structural_variant'
+NUM_SHARDS = 6
```
ch-kr, line 102: it might be nice to be able to adjust this at runtime, unless you don't expect needing a different shard number?

sv_pipeline/load_data.py
```diff
@@ -0,0 +1,392 @@
...
+
+    args = p.parse_args()
+
+    sample_subset = None
+    sample_remap = None
+    if not args.skip_sample_subset:
+        sample_subset = get_sample_subset(args.project_guid)
+        sample_remap = get_sample_remap(args.project_guid)
+        message = 'Subsetting to {} samples'.format(len(sample_subset))
+        if sample_remap:
```
ch-kr, line 359: this assumes the remap file will always exist for projects that need remapping, right? though I'm not sure how we would provide a list of projects that need remapping. I guess that would need to be a manual check

sv_pipeline/load_data.py
```diff
@@ -0,0 +1,392 @@
...
+            skipped_samples.add(sample_id)
+            return False
+
+    load_file(input_dataset, _parse_row, out_file_path=out_file_path)
+
+    print('Found {} sample ids'.format(len(found_samples)))
+    if sample_subset:
+        if len(found_samples) != len(sample_subset):
+            missed_samples = sample_subset - found_samples
+            missing_sample_error = 'Missing the following {} samples:\n{}'.format(
```
ch-kr, line 225: what happens if you find more samples than are in the sample subset?

hanars:
> Also, I don't think I'm qualified to review the export to elasticsearch code; not sure who can help out with that bit.

@nawatts ?

sv_pipeline/load_data.py
```diff
@@ -0,0 +1,392 @@
...
+            parse_sv_row(row, parsed_svs_by_name, header_indices, sample_id)
+            found_samples.add(sample_id)
+            return True
+        else:
+            skipped_samples.add(sample_id)
+            return False
+
+    load_file(input_dataset, _parse_row, out_file_path=out_file_path)
+
+    print('Found {} sample ids'.format(len(found_samples)))
```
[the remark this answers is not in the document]
hanars: I'm still running this in python 2 though

sv_pipeline/load_data.py
```diff
@@ -0,0 +1,392 @@
...
+from datetime import datetime
+from elasticsearch import helpers as es_helpers
+import re
+import subprocess
+from tqdm import tqdm
+
+from hail_scripts.shared.elasticsearch_client import ElasticsearchClient
+from hail_scripts.shared.elasticsearch_utils import ELASTICSEARCH_INDEX
+
+GS_SAMPLE_PATH = 'gs://seqr-datasets/v02/GRCh38/RDG_WES_Broad_Internal/base/projects/{project_guid}/{project_guid}_{file_ext}'
```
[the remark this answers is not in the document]
hanars: I'm not entirely sure on what this doe does or why its useful here. Can you explain a bit more?

sv_pipeline/load_data.py
```diff
@@ -0,0 +1,392 @@
...
+
+GENCODE_CHR_COL_IDX = 0
```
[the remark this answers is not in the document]
hanars: nope they aren't used anymore, I accidentally left them in from back when I thought the file format was going to be much more stable :)

sv_pipeline/load_data.py
```diff
@@ -0,0 +1,392 @@
...
+GENCODE_TYPE_COL_IDX = 2
+GENCODE_START_COL_IDX = 3
+GENCODE_END_COL_IDX = 4
+GENCODE_INFO_COL_IDX = 8
+
+CHR_COL = 'chr'
```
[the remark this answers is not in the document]
hanars: I fundamentally disagree that this pipeline needs to be so resilient to handle constantly changing file formats. This is a production grade loading pipeline designed to load a standard output format, and I developed it under the belief that we had a standard format already agreed on (which is what I was told). I will make this change to accept the reality that I can't seem to rely on the input to have any stability, but I am honestly still upset that I continue to have to spend even more time on this code to handle what I don't believe to be an acceptable level of instability in the input

sv_pipeline/load_data.py
```diff
@@ -0,0 +1,392 @@
...
+    SF_COL: {'field_name': SF_FIELD, 'format': float},
+    VAR_NAME_COL: {'field_name': VARIANT_ID_FIELD, 'format': lambda val, call='any': '{}_{}'.format(val, call)},
+    CALL_COL: {'field_name': CALL_FIELD},
+    START_COL: {'format': int},
+    END_COL: {'format': int},
+    QS_COL: {'field_name': QS_FIELD, 'format': int},
+    CN_COL: {'field_name': CN_FIELD, 'format': int},
+    NUM_EXON_COL: {'field_name': NUM_EXON_FIELD, 'format': lambda val: 0 if val == 'NA' else int(val)},
+    DEFRAGGED_COL: {'field_name': DEFRAGGED_FIELD, 'format': lambda val: BOOL_MAP[val]},
+    IN_SILICO_COL: {
```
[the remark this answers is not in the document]
hanars: yes, I confirmed with Anne and Julia

hail_scripts/shared/elasticsearch_client.py
```diff
@@ -339,3 +339,31 @@ def set_index_meta(self, index_name, index_type_name, _meta):
         }
 
         self.es.indices.put_mapping(index=index_name, doc_type=index_type_name, body=index_mapping)
+
+    def route_index_to_temp_es_cluster(self, index_name, to_temp_loading):
```
nawatts, line 343: This function seems like it could be confusing since passing `to_temp_loading=False` makes it do the exact opposite of what its name suggests that it does. Maybe it could use a different name or be split into two functions?

sv_pipeline/load_data.py
```diff
@@ -0,0 +1,392 @@
...
+    },
+    SAMPLE_COL: {
+        'field_name': SAMPLE_ID_FIELD,
+        'format': lambda val: re.search('(\d+)_(?P<sample_id>.+)_v\d_Exome_GCP', val).group('sample_id'),
+    },
+    GENES_COL: {'field_name': GENES_FIELD, 'format': lambda genes: [gene.split('.')[0] for gene in genes.split(',')]},
+}
+
+CORE_COLUMNS = [CHR_COL, SC_COL, SF_COL, CALL_COL, GENES_COL]
+SAMPLE_COLUMNS = [START_COL, END_COL, QS_COL, CN_COL,  NUM_EXON_COL, DEFRAGGED_COL, SAMPLE_COL]
```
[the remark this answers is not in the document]
hanars: So if we have an SV called in multiple samples, we always display it in seqr as its largest possible size with all possible genes. However, since there is some variance in size for individual calls even if they are considered the same call, when you hover over the individual to see the details for that ca we will show if the number of exons for that sample is different than the top level number

sv_pipeline/load_data.py
```diff
@@ -0,0 +1,392 @@
...
+]
+CHROM_TO_XPOS_OFFSET = {chrom: (1 + i)*int(1e9) for i, chrom in enumerate(CHROMOSOMES)}
+
+ES_TYPE_MAP = {
+    int: 'integer',
+    float: 'double',
+    str: 'keyword',
+    bool: 'boolean',
+}
+ES_FIELD_TYPLE_MAP = {
```
nawatts, line 96: Was this intended to be "TYPE_MAP"?

sv_pipeline/load_data.py
hanars: I copied this utility over from SNV utilities, so quite possibly no. Do you think there is harm in leaving this in?

sv_pipeline/load_data.py
hanars: Agreed, especially since the SNV pipeline makes this configurable. For what its worth, I think 6 may even be too large for many of these indices, they are all tiny and could probably even be 3

sv_pipeline/load_data.py
hanars: we usually do! We are loading from a file with all the CMG samples and we are only looking for one project's worth. We just don't load those

sv_pipeline/load_data.py
hanars: `xpos` is a representation of position by a single number instead of as a chrom-pos pair and seqr heavily relies on it but its doesn't exist in this raw data so we have to compute it. So for example `chr1:248367227` has an `xpos` of 1248367227. `CHROM_TO_XPOS_OFFSET` is how we figure out how to represent the chromosome - for the numbered chromosomes its just `chr * 1,000,000,000` (so 1000000000, 2000000000, 3000000000, etc) but we can't use that formula here because X and Y are not actually numbers but are real chromosomes, so we precompute the offsets by chromosome in a way that more easily handles those edge cases

sv_pipeline/load_data.py
hanars: idk, `xstart` is just because the SNV ES records have both so I figured we should be consistent. And seqr uses both `pos` and `xpos` for some reason so I wanted to keep both. Like objectively it might have been better to make seqr stick to one, but I was trying to add SVs without making any real changes to how SNV search works

sv_pipeline/load_data.py
hanars: We do! So we store the actual copy number on the `genotype` level object for the sample (which includes the CN and all the quality details) , and then we also store a list of sample ids where the copy number is greater than for. Thats because elasticsearch is WAY faster at answering the query "records where `cn_gte_4` contains "X" and "Y"" than "records where there is at least one nested `genotype` object with `sample_id` of either "X" or "Y"  and `CN` greater than 4". We do this for SNVs too with lists like `samples_num_alt_1`. FOR SVs we have `samples_cn_0`, `samples_cn_1`, ..., `samples_cn_3` and then everything more than that we group because for inheritance based searching (which is how you query based on CN) we consider all CNs greater than 4 to have the same inheritance pattern

sv_pipeline/load_data.py
hanars: This is actually based on a slack conversation I had with you while I was writing the first pass at this pipeline :) : https://the-tgg.slack.com/archives/GUVVDL6MQ/p1583964252004600

sv_pipeline/load_data.py
hanars: nope! `get_sample_remap` just returns `None` if the file doesn't exist. A good argument for why docstrings are helpful

sv_pipeline/load_data.py
```diff
@@ -0,0 +1,392 @@
...
+    if not args.skip_sample_subset:
+        sample_subset = get_sample_subset(args.project_guid)
+        sample_remap = get_sample_remap(args.project_guid)
+        message = 'Subsetting to {} samples'.format(len(sample_subset))
+        if sample_remap:
+            message += ' (remapping {} samples)'.format(len(sample_remap))
+        print(message)
+
+    print('Parsing BED file')
+    parsed_svs_by_name = subset_and_group_svs(
```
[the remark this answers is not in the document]
hanars: No, it iterates through the whole file, does a minimal amount of parsing on each row to determine if the sample id is one we want, and then if it is a sample we want to include parses that row, and returns the subset of parsed SVs for the sample subset. We don't compute frequencies here, they are in the file we are given.

hail_scripts/shared/elasticsearch_client.py
hanars: Agreed in theory, in practice this is used all over the place and I copy-pasted it exactly to make it part of a base class instead of a subclass so I'm really hesitant to actually make any changes to it

sv_pipeline/load_data.py
hanars: I'm not ure I understand

nawatts: I thought maybe it was intended to be TYPE instead of TYP**L**E.

hanars: 😳

sv_pipeline/load_data.py
```diff
@@ -0,0 +1,392 @@
+#!/usr/bin/env python
+
+import argparse
```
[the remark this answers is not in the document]
hanars: I think I fixed this, let me know if I made any mistakes

sv_pipeline/load_data.py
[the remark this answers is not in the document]
hanars: everything is lowercased now

sv_pipeline/load_data.py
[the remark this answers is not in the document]
ch-kr: ooh, thanks! I ran pylint to check, and it seems to think `from tqdm import tqdm` should be before `from elasticsearch import helpers as es_helpers`. this looks good to me otherwise!

sv_pipeline/load_data.py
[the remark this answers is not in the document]
ch-kr: I like using a resources file since it makes the code a little easier to read, and I like that I only have to update a path once if it's a resource that I use often/in multiple scripts

sv_pipeline/load_data.py
ch-kr: ooh I see, thanks for explaining that!

sv_pipeline/load_data.py
ch-kr: LOL sorry memory of a goldfish lately...

sv_pipeline/load_data.py
ch-kr: probably not, was just curious

hanars:
@ch-kr @ericminikel I think I've responded to all the feedback - is there anything blocking getting this merged?

PR #208, opened 2020-08-10
Updating spliceAI to have consequence term

mike-w-wilson:
This updates spliceAI to the latest data release and adds in the splice consequence term. When there is no max delta score, "No consequence" is the annotated term. This PR also includes an update to including the "was_split" annotation on ES export. I apologize for the format changes, I ran black.

### DIFF
--- a/luigi_pipeline/lib/model/seqr_mt_schema.py
+++ b/luigi_pipeline/lib/model/seqr_mt_schema.py
@@ -1,4 +1,3 @@
-
 import hail as hl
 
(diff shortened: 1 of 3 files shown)


luigi_pipeline/lib/model/seqr_mt_schema.py
```diff
@@ -29,6 +28,10 @@ def filters(self):
     @row_annotation()
     def aIndex(self):
         return self.mt.a_index
+    
+    @row_annotation()
+    def wasSplit(self):
+        return self.mt.was_split
```
hanars, line 34: 💃

PR #211, opened 2020-09-08
copy downloaded clinvar file to hdfs. Otherwise, hail can't read it when running on dataproc

bw2:


mike-w-wilson (approved):
Sorry this took so long @bw2 ! I didnt see it until I checked on the review-requested link Grace sent!

PR #215, opened 2020-11-09
Re-added code block importing elasticsearch

ch-kr:


hail_scripts/shared/elasticsearch_client_v7.py
```diff
@@ -6,6 +6,14 @@
 from pprint import pformat
 
 
+try:
+    import elasticsearch
```
hanars, line 10: you should remove the direct import on line 2 above

ch-kr: oops, good call

hanars (approved):

PR #217, opened 2020-11-24
smaller block size

hanars:


luigi_pipeline/lib/hail_tasks.py
```diff
@@ -233,6 +233,7 @@ def export_table_to_elasticsearch(self, table, num_shards):
                                                func_to_run_after_index_exists=func_to_run_after_index_exists,
                                                elasticsearch_mapping_id="docId",
                                                num_shards=num_shards,
+                                               block_size=500,
```
ch-kr, line 236: it looks like another script has the default block size set to 200 (https://github.com/broadinstitute/hail-elasticsearch-pipelines/blob/master/hail_scripts/v02/export_table_to_es.py#L32). do you think that a block size of 200 would have a higher chance of exporting successfully, or would this make the exports too slow?

hanars: sure, lets try 200

ch-kr (approved):

mike-w-wilson:
Closing this PR since temp loading nodes are back up but made an issue for adding blcoksize as a parameter in the luigi pipeline https://github.com/broadinstitute/hail-elasticsearch-pipelines/issues/220

PR #219, opened 2020-12-04
longer forcemerge timeout

hanars:


ch-kr (approved):
looks good to me. basic question, would it be helpful to also increase the general default timeout?

hail_scripts/v02/utils/elasticsearch_client.py
```diff
@@ -204,4 +204,4 @@ def export_table_to_elasticsearch(
         es.batch.write.refresh // default true  (Whether to invoke an index refresh or not after a bulk update has been completed)
         """
 
-        self.es.indices.forcemerge(index=index_name)
+        self.es.indices.forcemerge(index=index_name, request_timeout=60)
```
mike-w-wilson, line 207: I think the value is in seconds so this won't change the current 1m config. https://elasticsearch-py.readthedocs.io/en/7.10.0/api.html#timeout

mike-w-wilson (changes requested):
I may have looked at the wrong docs but I think we need this to be 3600 to be an hour timeout?

mike-w-wilson (approved):

mike-w-wilson:
Hana explained the default is currently 10s

PR #228, opened 2021-05-26
set lower minimum num shards

hanars:
small shards create a huge memory overhead and minor search performance decrease, so if data fits in one shard we should do that

### DIFF
--- a/luigi_pipeline/lib/hail_tasks.py
+++ b/luigi_pipeline/lib/hail_tasks.py
@@ -203,7 +203,7 @@ class HailElasticSearchTask(luigi.Task):
     es_index = luigi.Parameter(description='ElasticSearch index.', default='data')
     es_username = luigi.Parameter(description='ElasticSearch username.', default='pipeline')
     es_password = luigi.Parameter(description='ElasticSearch password.', visibility=ParameterVisibility.PRIVATE, default=None)
-    es_index_min_num_shards = luigi.IntParameter(default=6,
+    es_index_min_num_shards = luigi.IntParameter(default=1,
                                                  description='Number of shards for the index will be the greater of '
                                                              'this value and a calculated value based on the matrix.')
 
--- a/sv_pipeline/exome/load_data.py
+++ b/sv_pipeline/exome/load_data.py
@@ -405,7 +405,7 @@ def main():
     p.add_argument('--sample-type', default='WES')
     p.add_argument('--es-host', default='localhost')
     p.add_argument('--es-port', default='9200')
-    p.add_argument('--num-shards', default=6)
+    p.add_argument('--num-shards', default=1)
 
     args = p.parse_args()
 
--- a/sv_pipeline/exome/load_data.py
+++ b/sv_pipeline/exome/load_data.py
@@ -346,7 +346,7 @@ def get_es_schema(all_fields, nested_fields):
     return schema
 
 
-def export_to_elasticsearch(es_host, es_port, rows, index_name, meta, es_password, num_shards=6):
+def export_to_elasticsearch(es_host, es_port, rows, index_name, meta, es_password, num_shards=1):
     """
     Export SV data to elasticsearch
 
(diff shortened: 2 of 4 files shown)


sv_pipeline/exome/load_data.py
```diff
@@ -405,7 +405,7 @@ def main():
     p.add_argument('--sample-type', default='WES')
     p.add_argument('--es-host', default='localhost')
     p.add_argument('--es-port', default='9200')
-    p.add_argument('--num-shards', default=6)
+    p.add_argument('--num-shards', default=1)
```
williamphu, line 408: L349 has `num_shards=6`; I think since there's a default for the argument, that the function should not default to 6, but should we change that as well?

williamphu (commented):
One small question.

luigi_pipeline/lib/hail_tasks.py
```diff
@@ -203,7 +203,7 @@ class HailElasticSearchTask(luigi.Task):
     es_index = luigi.Parameter(description='ElasticSearch index.', default='data')
     es_username = luigi.Parameter(description='ElasticSearch username.', default='pipeline')
     es_password = luigi.Parameter(description='ElasticSearch password.', visibility=ParameterVisibility.PRIVATE, default=None)
-    es_index_min_num_shards = luigi.IntParameter(default=6,
+    es_index_min_num_shards = luigi.IntParameter(default=1,
```
ch-kr, line 206: this parameter looks like it's used on line 246 (https://github.com/broadinstitute/hail-elasticsearch-pipelines/blob/master/luigi_pipeline/lib/hail_tasks.py#L246) to determine whether this int or the calculated int is larger. would it make sense to make the `max` on line 246 a `min`? or should we just always set the number of shards equal to the calculated number. I can't imagine this number will ever be larger than the calculated number now. for example, I checked this project MT: `gs://seqr-datasets/v02/GRCh38/RDG_WGS_Broad_Internal/v18/projects/R0312_cmg_bonnemann_genomes.mt`, and this MT would have 3 shards based on `_mt_num_shards`

sv_pipeline/exome/load_data.py
```diff
@@ -346,7 +346,7 @@ def get_es_schema(all_fields, nested_fields):
     return schema
 
 
-def export_to_elasticsearch(es_host, es_port, rows, index_name, meta, es_password, num_shards=6):
+def export_to_elasticsearch(es_host, es_port, rows, index_name, meta, es_password, num_shards=1):
```
ch-kr, line 349: completely unrelated to the num_shards argument here, I just noticed this is `export_to_elasticsearch` in the exomes SV but `export_to_es` in the genomes (might be nice to rename one for consistency)

ch-kr (commented):
small question about `hail_tasks` code

luigi_pipeline/lib/hail_tasks.py
mike-w-wilson: I'm not sure I understand how min would work. The max is there because we hit limits with the number of docs within an index so we should always take the max number.

ch-kr: ah gotcha. the max isn't required anymore then is it? would any of our project MTs calculate to have a fewer than 1 shard? line 246 will probably always return `calculated_num_shards` now, right?

mike-w-wilson: Yeah,  the calculation is always rounded to the next integer so 1 is the smallest -- max is not needed unless for some reason we would want to spread the information over more shards than the calculation? I can't imagine a situation when we'd want that considering this PR

sv_pipeline/exome/load_data.py
hanars: given that this is an outage-related urgent fix I'm going to consider style fixes to be out of the scope of this PR

luigi_pipeline/lib/hail_tasks.py
hanars: so to clarify, changing this default and leaving the rest of the code as-is has the desired effect, correct?

mike-w-wilson: Yes it does

mike-w-wilson (approved):
LGTM. I think we can revisit removing max considering the urgency but we should probably think through that more.

williamphu (approved):
LGTM as well.

PR #247, opened 2021-06-07
Batch apply row annotations

### LINKED ISSUE #242: Seqr load: Row annotations are applied separately
Hi, I've been testing the hail-elastic search pipeline, and noticed that each annotation is applied separately by the `row_annotation` decorator. 

The `_ref_data` annotations (eg: [`cadd`, `dbnsfp`, `geno2mp`, etc](https://github.com/broadinstitute/hail-elasticsearch-pipelines/blob/a12cf68c733c1937cd8334a9cba86fec8d34a95b/luigi_pipeline/lib/model/seqr_mt_schema.py#L122-L176)) use an externally shared reference table, which as I understand because the matrix table (and hence `mt.row_key`) changes, hail has to generate a new join for each annotation which is pretty resource intensive.

I believe it's possible to batch changes non-dependent changes into a single `annotate_rows`. I've put together a [PR on our populationgenomics fork](https://github.com/populationgenomics/hail-elasticsearch-pipelines/pull/3) to test this implementation, which returns consistent results.

Brief summary of changes:

- The `row_annotation` returns a data class that stores properties
- Implemented a list of requirements, rather than just the single
- Apply changes in `annotate_all`, rather than in the `row_annotation` by:
  - Applying "non-dependent" annotations in rounds until all annotations have been applied, or a circular dependency is detected.
  - Row annotations that use the [selected_ref_data](https://github.com/populationgenomics/hail-elasticsearch-pipelines/blob/fa5e46771bed673f18bacb472d677122320db2c4/luigi_pipeline/lib/model/seqr_mt_schema.py#L142-L144), pull it from a [cached attribute](https://github.com/populationgenomics/hail-elasticsearch-pipelines/blob/fa5e46771bed673f18bacb472d677122320db2c4/luigi_pipeline/lib/model/seqr_mt_schema.py#L30-L34).

I'm raising an issue first to see if there's interest in upstreaming this, and what the process should be (as it's a structural change) for testing and upstreaming this.

Thanks!

illusional:
Fixes #242.

Brief summary of changes:

- The row_annotation returns a data class that stores properties
- Allows a list of requirements, rather than just the single requirement
- Apply changes in annotate_all, rather than in the row_annotation by:
  - Applying "non-dependent" annotations in rounds until all annotations have been applied, or a circular dependency is detected.
- Row annotations that use the selected_ref_data, pull it from a cached attribute.

From some rough testing (thanks @tommyli), we saw:

- about 2-4 times speedup during the annotation;
- 70-75% reduction in network traffic (I presume when pulling reference data);
- Still about 15% cpu utilisation (over 2 cpus in the spark cluster).


luigi_pipeline/lib/model/seqr_mt_schema.py
```diff
@@ -8,11 +8,31 @@
...
-        super().__init__(*args, **kwargs)
         self._ref_data = ref_data
         self._clinvar_data = clinvar_data
         self._hgmd_data = hgmd_data
 
+        # reuse:
+        #   self._ref_data[self.mt.row_key]
+        # for all annotations. We'll use the @property _selected_ref_data
+        # to access this value, and lazily cache it (so as not to compute
+        # it if it doesn't get get accessed after an update to self.mt
```
hanars, line 19: this comment would be more helpful as a docstring for the `_selected_ref_data` property

luigi_pipeline/lib/model/base_mt_schema.py
```diff
@@ -109,30 +92,104 @@ def c_1(self):
...
 
     def annotate_all(self, overwrite=False):
         """
         Iterate over all annotation functions and call them on the instance.
         :return: instance object
         """
-        for atn_fn in self.all_annotation_fns():
-            getattr(self, atn_fn[0])(overwrite=overwrite)
+        called_annotations = set()
+        rounds: List[List[RowAnnotation]] = [self.all_annotation_fns()]
```
hanars, line 129: the name `rounds` is a bit confusing, and I think what you are achieving via a list of lists thats modified during iteration would be much cleaner with a recursive helper function. Let me know if you have any questions about that, I'm happy to help code that out for you

luigi_pipeline/lib/model/base_mt_schema.py
```diff
@@ -109,30 +92,104 @@ def c_1(self):
...
+        print(f'Will attempt to apply {len(rounds[0])} row annotations')
```
hanars, line 130: we use `loggger`, not `print` for logging messages. Anywhere that used `print` for debugging reasons can probably taken out, and anywhere that had useful information for production can get switched to `logger.info`

hanars (changes requested):
Thanks so much for submitting this! Really excited for the performance gains, I just have some style/ maintainability change requests

mike-w-wilson:
Thanks for these updates @illusional! We discussed this PR in person and actually prefer the previous commit with the exception of the print->logger and docstring updates. Apologies for the extra work caused but we really appreciate the improvements. Could you roll back to https://github.com/broadinstitute/hail-elasticsearch-pipelines/pull/247/commits/5eeb5a9841cb762bd802dfd41ec2b12f188197dc and just address the previous comments on print and the docstrings? Thanks!

illusional:
Hey @mike-w-wilson and @hanars, thanks for reviewing those additional changes. I'm curious why you didn't like the refactor? I think preparing the list of annotations to apply is cleaner, as you get essentially a set of instructions to apply which is easier to test.

I've reverted the change, and applied the review fixes.

illusional:
Hi @mike-w-wilson and @hanars, any update here?

mike-w-wilson (approved):
Sorry @illusional, this slipped through. I'm going to merge it with master now!

PR #250, opened 2021-06-29
Add 'nodes.wan.only' parameter to export_table_to_elasticsearch()

ShifaSZ:
When the ES is running in a docker container, it could incorrectly get the internal container IP for the address of the ES node rather than the one that can be accessed externally. Add 'nodes.wan.only' can solve the problem. See [the reference](https://discuss.elastic.co/t/get-org-elasticsearch-hadoop-rest-eshadoopnonodesleftexception-connection-error-check-network-and-or-proxy-settings-all-nodes-failed-tried/79509).

### DIFF
--- a/hail_scripts/v02/utils/elasticsearch_client.py
+++ b/hail_scripts/v02/utils/elasticsearch_client.py
@@ -42,6 +42,7 @@ class ElasticsearchClient(BaseElasticsearchClient):
         export_globals_to_index_meta=True,
         verbose=True,
         write_null_values=False,
+        elasticsearch_config={},
     ):
         """Create a new elasticsearch index to store the records in this table, and then export all records to it.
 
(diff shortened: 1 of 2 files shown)


hail_scripts/v02/utils/elasticsearch_client.py
```diff
@@ -42,6 +42,7 @@ def export_table_to_elasticsearch(
         export_globals_to_index_meta=True,
         verbose=True,
         write_null_values=False,
+        elasticsearch_config={},
```
hanars, line 45: its really bad practice to use list or dict literals as default kwargs- I'm surprised the linter isn't complaining. Its beeter to use None as the default and then handle it in the code (i.e `elasticsearch_config = elasticsearch_config or {}`

mike-w-wilson (approved):
Looks good to me!

PR #261, opened 2021-07-28
Add cleanup step to  wait for shard transfer

mike-w-wilson:
This adds a waiting step for the shard transfer for all projects except RGP. Not hardcoding this bit is probably ideal but I was trying to take the manual step out of the airflow config. The code was based off of v01 but I changed the logger.info to logger.warning because we recently changed Airflow's logging config to warn. I debated increasing the wait time to 10 but stuck with 5 in the end. Happy to change if there are strong opinions. I tested this with this pipeline version: gs://seqr-luigi/releases/optimized-v1.4.5.test/ and the dataproc job is here: https://console.cloud.google.com/dataproc/jobs/pyspark_export_project_R0449_pathways_mgh_20210723_01d8d1ec?region=global&authuser=0&project=seqr-project 

This also adds a changelog to the repo.

### DIFF
--- a/hail_scripts/shared/elasticsearch_client_v7.py
+++ b/hail_scripts/shared/elasticsearch_client_v7.py
@@ -131,7 +131,7 @@ class ElasticsearchClient:
         mappings = self.es.indices.get_mapping(index=index_name)
         return mappings.get(index_name, {}).get('mappings', {}).get('_meta', {})
 
-    def wait_for_loading_shards_transfer(self, index_name, num_attempts=1000):
+    def wait_for_shard_transfer(self, index_name, num_attempts=1000):
         """
         Wait for shards to move off of the loading nodes before connecting to seqr 
         """
--- a/hail_scripts/shared/elasticsearch_client_v7.py
+++ b/hail_scripts/shared/elasticsearch_client_v7.py
@@ -139,7 +139,7 @@ class ElasticsearchClient:
         if "r0384_rare_genomes_project_gen" not in index_name:
             for i in range(num_attempts):
                 shards = self.es.cat.shards(index=index_name)
-                if "es-data-loading" not in shards:
+                if LOADING_NODES_NAME not in shards:
                     logger.warning("Shards are on {}".format(shards))
                     return
                 logger.warning("Waiting for {} shards to transfer off the es-data-loading nodes: \n{}".format(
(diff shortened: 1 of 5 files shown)


hail_scripts/shared/elasticsearch_client_v7.py
```diff
@@ -130,3 +130,22 @@ def _update_settings(self, index_name, body):
     def get_index_meta(self, index_name):
         mappings = self.es.indices.get_mapping(index=index_name)
         return mappings.get(index_name, {}).get('mappings', {}).get('_meta', {})
+
+    def wait_for_shard_transfer(self, index_name, num_attempts=1000):
+        """
+        Wait for shards to move off of the loading nodes before connecting to seqr 
+        """
+        # RGP is too large and will not transfer until old index is deleted
+        if "r0384_rare_genomes_project_gen" not in index_name:
```
hanars, line 139: I understand wanting to take a manual step out of the airflow config, but I am worried that this is really fragile. what if we instead checked if it was a WGS project with > 1000 samples? That would basically only affect RGP, but would be more resilient and would probably help if we ever did get another huge WGS project

hail_scripts/shared/elasticsearch_client_v7.py
```diff
@@ -130,3 +130,22 @@ def _update_settings(self, index_name, body):
...
+            for i in range(num_attempts):
+                shards = self.es.cat.shards(index=index_name)
+                if "es-data-loading" not in shards:
```
hanars, line 142: would be better to use the `LOADING_NODES_NAME` constant here

hanars (changes requested):

hail_scripts/shared/elasticsearch_client_v7.py
mike-w-wilson: Yeah that is a good point. Do you think 1000 is the right cut off? I'm wondering if we should go off total index size instead? Though I'm not sure if that can be retrieved.

hanars: I think you are right, total index size would be better than number of samples. Maybe use num shards as a proxy for that, since we have that computed in the pipeline and its based off the number of total variants?

mike-w-wilson: Great idea!

hanars (approved):

mike-w-wilson:
I updated to the logic around the wait to use the shards. I set the threshold at 25 because thats roughly ~1TB index size, probably a bit more.

PR #262, opened 2021-07-28
Fixed the names for sample qs fields

ShifaSZ:
Changed the field name `sample_qs_{X}_to_{Y}` to `samples_qs_{X}_{Y}` (add an `s`).

### DIFF
--- a/sv_pipeline/genome/load_data.py
+++ b/sv_pipeline/genome/load_data.py
@@ -57,7 +57,6 @@ DERIVED_FIELDS = {
     'samples_no_call': lambda rows: get_sample_num_alt_x(rows, -1),
     'samples_num_alt_1': lambda rows: get_sample_num_alt_x(rows, 1),
     'samples_num_alt_2': lambda rows: get_sample_num_alt_x(rows, 2),
-    'samples': lambda rows: rows.genotypes.map(lambda x: x.sample_id),
 }
 
 SAMPLES_QS_FIELDS = {'samples_qs_{}_to_{}'.format(i, i+QS_BIN_SIZE): i for i in range(0, 1000, QS_BIN_SIZE)}
--- a/sv_pipeline/genome/load_data.py
+++ b/sv_pipeline/genome/load_data.py
@@ -57,12 +57,13 @@ DERIVED_FIELDS = {
     'samples_no_call': lambda rows: get_sample_num_alt_x(rows, -1),
     'samples_num_alt_1': lambda rows: get_sample_num_alt_x(rows, 1),
     'samples_num_alt_2': lambda rows: get_sample_num_alt_x(rows, 2),
+    'samples': lambda rows: rows.genotypes.map(lambda x: x.sample_id),
 }
 
-SAMPLE_QS_FIELDS = {'sample_qs_{}_to_{}'.format(i, i+QS_BIN_SIZE): i for i in range(0, 1000, QS_BIN_SIZE)}
+SAMPLES_QS_FIELDS = {'samples_qs_{}_to_{}'.format(i, i+QS_BIN_SIZE): i for i in range(0, 1000, QS_BIN_SIZE)}
 
 FIELDS = list(CORE_FIELDS.keys()) + list(DERIVED_FIELDS.keys()) + ['variantId', SORTED_TRANS_CONSEQ, 'genotypes'] +\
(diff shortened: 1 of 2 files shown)


sv_pipeline/genome/load_data_tests.py
```diff
@@ -109,19 +109,19 @@
...
+                             'samples_qs_990_to_1000': hl.literal(['SAMPLE-1', 'SAMPLE-2', 'SAMPLE-3', 'SAMPLE-4'])})
+SAMPLES_QS_FIELDS_DUP = SAMPLES_QS_FIELDS.copy()
+SAMPLES_QS_FIELDS_DUP.update({'samples_qs_0_to_10': ['SAMPLE-4'],
+                             'samples_qs_10_to_20': ['SAMPLE-3'],
+                             'samples_qs_30_to_40': ['SAMPLE-5'],
+                             'samples_qs_50_to_60': ['SAMPLE-2'],
+                             'samples_qs_990_to_1000': ['SAMPLE-1']})
+SAMPLES_QS_FIELDS_INS = SAMPLES_QS_FIELDS.copy()
+SAMPLES_QS_FIELDS_INS.update({'samples_qs_0_to_10': ['SAMPLE-5'],
+                             'samples_qs_990_to_1000': ['SAMPLE-1', 'SAMPLE-2', 'SAMPLE-3', 'SAMPLE-4']})
```
ShifaSZ, line 124: The only change is adding an 's'/'S' to the field/constant names.

sv_pipeline/genome/load_data.py
```diff
@@ -57,12 +57,13 @@
     'samples_no_call': lambda rows: get_sample_num_alt_x(rows, -1),
     'samples_num_alt_1': lambda rows: get_sample_num_alt_x(rows, 1),
     'samples_num_alt_2': lambda rows: get_sample_num_alt_x(rows, 2),
+    'samples': lambda rows: rows.genotypes.map(lambda x: x.sample_id),
```
ShifaSZ, line 60: Add `samples` field.

sv_pipeline/genome/load_data_tests.py
```diff
@@ -278,7 +281,7 @@ def test_export_to_es(self, mock_get_index, mock_es_client, mock_os):
         mock_os.environ.get.return_value = TEST_PASSWORD
         mock_es = mock_es_client.return_value
         rows = self.mt.rows().head(5)
-        export_to_es(rows, TEST_INPUT_DATASET, TEST_GUID, TEST_HOST, TEST_PORT, TEST_BLOCK_SIZE, TEST_NUM_SHARDS)
+        export_to_es(rows, TEST_INPUT_DATASET, TEST_GUID, TEST_HOST, TEST_PORT, TEST_BLOCK_SIZE, TEST_NUM_SHARDS, es_nodes_wan_only='false')
```
ShifaSZ, line 284: Tests are changed for the code changes in the other PR.

hanars:
`samples` field is for the exome SVs only - its a totally different way of searching because there are no unaffected samples in the exome data. We actually discussed this extensively on the intial genome SV code, and I have no idea why you felt the need to make that change, but in the future you need to talk to me about this first. If there is a problem you are trying to solve this is not the fix - schedule time with me to discuss if you need to

ShifaSZ:
> `samples` field is for the exome SVs only - its a totally different way of searching because there are no unaffected samples in the exome data. We actually discussed this extensively on the intial genome SV code, and I have no idea why you felt the need to make that change, but in the future you need to talk to me about this first. If there is a problem you are trying to solve this is not the fix - schedule time with me to discuss if you need to

I have taken out the `samples` field. There were quite a lot of variants unrelated to the samples from the family for search returned. I thought its reason might be we don't have the `samples` field in the Elasticsearch and the search conditions include `Bool[should{term, "samples"=["GLE_4111_4_5"]}]`. But later I found these are not the causes. 
Now, after I updated the `sample_qs_{X}_{Y}` names, the unrelated variants disappear. Can it be because the search with the condition `Bool[must_not{term, "none-exist-field"="something"}]` returns all the entries in the index?
Anyway, we need to change the field name otherwise it will not match the search in seqr.

hanars:
I'm happy to change the samples_qs field names, but you need to take the `samples` field out of this Pr first

hanars (approved):

PR #267, opened 2021-08-24
delete unused files

hanars:
I had a free hour to kill and decided to go through and delete all the unused code in the pipeline repo. If it makes you too nervous feel free to decline, but I tried my best to check and only remove things that were genuinely not used so it should be safe, and then it will make the repo A LOT more readable/ usable/ maintainable


hail_scripts/shared/elasticsearch_client.py
```diff
@@ -1,369 +0,0 @@
...
-    os.system("pip install elasticsearch")
-    import elasticsearch
-
-
-handlers = set(logging.root.handlers)
-logging.root.handlers = list(handlers)
-logger = logging.getLogger()
-
-
-class ElasticsearchClient:
```
mike-w-wilson, line 20: This is used in the sv pipeline and by the luigi pipeline through hail_script utils.

mike-w-wilson (changes requested):
Looks like hail-elasticsearch-pipelines/sv_pipeline/exome script imports from hail_scripts/shared/elasticsearch_client.py in L13 as does hail-elasticsearch-pipelines/hail_scripts/v02/utils/elasticsearch_client.py in L8-13. That file, hail_scripts/shared/elasticsearch_client.py, is the only one I'm hesitant to delete until we make sure it's not using the actual imports but it looks like they are. This is awesome though, thank you for going through all of this -- it makes this repo so much more approachable.

hail_scripts/shared/elasticsearch_client.py
hanars: so it actually isn't used as this only works for the old v6 elasticsearch that no longer exists and won't work with v7 elasticsearch which is what we have live. The sv_pipeline and hail_scripts/v02/utils/elasticsearch_client.py use the v7 elasticsearch client imported from `hail_scripts.shared.elasticsearch_client_v7` which I left in the repo, and that file does not inherit from this file at all. And the luigi piepline actually imports `hail_scripts.v02.utils.elasticsearch_client` which, as I mentioned, itself imports the v7 client. So its a REALLY confusing series of names and inheritance, but I'm really confident this isn't used

mike-w-wilson: ugh yeah, I was definitely getting confused by naming-- I was reading hail_scripts/shared/elasticsearch_utils.py, not client. 🤦

mike-w-wilson (approved):

PR #268, opened 2021-08-25
Es client naming/ hail_scripts subdirectory cleanup

hanars:
After deleting the old code, the distinction between the "shared" and "v02" subdirectories for hail_scripts was confusing/ meaningless. This merges them into a more logical subdirectory structure, and also cleans up the 2 elaasticsearch_utils and elastisearch_client files by merging the utils files into 1 file and renamimg one of the client files to make the usage more explicit

If the changes make you nervous feel free to reject them


mike-w-wilson (approved):
Sorry I missed this! The structure was so confusing, this is great, way better! Thank you!

PR #269, opened 2021-08-27
add contig check to resolve issue184

averywpx:


mike-w-wilson (changes requested):
Thanks for adding this in! I just have a few things that should be updated.

luigi_pipeline/seqr_loading.py
```diff
@@ -11,12 +11,37 @@
...
 
 
 def check_if_path_exists(path, label=""):
     if (path.startswith("gs://") and not hl.hadoop_exists(path)) or (not path.startswith("gs://") and not os.path.exists(path)):
         raise ValueError(f"{label} path not found: {path}")
 
+def contig_check(mt, threshold):
+    row_dict = mt.aggregate_rows(hl.agg.counter(mt.locus.contig))
+    contigs_set = set(row_dict.keys())
+    missing_contigs = STANDARD_CONTIGS - contigs_set
```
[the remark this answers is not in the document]
hanars: I think since the build is passed to the validate method its cleanest to pass the build to this function as well

luigi_pipeline/seqr_loading.py
```diff
@@ -11,12 +11,37 @@
...
+    invalid_contigs = {contig: None for contig in missing_contigs}
+    result = True
+    
+    for k,v in row_dict.items():
+        if k not in STANDARD_CONTIGS:
+            invalid_contigs[k] = ''
+        if v < threshold:
```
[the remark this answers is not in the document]
hanars: Yeah the original idea was to start with a basic check, especially since so far our issues have been all-or-nothing. I think in the future we can update this for chromosome-specific threshholds

luigi_pipeline/seqr_loading.py
```diff
@@ -109,6 +134,12 @@ def validate_mt(mt, genome_version, sample_type):
         :param sample_type: WGS or WES
         :return: True or Exception
         """
+
+        contig_check_result = contig_check(mt, 100)
```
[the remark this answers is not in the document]
hanars: I do think that this should either be a keyword argument to the `validate_mt` function, or at the very lease a top-level constant. Hard coding numbers into blocks of code is bad practice because its really hard to find and update them later

luigi_pipeline/seqr_loading.py
```diff
@@ -11,12 +11,42 @@
...
+    contigs_set = set(row_dict.keys())
+    # chromosomes that are not in the VCF
+    missing_contigs = standard_contigs - contigs_set
+    result_dict = {contig: None for contig in missing_contigs}
+    result = True
+    
+    for k,v in row_dict.items():
+        if k not in standard_contigs:
+            result_dict[k] = 'Unexpected string'
+        if v < threshold and 'Y' not in k:
```
hanars, line 34: `'Y'` shouldn't be hardcoded here now that its in `OPTIONAL_CHROMOSOMES`

luigi_pipeline/seqr_loading.py
```diff
@@ -11,12 +11,42 @@
...
+            result = False
+            if k in OPTIONAL_CHROMOSOMES:
+                result = True
+            elif not result_dict[k]: 
+                logger.warning('Chromosome %s is not in the VCF.', k)
+            elif result_dict[k] == 'Unexpected string':
+                logger.warning('Chromosome %s is unexpected.', k)
+            else:
+                v = result_dict[k]
+                logger.warning('Chromosome %s has %d rows, which is lower than threshold %d.', k, v, threshold)
```
hanars, line 48: iterating through your `result_dict` to log warnings feels pretty unnecessary/ confusing - just write the warning at the time you discover the error

luigi_pipeline/seqr_loading.py
```diff
@@ -11,12 +11,42 @@
...
 def check_if_path_exists(path, label=""):
     if (path.startswith("gs://") and not hl.hadoop_exists(path)) or (not path.startswith("gs://") and not os.path.exists(path)):
         raise ValueError(f"{label} path not found: {path}")
 
+def contig_check(mt, standard_contigs, threshold):
+    row_dict = mt.aggregate_rows(hl.agg.counter(mt.locus.contig))
+    contigs_set = set(row_dict.keys())
+    # chromosomes that are not in the VCF
+    missing_contigs = standard_contigs - contigs_set
+    result_dict = {contig: None for contig in missing_contigs}
```
hanars, line 28: I'm not really sure why you are merging all the different errors into a single result dictionary and then parsing out the values to figure out what the error was just so you can log it. It would be much clearer if you log errors as they happen and keep the errors in whatever format you plan to return them from the function, as nothing you see later on is going to change the error that already happened. For example, here you would do something like
```
if missing_contigs:
    logger.warning('Missing the following chromosome(s): {}'.format(', '.join(missing_contigs))
```

luigi_pipeline/seqr_loading.py
```diff
@@ -109,6 +139,14 @@ def validate_mt(mt, genome_version, sample_type):
         :param sample_type: WGS or WES
         :return: True or Exception
         """
+        if genome_version == '37':
+            contig_check_result = contig_check(mt, GRCh37_STANDARD_CONTIGS, VARIANT_THRESHOLD)
+        elif genome_version == '38':
+            contig_check_result = contig_check(mt, GRCh38_STANDARD_CONTIGS, VARIANT_THRESHOLD)
+
+        if not contig_check_result:
+            raise SeqrValidationError('The VCF failed contig check')
```
hanars, line 148: this is kind of an unhelpful failure message - you go to all the trouble of figuring out why its failing in detail, and then all you include in the failure is just that it failed. The user will have to go digging through the logs to figure out what happened instead of getting a clear message in the job failure. `contig_check_result` should include information about what failed - it can even be a formatted error message if you want, or a list of errors - and then the exception you raise should include that information. At a high level, this message should tell you which if any chromosomes are missing completely, which if any are are below threshhold, and which if any are extra

hanars (changes requested):

luigi_pipeline/seqr_loading.py
```diff
@@ -11,12 +11,38 @@
...
+    all_missing_contigs = standard_contigs - contigs_set
+    missing_contigs_without_optional = [contig for contig in all_missing_contigs if contig not in OPTIONAL_CHROMOSOMES]
+
+    if missing_contigs_without_optional:
+        check_result_dict['Missing contig(s)'] = missing_contigs_without_optional
+        logger.warning('Missing the following chromosomes(s):{}'.format(', '.join(missing_contigs_without_optional)))
+                       
+    for k,v in row_dict.items():
+        if k not in standard_contigs:
+            check_result_dict.setdefault('Unexpected string(s)',[]).append(k)
```
hanars, line 39: I would say "chromosome" or "contig" instead of "string"

luigi_pipeline/seqr_loading.py
```diff
@@ -11,12 +11,38 @@
...
+            logger.warning('Chromosome %s is unexpected.', k)
+        if (k not in OPTIONAL_CHROMOSOMES) and (v < threshold):
```
hanars, line 41: probably should be an elif` instead of an `if`, as we don't really care whether the non-standard chromosome that we don't want in our dataset at all hits the threshhold

luigi_pipeline/seqr_loading.py
```diff
@@ -109,6 +135,17 @@ def validate_mt(mt, genome_version, sample_type):
         :param sample_type: WGS or WES
         :return: True or Exception
         """
+        if genome_version == '37':
+            contig_check_result = contig_check(mt, GRCh37_STANDARD_CONTIGS, VARIANT_THRESHOLD)
+        elif genome_version == '38':
```
hanars, line 140: I would use constants rather than strings for the genome versions

luigi_pipeline/seqr_loading.py
```diff
@@ -109,6 +135,17 @@ def validate_mt(mt, genome_version, sample_type):
...
+            contig_check_result = contig_check(mt, GRCh38_STANDARD_CONTIGS, VARIANT_THRESHOLD)
+
+        if bool(contig_check_result):
```
hanars, line 143: its fine to leave this as is, but just so you know python will check the "truthyness" of anything you provide to an `if` clause without explicitly casting it to a boolean. So empty dicts and lists will be treated as false, while dicts and lists with anything in them will be treated as true

averywpx: Yes that make sense. If the VCF passes the contig check, `contig_check()` will return an empty dict. Therefore, it won't raise any SeqrValidationErrors.

mike-w-wilson (approved):
This looks good to me! This is really clean and  reads a lot easier now. Thanks for adding this in!

PR #272, opened 2021-09-24
Added id-file command line argument to overwrite the built-in id file path

ShifaSZ:
The file name for the sample IDs is usually internally generated. But it is more convenient if it can also be provided with the command line argument.

### DIFF
--- a/sv_pipeline/genome/load_data.py
+++ b/sv_pipeline/genome/load_data.py
@@ -102,9 +102,9 @@ def load_mt(input_dataset, matrixtable_file, overwrite_matrixtable):
     return hl.read_matrix_table(matrixtable_file)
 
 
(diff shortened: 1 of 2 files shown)


sv_pipeline/genome/load_data.py
```diff
@@ -205,6 +205,7 @@ def main():
     p.add_argument('--num-shards', type=int, default=1)
     p.add_argument('--block-size', type=int, default=2000)
     p.add_argument('--es-nodes-wan-only', action='store_true')
+    p.add_argument('--id-file', help='The full path (can start with gs://) of the id file.')
```
hanars, line 208: can you update this to say something like "should only be used for testing purposes, not intended for use in production"

hanars (changes requested):

hanars (approved):

PR #273, opened 2021-09-24
Change samples_qs_X_to_Y to samples_gq_sv_X_to_Y.

ShifaSZ:
To support separated GQ sliders for Exome and Genome SV searches, we need to rename the samples_qs_X_to_Y field.

### DIFF
--- a/sv_pipeline/genome/load_data.py
+++ b/sv_pipeline/genome/load_data.py
@@ -18,7 +18,7 @@ TRANS_CONSEQ_TERMS = 'transcriptConsequenceTerms'
 SORTED_TRANS_CONSEQ = 'sortedTranscriptConsequences'
 SV_TYPE = 'sv_type'
 MAJOR_CONSEQ = 'major_consequence'
-QS_BIN_SIZE = 10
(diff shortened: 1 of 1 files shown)


sv_pipeline/genome/load_data.py
```diff
@@ -59,7 +59,7 @@
     'samples_num_alt_2': lambda rows: get_sample_num_alt_x(rows, 2),
 }
 
-SAMPLES_QS_FIELDS = {'samples_qs_{}_to_{}'.format(i, i+QS_BIN_SIZE): i for i in range(0, 1000, QS_BIN_SIZE)}
+SAMPLES_QS_FIELDS = {'samples_gq_sv_{}_to_{}'.format(i, i+QS_BIN_SIZE): i for i in range(0, 1000, QS_BIN_SIZE)}
```
hanars, line 62: you need to rename all the constants that have `QS` in them, otherwise its really confusing since qs is in neither the input nor the output

hanars (changes requested):

hanars (approved):

PR #286, opened 2021-11-17
Wphu/permit shuffle on split multi

williamphu:


luigi_pipeline/seqr_loading.py
```diff
@@ -96,7 +96,7 @@ def annotate_old_and_split_multi_hts(self, mt):
         :return: mt that has pre-annotations
         """
         # Named `locus_old` instead of `old_locus` because split_multi_hts drops `old_locus`.
-        return hl.split_multi_hts(mt.annotate_rows(locus_old=mt.locus, alleles_old=mt.alleles))
+        return hl.split_multi_hts(mt.annotate_rows(locus_old=mt.locus, alleles_old=mt.alleles), permit_shuffle=True)
```
mike-w-wilson, line 99: Could you checck in with the hail team through zulip to confirm this will only try to shuffle when we have the keys out of order with duplicate loci? I know we discussed it and I read the docs as well but I'd like to confirm with them before we merge.

williamphu:
Okay @mike-w-wilson just checked with the hail team - https://hail.zulipchat.com/#narrow/stream/123010-Hail-Query.200.2E2.20support/topic/permit_shuffle.20on.20hl.2Esplit_multi_hts - the code doesn't know beforehand if the VCF needs to be shuffled, so there would be slowdowns across the board. Dan's solution was to implement a check for duplicate loci: https://discuss.hail.is/t/find-duplicates-in-matrixtable/1957. Let me know if this is something we should work on this quarter, in terms of dev-time and run-time (we'd have to check for duplicates, which uses an aggregate.

PR #280, opened 2021-10-27
disabled field indices that are not necessary

averywpx:
The following fields are disabled: 
`'aIndex', 'alt', 'AN', 'codingGeneIds', 'contig', 'docId', 'domains', 'end', 'geno2mp', 'genotypes', 'gnomad_exome_coverage', 'gnomad_genome_coverage', 'mainTranscript', 'originalAltAlleles', 'pos', 'ref', 'rg37_locus', 'sortedTranscriptConsequences', 'start', 'transcriptIds', 'wasSplit', 'xstart', 'xstop'`

PR #290, opened 2021-11-23
Aw/disable flattened indices

averywpx:
Resolved the issue occurred in #280 by disabling the flattened names in `properties` as well.

### DIFF
--- a/hail_scripts/elasticsearch/elasticsearch_utils.py
+++ b/hail_scripts/elasticsearch/elasticsearch_utils.py
@@ -100,11 +100,17 @@ def elasticsearch_schema_for_table(table, disable_doc_values_for_fields=(), disa
         logger.info("==> will disable index fields for %s", ", ".join(disable_index_for_fields))
         for es_field_name in disable_index_for_fields:
             if es_field_name not in properties:
-                raise ValueError(
-                    "'%s' in disable_index_for_fields arg is not in the elasticsearch schema: %s"
-                    % (es_field_name, properties)
-                )
-            properties[es_field_name]["index"] = False
+                flattened_fields = [key for key in properties if f"{es_field_name}_" in key]
+                if flattened_fields:
+                    for flattened_es_field_name in flattened_fields:
+                        properties[flattened_es_field_name]["index"] = False
+                else:
+                    raise ValueError(
+                        "'%s' in disable_index_for_fields arg is not in the elasticsearch schema: %s"
+                        % (es_field_name, properties)
+                    )
+            else:
+                properties[es_field_name]["index"] = False
 
     return properties
 
--- a/hail_scripts/elasticsearch/elasticsearch_utils.py
+++ b/hail_scripts/elasticsearch/elasticsearch_utils.py
@@ -100,7 +100,7 @@ def elasticsearch_schema_for_table(table, disable_doc_values_for_fields=(), disa
         logger.info("==> will disable index fields for %s", ", ".join(disable_index_for_fields))
         for es_field_name in disable_index_for_fields:
             if es_field_name not in properties:
-                flattened_fields = [key for key in properties if f"{es_field_name}_" in key]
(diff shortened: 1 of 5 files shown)


hail_scripts/elasticsearch/elasticsearch_utils.py
```diff
@@ -100,11 +100,17 @@ def elasticsearch_schema_for_table(table, disable_doc_values_for_fields=(), disa
         logger.info("==> will disable index fields for %s", ", ".join(disable_index_for_fields))
         for es_field_name in disable_index_for_fields:
             if es_field_name not in properties:
-                raise ValueError(
-                    "'%s' in disable_index_for_fields arg is not in the elasticsearch schema: %s"
-                    % (es_field_name, properties)
-                )
-            properties[es_field_name]["index"] = False
+                flattened_fields = [key for key in properties if f"{es_field_name}_" in key]
```
hanars, line 103: So the key has to start with the prefix not just contain it, so it should not be `f"{es_field_name}_" in key` it should be `key.startswith(f'{es_field_name}_')

hanars (changes requested):

hanars:
looks good t me, @mike-w-wilson thoughts?

mike-w-wilson:
This looks good to me as well!

mike-w-wilson (approved):

hanars (approved):

mike-w-wilson:
@averywpx , @hanars I've tested this on our airflow2 instance and have gotten this error for `rg37_locus`, `sortedTranscriptConsequences` and now `genotypes`. Should I just continue removing the disable_index=True? This seems like it may be an issue outside of the nested fields?
```
'Mapping definition for [genotypes] has unsupported parameters:  [index : false]')
```

hanars:
`genotypes` is also a nested field so yeah I would just remove `disable_index=True` for that. I checked a successful index mapping, and only `sortedTranscriptConsequences` and  `genotypes` have type "nested". `rg37_locus` actually doesn't have a type at all, it has a nested "properties" dict that has the type for each sub-field, I guess because it is a single nested object instead of a list of nested objects (the 2 nested fields also have this "properties" dict). There are no other fields with these nested properties, so I think these should be the only affected fields

mike-w-wilson:
So the index grew ~20% larger..... the 2/16/2022 index is 21.2gb, today's is 27.4gb. I confirmed there is the same number of records in both so there must be something else going on.

mike-w-wilson:
Also, I just reran export, so this is the same variant and project MTs as the previous index

mike-w-wilson:
And now I'm a liar -- the data management page is now showing 21 GB for today... 😕

hanars:
so we're saying this doesn't actually save any space? Or that we aren't testing properly

mike-w-wilson:
It looks like it saved a tiny bit of space, 0.2 GB for this index. I'm not sure why it showed a larger index at first. Maybe it was reporting the size during a temp to data node transfer or something.

mike-w-wilson:
I've pushed a commit with the three fields disabled. Sorry for the confusion yesterday. The PR was tested properly. Even though it didn't save us a ton of space, looks like just 1%, I think we still merge this but would like to hear your thoughts @hanars on adding more complexity to the repo vs data savings. Also, I couldnt find if it is possible to disable indexing on nested field types in a different way but perhaps we would see larger savings if we could figure that out for genotypes in a different PR?

hanars:
I would say my intuition is to merge this as is, because while there is added complexity I don't think it's too much to maintain and there is some data saving. I think for the 3 nested-ish fields, my intuition is we would need to set `"enabled": False` for each sub-field of the nested properties dict, which should only be a couple more lines of code in the handling of `disable_index_for_fields`. I hope those fields will add more noticeable space saving (they are very big) but if we would rather merge this working, tested code now and add that additional functionality as a separate PR that would make sense to me

PR #298, opened 2022-01-13
Remove non-number StrVCTRVRE scores.

ShifaSZ:


sv_pipeline/genome/load_data.py
```diff
@@ -37,7 +37,7 @@
     'pos': lambda rows: rows.locus.position,
     'filters': lambda rows: hl.array(rows.filters.filter(lambda x: x != 'PASS')),
     'algorithms': lambda rows: rows.info.ALGORITHMS,
-    'StrVCTVRE_score': lambda rows: rows.info.StrVCTVRE,
+    'StrVCTVRE_score': lambda rows: hl.parse_float(rows.info.StrVCTVRE),
```
ShifaSZ, line 40: `hl.parse_float(x)` return None if x is not a number.

hanars (approved):

PR #299, opened 2022-01-13
gCNV loading updates

hanars:
Updates the gCNV loading pipeline to conform to the latest file format. Includes adding indication of whether a variant has been previously called, and also adds predicted consequence


sv_pipeline/exome/load_data.py
```diff
@@ -23,13 +23,16 @@
...
+SAMPLE_COL = 'sample_fix'
 NUM_EXON_COL = 'genes_any_overlap_totalexons'
 DEFRAGGED_COL = 'defragmented'
 SC_COL = 'vac'
 SF_COL = 'vaf'
-VAR_NAME_COL = 'name'
+VAR_NAME_COL = 'variant_name'
 GENES_COL = 'genes_any_overlap_ensemble_id'
+PREV_IDENTICAL_COL = 'identical_round1'
+PREV_OVERLAP_COL = 'any_round1'
```
ShifaSZ, line 34: Should it be `any_ovl_in_round1`?

sv_pipeline/exome/load_data.py
```diff
@@ -23,13 +23,16 @@
...
+PREV_MISSING_COL = 'no_ovl_in_round1'
```
ShifaSZ, line 35: How about replacing the `round1` to f'round{ROUND-1}'?

sv_pipeline/exome/load_data.py
```diff
@@ -89,12 +105,19 @@ def _get_seqr_sample_id(raw_sample_id, sample_type='WES'):
     },
     GENES_COL: {
         'field_name': GENES_FIELD,
-        'format': lambda genes: set() if genes == 'None' else {gene.split('.')[0] for gene in genes.split(',')},
+        'format': _parse_genes,
     },
+    PREV_IDENTICAL_COL: {'field_name': 'prev_call', 'format': _parse_prev_call},
```
ShifaSZ, line 110: As suggested in the other PR, how about using `prev_identical` rather than `prev_call`?

sv_pipeline/exome/load_data.py
```diff
@@ -296,19 +326,24 @@ def format_sv(sv):
...
     sv['samples'] = []
+    sv['samples_new_call'] = []
+
+    gene_consequences = {}
     for genotype in sv[GENOTYPES_FIELD]:
         sample_id = genotype['sample_id']
         sv['samples'].append(sample_id)
 
+        if genotype[NEW_CALL_FIELD]:
+            sv['samples_new_call'].append(sample_id)
```
ShifaSZ, line 345: A constant for `samples_new_call` is better.

ShifaSZ (changes requested):

sv_pipeline/exome/load_data.py
hanars: no because thats not what the column is in the file? its not like we get to pick how the delivered data is formatted

sv_pipeline/exome/load_data.py
hanars: no I think its clearer to state that this is previously called, because that what it means. Identical is ambiguous outside of the file - identical to what? and I want to use short field names here which makes `prev_identical_call` and the like too long to use

sv_pipeline/exome/load_data.py
hanars: more consistant this way with how we are handling `sv['samples']` (see the above line) but okay

ShifaSZ (approved):

PR #300, opened 2022-01-19
Add an option for a separated strvctvre input file.

ShifaSZ:


sv_pipeline/genome/load_data.py
```diff
@@ -192,6 +196,13 @@ def export_to_es(rows, input_dataset, project_guid, es_host, es_port, block_size
     )
 
 
+def annotate_from_file(rows, filename, fields):
+    src_rows = load_mt(filename, None, False).rows()
+    src_rows = src_rows.key_by('rsid')
+
+    return rows.annotate(**{k: v(rows[VARIANT_ID], src_rows) for k, v in fields.items()})
```
hanars, line 203: using the genric function makes this really hard to read. Like I said, making it generic is overkill. It would be much easier to understand and maintain if you actually implement this inline with the correct value - I think like
```
return rows.annotate(StrVCTVRE_score=hl.parse_float(src_rows[rows[VARIANT_ID]].info.StrVCTVRE)
```

hanars (changes requested):

sv_pipeline/genome/load_data.py
ShifaSZ: I was thinking about making it generic in the command line which means allowing the user to specify the fields. We agree that is not good. But the function can be generic by using a field mapping parameter. How about I change the parameter `fields` to `field_mappings`?

hanars: no, this is over-engineered and really impossible to read, which will make it impossible to maintain. We do not need a generic `annotate_from_file` function. We need a specific, clear, readable, maintainable `add_stvctre` function. You do not build in extinsibility at the begining. You build in extensibility later, when you actually have a use case worth extending for

hanars (approved):

PR #301, opened 2022-01-25
update gCNV loading to use latest stvctre format

hanars:


sv_pipeline/exome/load_data.py
```diff
@@ -97,7 +97,7 @@ def _parse_prev_call(call):
     DEFRAGGED_COL: {'field_name': DEFRAGGED_FIELD, 'format': lambda val: BOOL_MAP[val]},
     IN_SILICO_COL: {
         'field_name': 'StrVCTVRE_score',
-        'format': lambda val: None if val == 'not_exonic' else float(val),
+        'format': lambda val: None if val.strip() == 'not_exonic' else float(val),
```
ShifaSZ, line 100: There are values in the field like 'less_than_50bp' and 'not_dup_or_del' for the genome SV dataset. Is this the case for the exome data?

ShifaSZ (changes requested):

sv_pipeline/exome/load_data.py
hanars: nope! `not_dup_or_del` is impossible for gCNV because it only calls dups and dels. And I think the calling algorthm for gCNV also can;t actually detect small events, because I never saw `less_than_50bp` as an option

ShifaSZ (approved):

PR #302, opened 2022-01-25
Update the gnomAD SV ID and AF fields.

ShifaSZ:
Change the field names for the gnomAD SV ID and AF. And the SV ID field type has been changed from string to array of string.


sv_pipeline/genome/load_data.py
```diff
@@ -33,8 +33,10 @@
     'end': lambda rows: hl.if_else(hl.is_defined(rows.info.END2), rows.info.END2, rows.info.END),
     'sv_callset_Het': lambda rows: rows.info.N_HET,
     'sv_callset_Hom': lambda rows: rows.info.N_HOMALT,
-    'gnomad_svs_ID': lambda rows: rows.info['gnomad_v2.1_sv_SVID'],
-    'gnomad_svs_AF': lambda rows: rows.info['gnomad_v2.1_sv_AF'],
+    'gnomad_svs_ID': lambda rows: hl.if_else(hl.is_defined(rows.info['gnomAD_V2_SVID']),
```
hanars, line 36: since they took out the `.` from the field name I think it would be more consistant to do `rows.info. gnomAD_V2_SVID` instead of using the bracket notation

hanars (changes requested):

sv_pipeline/genome/load_data.py
ShifaSZ: Okay. Updated.

hanars (approved):

PR #306, opened 2022-02-08
Add creating combined reference dataset with a single code.

ShifaSZ:
The `HmtVar` dataset is not working yet. It might not be able to combine all the datasets together. Hail reports running out of memory when combining all the datasets. It may be because of hail configuration.

### DIFF
--- a/download_and_create_reference_datasets/v02/mito/create_ht__mito_reference_data.py
+++ b/download_and_create_reference_datasets/v02/mito/create_ht__mito_reference_data.py
@@ -1,0 +1,155 @@
+#!/usr/bin/env python3
+
+# Download/load mitochondrial reference datasets, annotate required fields into a hail table,
+# and save it to a file.
+
+import argparse
+import logging
+import json
+import tempfile
+import os
+
+from datetime import datetime
+from functools import reduce
+import hail as hl
+
+logging.basicConfig(format='%(asctime)s %(levelname)-8s %(message)s')
+logger = logging.getLogger()
+
+contig_recoding={'1': 'chr1', '10': 'chr10', '11': 'chr11', '12': 'chr12', '13': 'chr13', '14': 'chr14', '15': 'chr15',
+ '16': 'chr16', '17': 'chr17', '18': 'chr18', '19': 'chr19', '2': 'chr2', '20': 'chr20', '21': 'chr21', '22': 'chr22',
+ '3': 'chr3', '4': 'chr4', '5': 'chr5', '6': 'chr6', '7': 'chr7', '8': 'chr8', '9': 'chr9', 'X': 'chrX', 'Y': 'chrY',
+ 'MT': 'chrM', 'NW_009646201.1': 'chr1'}
+
+CONFIG = {
+    'gnomad': {
+        'path': 'gs://gcp-public-data--gnomad/release/3.1/ht/genomes/gnomad.genomes.v3.1.sites.chrM.ht',
+        'input_type': 'ht',
+        'select': ['AN', 'AC_hom', 'AC_het', 'AF_hom', 'AF_het', 'max_hl']
+    },
+    'mitomap': {
+        'path': 'gs://seqr-reference-data/GRCh38/MITOMAP/Mitomap Confirmed Mutations Feb. 04 2022.tsv',
+        'input_type': 'tsv',
+        'annotate': {
+            'locus': lambda ht: hl.locus('chrM', hl.parse_int32(ht.Allele.first_match_in('m.([0-9]+)')[0])),
+            'alleles': lambda ht: ht.Allele.first_match_in('m.[0-9]+([ATGC]+)>([ATGC]+)'),
+            'pathogenic': lambda ht: hl.is_defined(ht['Associated Diseases'])
+        },
+        'select': ['pathogenic']
+    },
+    'mitimpact': {
+        'path': 'gs://seqr-reference-data/GRCh38/MitImpact/MitImpact_db_3.0.7.txt',  # from https://mitimpact.css-mendel.it/cdn/MitImpact_db_3.0.7.txt.zip',
+        'input_type': 'tsv',
+        'annotate': {
+            'locus': lambda ht: hl.locus('chrM', hl.parse_int32(ht.Start)),
+            'alleles': lambda ht: [ht.Ref, ht.Alt],
+        },
+        'select': ['APOGEE_score']
+    },
+    'hmtvar': {
+        'path': 'gs://seqr-reference-data/GRCh38/HmtVar/HmtVar Jan. 10 2022.json',  # from https://www.hmtvar.uniba.it/api/main/',
+        'input_type': 'json',
+        'annotate': {
+            'locus': lambda ht: hl.locus('chrM', hl.parse_int32(ht.nt_start)),
+            'alleles': lambda ht: [ht.ref_rCRS, ht.alt],
+        },
+        'select': ['disease_score']
+    },
+    'helix': {
+        'path': 'gs://seqr-reference-data/GRCh38/Hilex/HelixMTdb_20200327.tsv',  # from https://helix-research-public.s3.amazonaws.com/mito/HelixMTdb_20200327.tsv',
+        'input_type': 'tsv',
+        'annotate': {
+            'locus': lambda ht: hl.locus('chrM', hl.parse_int32(ht.locus.split(':')[1])),
+            'alleles': lambda ht: ht.alleles.first_match_in('\["([AGTC]+)","([AGTC]+)"\]')
+        },
+        'select': ['counts_hom', 'AF_hom', 'counts_het', 'AF_het', 'max_ARF']
+    },
+    'clinvar': {
+        'path': 'gs://seqr-reference-data/GRCh38/clinvar/clinvar.GRCh38.2022-01-10.vcf.gz',  # from ftp://ftp.ncbi.nlm.nih.gov/pub/clinvar/vcf_GRCh38/clinvar.vcf.gz',
+        'input_type': 'vcf',
+        'annotate': {
+            'ALLELEID': lambda ht: ht.info.ALLELEID,
+            'CLNSIG': lambda ht: ht.info.CLNSIG,
+            'CLNREVSTAT': lambda ht: ht.info.CLNREVSTAT,
+        },
+        'select': ['ALLELEID', 'CLNSIG', 'CLNREVSTAT']
+    },
+    'dbnsfp': {
+        'path': 'gs://seqr-reference-data/GRCh38/all_reference_data/v2/combined_reference_data_grch38-2.0.4.ht',
+        'input_type': 'ht',
+        'select': ['dbnsfp']
+    }
+}
+
+
+def load_hts(datasets):
+    hts = []
+    for dataset in datasets:
+        logger.info(f'Loading dataset {dataset}.')
+        if CONFIG[dataset]['input_type'] == 'ht':
+            ht = hl.read_table(CONFIG[dataset]['path'])
+        elif CONFIG[dataset]['input_type'] == 'tsv':
+            ht = hl.import_table(CONFIG[dataset]['path'])
+        elif CONFIG[dataset]['input_type'] == 'vcf':
+            ht = hl.import_vcf(CONFIG[dataset]['path'], force_bgz=True, contig_recoding=contig_recoding).rows()
+        elif CONFIG[dataset]['input_type'] == 'json':
+            with open(CONFIG[dataset]['path'], 'r') as f:
+                data = json.load(f)
+            with tempfile.mkstemp(suffix='tsv', text=True) as f:
+                header = '\t'.join(list(data[0].keys()))
+                f.write(header+'\n')
+                for row in data:
+                    f.wirte('\t'.join(list(row.values()))+'\n')
+                f_name = f.name
+            ht = hl.import_table(f_name)
+            os.remove(f_name)
+        if 'annotate' in CONFIG[dataset].keys():
+            ht = ht.annotate(**{field: func(ht) for field, func in CONFIG[dataset]['annotate'].items()})
+        ht = ht.annotate(**{dataset: hl.struct(**{field: ht[field] for field in CONFIG[dataset]['select']})})
+        ht = ht.filter(ht.locus.contig == 'chrM')
+        ht = ht.key_by('locus', 'alleles')
+        ht = ht.select(dataset)
+        hts.append(ht)
+    return hts
+
+
+def join_hts(datasets):
+    # Get a list of hail tables and combine into an outer join.
+    hts = load_hts(datasets)
+    joined_ht = reduce((lambda joined_ht, ht: joined_ht.join(ht, 'outer')), hts)
+
+    # Track the dataset we've added as well as the source path.
+    included_dataset = {k: v['path'] for k, v in CONFIG.items() if k in datasets}
+    # Add metadata, but also removes previous globals.
+    joined_ht = joined_ht.select_globals(date=datetime.now().isoformat(),
+                                         datasets=hl.dict(included_dataset))
+    logger.info(joined_ht.describe(str))
+    return joined_ht
+
+
+def run(args):
+    hl.init(default_reference='GRCh38', min_block_size=128,master='local[32]')
+    datasets = args.dataset.split(',')
+    not_supported_dataset = [d for d in datasets if d not in CONFIG.keys()]
+    if len(not_supported_dataset) > 0:
+        logger.error(f'{len(not_supported_dataset)} datasets are not supported: {not_supported_dataset}')
+        return
+
+    logger.info(f'Loading and combining {datasets}')
+    joined_ht = join_hts(datasets)
+
+    logger.info(f'Writing to {args.output_path}')
+    joined_ht.write(args.output_path, overwrite=args.force_write)
+    logger.info('Done')
+
+
+if __name__ == "__main__":
+
+    datasets = ','.join(CONFIG.keys())
+    parser = argparse.ArgumentParser()
+    parser.add_argument('-d', '--dataset', help=f'Reference dataset list, separated with commas, e.g. {datasets}', required=True)
+    parser.add_argument('-o', '--output-path', help='Path and file name for the combined reference dataset', required=True)
+    parser.add_argument('-f', '--force-write', help='Force write to exist output file', action='store_true')
+    args = parser.parse_args()
+
+    run(args)


hanars (changes requested):
This seems to be loosely based on but somewhat different from the wriate_combined_reference_table file. Rather than kind of doing the same thing but kind of not and duplicating a lot of efforts, you should use a shared utility from that file. I think converting the existing file to a class that defines its own `CONFIG`, and then superclassing it here would be a good approach

In the future, when I say "base your code on this fle" I do not mean "copy the file and change it". I mean, "share code with the file as much as possible". Copy-pasting is almost always the wrong approach.

hanars:
An example of what the class for `write_combined_reference_table` would look like is as follows:

```
class WriteCombinedReferenceData(object):
  CONFIG = {...}
  ...
  @classmethod
  def get_ht(cls, dataset, reference_genome):
    config = cls.CONFIG[dataset][reference_genome]
    ...
  
  ...
  @classmethod
  def run(cls, args):
    ...
    cls.join_hts([...])

if __name__ == "__main__":
  ...
  WriteCombinedReferenceData.run(args)

``` 

and then for mito all you *should* need to do is 
```
class WriteCombinedMitoReferenceData(WriteCombinedReferenceData):
  CONFIG = {...} # the mito config goes 

if __name__ == "__main__":
  ...
  WriteCombinedMitoReferenceData.run(args)
```

hanars:
Alternatively, you can leave `write_combined_reference_table` as it is and just add the mito data sources to the main config. Since the script specifically  sets which subset of configs it wants to use that won't break the exisiting script, and then for the mito you could import functionality and do something like 

```
from hail_scripts.write_combined_reference_table import join_hts

if __name__ == "__main__":
   ...
   join_hts([...], args)
```

ShifaSZ:
I updated `download_and_create_reference_datasets/v02/hail_scripts/write_combined_reference_data_ht.py` to support writing the combined mitochondrial reference data. Basically, its functionalities for nuclear SNPs don't change.

hanars:
so this is still not what we discussed. As I have repeatedly mentioned, we want this to match the usage of/ share code with `write_combined_reference_data_ht`. That means there is one `create_ht__<dataset>` script per dataset type, and then we will have a new file `write_combined_mito_reference_data_ht.py` that basically just imports `join_hts` and calls it with the datasets we care about for mito. If you really want to add a helper that can allow you to call multiple `create_ht__<dataset>` at the same time you can do that, but that needs to be separate from the actual script for writing the combined file, and there still needs to be a way to run them independently. We discussed in standup that going forward we need to be able to easily update just one resource without needing to re-download the others. If you are still confused, please set up time to meet with me

download_and_create_reference_datasets/v02/hail_scripts/write_combined_reference_data_ht.py
```diff
@@ -294,25 +341,42 @@ def join_hts(datasets, coverage_datasets=[], reference_genome='37'):
...
+    output_path = os.path.join(OUTPUT_TEMPLATE.format(genome_version=args.build, version=VERSION))
+    logger.info(f'Writing to {output_path}')
+    joined_ht.write(os.path.join(output_path))
+
 
 if __name__ == "__main__":
 
     parser = argparse.ArgumentParser()
     parser.add_argument('-b', '--build', help='Reference build, 37 or 38', choices=["37", "38"], required=True)
+    parser.add_argument('-m', '--mitochondrial', help='Create combined mitochondrial reference dataset hail table', action='store_true')
```
hanars, line 378: as I have repeatedly stated, we need a separate script for writing mito data that imports the functions it needs from here. adding extra entries to the config is fine, but thats the only change to make to this file

download_and_create_reference_datasets/v02/mito/write_combined_mito_reference_data_hts.py
```diff
@@ -0,0 +1,191 @@
...
+        'input_type': 'vcf',
+        'output_path': 'gs://seqr-reference-data/GRCh38/mitochondrial/clinvar/clinvar.GRCh38.chrM.ht',
+        'annotate': {
+            'ALLELEID': lambda ht: ht.info.ALLELEID,
+            'CLNSIG': lambda ht: ht.info.CLNSIG,
+            'CLNREVSTAT': lambda ht: ht.info.CLNREVSTAT,
+        },
+    },
+    'dbnsfp': {
+        'input_path': 'gs://seqr-reference-data/GRCh38/all_reference_data/v2/combined_reference_data_grch38-2.0.4.ht',
```
hanars, line 78: I don't really want to pay to store a subset of the exact same data twice. There is no good reason you can't tweak the config for `dbnsfp_mito` to use the full dbnsfp ht as its input

download_and_create_reference_datasets/v02/hail_scripts/write_combined_reference_data_ht.py
ShifaSZ: I see what you mean. I can change the `write_combined_mito_reference_data_hts` to call the joined_ht() from `write_combined_reference_data`. I still need to change the `joined_ht()` function a little bit to write to a different output file.

download_and_create_reference_datasets/v02/mito/write_combined_mito_reference_data_hts.py
ShifaSZ: It is possible only if we add a `filter` configuration and we update the `joined_ht()` function to filter out the variants based on the configuration.

hanars: sure that sounds fine

hanars: No. We really want to keep single source files for the reference data, the exact same way we already do for all the other reference data. In the future if we need to update one reference source and keep the other the same we want to have all the other reference sources saved from the last time so we don't need to recompute them. Again, if you are having trouble understanding the approach we are taking here, I would ask you to schedule time to meet and discuss

hanars: As much as possible we want this to maintain the exact same behavior we already have for reference data, and this whole pipeline in general. In general the answer to "should I do this differently" is always no. The small amount of possible savings will never be worth the technical and mental overhead of maintaining two different ways of doing what is basically the same thing. The mito pipeline NEEDS to run the same way a the regular SNP pipeline wherever possible. If you think there are better ways to do things, make a ticket in the backlog and when we have bandwidth in the future we can make improvements to both the main pipeline and the mito pipeline through shared code. But for now, unless there is an actual data reason things need to be different, they need to be the same

download_and_create_reference_datasets/v02/hail_scripts/write_combined_reference_data_ht.py
```diff
@@ -168,6 +172,67 @@
...
+        '38': {
+            'path': 'gs://seqr-reference-data/GRCh38/mitochondrial/clinvar/clinvar.GRCh38.chrM.ht',
+            'select': {
+                'allele_id': 'ALLELEID',
+                'clinical_significance': 'CLNSIG',
+                'gold_stars': 'CLNREVSTAT'
+            }
+        }
+    },
+    'dbnsfp_mito': {
```
hanars, line 222: I think it would be better to re-use the SNP dbnsfp configuration. Rather than defining it as a standalone entry here, I think it would be better to do the following after the `CONFIG` definition
```
CONFIG['dbnsfp_mito'] = deepcopy(CONFIG['dbnsfp'] )
CONFIG['dbnsfp_mito']['filter'] = lambda ht: ht.locus.contig == 'chrM'
```

download_and_create_reference_datasets/v02/hail_scripts/write_combined_reference_data_ht.py
```diff
@@ -294,20 +362,20 @@ def join_hts(datasets, coverage_datasets=[], reference_genome='37'):
...
-    print('Writing to %s' % output_path)
-
-    joined_ht.write(os.path.join(output_path))
+    return joined_ht
 
 
 def run(args):
-    hl._set_flags(no_whole_stage_codegen='1') # hail 0.2.78 hits an error on the join, this flag gets around it 
-    join_hts(['cadd', '1kg', 'mpc', 'eigen', 'dbnsfp', 'topmed', 'primate_ai', 'splice_ai', 'exac',
+    # hl._set_flags(no_whole_stage_codegen='1') # hail 0.2.78 hits an error on the join, this flag gets around it
```
hanars, line 369: this should either stay the same or be fully removed, but we don't want to just keep it around commented out

download_and_create_reference_datasets/v02/hail_scripts/write_combined_reference_data_ht.py
```diff
@@ -1,4 +1,5 @@
 import argparse
+import logging
```
hanars, line 2: what was the reason for adding `logging`? It doesn't look like you actually switched all the `print` statements everywhere, which means we now just have 2 inconsistent ways of logging instead of one

download_and_create_reference_datasets/v02/mito/utils.py
```diff
@@ -0,0 +1,102 @@
...
+
+    with open(local_file_path, 'wb') as f:
+        f.writelines(input_iter)
+
+    input_iter.close()
+
+    return local_file_path
+
+
+def convert_json2tsv(json_path):
```
hanars, line 49: using the number 2 to mean to word "to" makes code a bit harder to read - just use `convert_json_to_tsv`, or even just `json_to_tsv`

download_and_create_reference_datasets/v02/mito/utils.py
```diff
@@ -0,0 +1,102 @@
...
+logging.basicConfig(format='%(asctime)s %(levelname)-8s %(message)s', level='INFO')
+logger = logging.getLogger(__name__)
+
+contig_recoding={'1': 'chr1', '10': 'chr10', '11': 'chr11', '12': 'chr12', '13': 'chr13', '14': 'chr14', '15': 'chr15',
+ '16': 'chr16', '17': 'chr17', '18': 'chr18', '19': 'chr19', '2': 'chr2', '20': 'chr20', '21': 'chr21', '22': 'chr22',
+ '3': 'chr3', '4': 'chr4', '5': 'chr5', '6': 'chr6', '7': 'chr7', '8': 'chr8', '9': 'chr9', 'X': 'chrX', 'Y': 'chrY',
+ 'MT': 'chrM', 'NW_009646201.1': 'chr1'}
+
+
+def download_file(url, to_dir=tempfile.gettempdir(), verify=True):
```
hanars, line 21: it looks like `load` is the only function you actually use outside of this file, so I think it would be best to make the rest of the methods in the file private (i.e. `_download_file`)

download_and_create_reference_datasets/v02/mito/utils.py
```diff
@@ -0,0 +1,102 @@
...
+        header = '\t'.join(data[0].keys())
+        f.write(header + '\n')
+        for row in data:
+            f.write('\t'.join([str(v) for v in row.values()]) + '\n')
+    return tsv_path
+
+
+def load_ht(config, force_write=True):
+    logger.info(f'Downloading dataset from {config["input_path"]}.')
+    dn_path = download_file(config['input_path'], verify=config.get('verify_ssl'))
```
hanars, line 63: the `verify` logic above is a bit unnecessarily complicated. I would either change this to `verify=config.get('verify_ssl'), True)` so the default is as expected, or I would actually change the config from `verify_ssl ` to `skip_ verify_ssl ` and then use `skip_verify` as the keyword arg so it has falsy behavior when unset without needing complicated `if None` logic

download_and_create_reference_datasets/v02/mito/utils.py
```diff
@@ -0,0 +1,102 @@
...
+    elif config['input_type'] == 'json':
+        tsv_path = convert_json2tsv(dn_path)
+        ht = hl.import_table(tsv_path, types=types)
+    else:
+        ht = hl.import_table(dn_path, types=types)
+
+    if config.get('annotate'):
+        ht = ht.annotate(**{field: func(ht) for field, func in config['annotate'].items()})
+
+    ht = ht.filter(ht.locus.contig == 'chrM')
```
hanars, line 83: given this, I think it would be better to rename this function `_load_mito_ht` so no one mistakenly thinks they can just use it for SNPs

download_and_create_reference_datasets/v02/mito/utils.py
```diff
@@ -0,0 +1,102 @@
...
+    tsv_path = f'{json_path[:-5]}.tsv' if json_path.endswith('.json') else f'{json_path}.tsv'
+    with open(tsv_path, 'w') as f:
+        header = '\t'.join(data[0].keys())
+        f.write(header + '\n')
+        for row in data:
+            f.write('\t'.join([str(v) for v in row.values()]) + '\n')
+    return tsv_path
+
+
+def load_ht(config, force_write=True):
```
hanars, line 61: If you have a function that is always being called with all of the arguments, its actually better not to define the arguments as keyword arguments. You can always add defaults later if someone decides to use it in another way, but in general the assumption is that if a function has a keyword argument its being used somewhere and you need to be careful about changing it, so if thats not the case its better to not set one at all

download_and_create_reference_datasets/v02/mito/utils.py
```diff
@@ -0,0 +1,102 @@
...
+    ht.write(config['output_path'], overwrite=force_write)
+    logger.info('Done')
+
+
+def load(config):
+    parser = argparse.ArgumentParser()
+    parser.add_argument('-f', '--force-write', help='Force write to an existing output file', action='store_true')
+    args = parser.parse_args()
+
+    # If there are out-of-memory error, set the environment variable with the following command
```
hanars, line 97: this seems like its copy-pasted from somewhere - if so, can you add a link to the source in the comment? It means in the future its easier to check and see if a workaround is still necessary/ recommended

hanars (changes requested):

download_and_create_reference_datasets/v02/hail_scripts/write_combined_reference_data_ht.py
ShifaSZ: Remove it.

download_and_create_reference_datasets/v02/hail_scripts/write_combined_reference_data_ht.py
ShifaSZ: The logging is easier in controlling the level of the messages to be displayed.  But changing all the prints to logging is too distractive in this PR.

hanars: but since you are only using logging.info in this file I think its okay to just keep with print for consistency

download_and_create_reference_datasets/v02/mito/utils.py
ShifaSZ: Good point. I will change it to `skip_verify_ssl`.

download_and_create_reference_datasets/v02/mito/utils.py
ShifaSZ: I see. Remove the default value.

download_and_create_reference_datasets/v02/mito/utils.py
ShifaSZ: It is about the memory size setting for Hail. I occasionally found the environment variable PYSPARK_SUBMIT_ARGS setting in a [forum](https://discuss.hail.is/t/java-heap-space-out-of-memory/1575/6). I can't find it in Hail docs. So, I doc it here to remind whoever may need it later.

hanars: cool, so just add the link to that forum in the comment here

download_and_create_reference_datasets/v02/hail_scripts/write_combined_reference_data_ht.py
ShifaSZ: The logger is only used for once. I will change it back to print.

download_and_create_reference_datasets/v02/mito/utils.py
ShifaSZ: I found that it doesn't require any special settings to import each specific mito reference dataset. So, I removed the comments and the corresponding parameters for `hl.init()`. But the settings are still required for combining all reference datasets.

download_and_create_reference_datasets/v02/mito/write_mito_clinvar_ht.py
```diff
@@ -0,0 +1,19 @@
+import hail as hl
+
+from download_and_create_reference_datasets.v02.mito.utils import load
+from hail_scripts.utils.clinvar import CLINVAR_GOLD_STARS_LOOKUP
+
+CONFIG = {
+    'input_path': 'http://ftp.ncbi.nlm.nih.gov/pub/clinvar/vcf_GRCh38/clinvar.vcf.gz',
```
hanars, line 7: this looks like the same exact source for the SNP mito reference data. Given that we should not be re-parsing and storing a subsetted duplicate of this data - instead we should update the CONFIG to reference the already parsed clinvar flile at `gs://seqr-reference-data/GRCh38/clinvar/clinvar.GRCh38.ht`

download_and_create_reference_datasets/v02/mito/write_mito_mitomap_ht.py
```diff
@@ -0,0 +1,19 @@
+import hail as hl
+
+from download_and_create_reference_datasets.v02.mito.utils import load
+
+CONFIG = {
+    # The data source is a web page. So we download it manually and save them to the GCS.
```
hanars, line 6: this comment is insuffiecient. It needs to be detailed and include the exact URL where the data was copied from

download_and_create_reference_datasets/v02/mito/write_mito_mitomap_ht.py
```diff
@@ -0,0 +1,19 @@
...
+    'input_path': 'https://storage.googleapis.com/seqr-reference-data/GRCh38/mitochondrial/MITOMAP/Mitomap%20Confirmed%20Mutations%20Feb.%2004%202022.tsv',
+    'input_type': 'tsv',
+    'output_path': 'gs://seqr-reference-data/GRCh38/mitochondrial/MITOMAP/Mitomap Confirmed Mutations Feb. 04 2022.ht',
+    'annotate': {
+        'locus': lambda ht: hl.locus('chrM', hl.parse_int32(ht.Allele.first_match_in('m.([0-9]+)')[0])),
+        'alleles': lambda ht: ht.Allele.first_match_in('m.[0-9]+([ATGC]+)>([ATGC]+)'),
+        'pathogenic': lambda ht: hl.is_defined(ht['Associated Diseases'])
```
hanars, line 13: I thought we were using a file that only had the pathogenic mitopmap variants? if so, how is this column helpful?

hanars (changes requested):

download_and_create_reference_datasets/v02/mito/write_mito_clinvar_ht.py
ShifaSZ: Good point. But, the hail table `gs://seqr-reference-data/GRCh38/clinvar/clinvar.GRCh38.ht` doesn't have mitochondrial data. I found that the `contig_recoding` parameter they used for importing the Clinvar VCF file doesn't work for mitochondrial. The entry in the dict `contig_recoding` should be 'MT': 'chrM'. But that in the code for SNP reference data is 'M': 'chrM'. That's why the mitochondrial variants were not loaded. I could update the SNP reference data code and regenerate the Clinvar Hail table but it could break the SNP pipeline because it doesn't filter out mitochondrial data.

download_and_create_reference_datasets/v02/mito/write_mito_mitomap_ht.py
ShifaSZ: The file contains the data copied and pasted from the web page. If we reformat the fields of `locus`, `alleles`, and `pathogenic` before saving the file, then this column will become much simpler.

download_and_create_reference_datasets/v02/mito/write_mito_clinvar_ht.py
hanars: My intuition is we want to fix this for the SNP pipeline, as that does occasionally include mitochondrial variants and it would be good for those to have the correct clinvar annotations, but I would want to confirm with @mike-w-wilson

download_and_create_reference_datasets/v02/mito/write_mito_mitomap_ht.py
hanars: but I don't understand what purpose the `pathogenic` field serves at all, the file should only include pathogenic variants in the first place so what is the point of having an entry for every variant telling us something we already know?

ShifaSZ: I see what you mean now. The values for the `pathogenic` field are always `true` in the data file. But we still need it when joining with other reference data. We use the `outer` join function of Hail. This means that the missing rows will have a value of `missing` for the field.
In the end, the field can have two different values (`true` or `missing`).

hanars: Okay, but then couldn't you just make it a hail literal for true, instead of this expression which implies theres multiple possibilities for the value?

hanars (approved):
This looks good to me, @mike-w-wilson can you take a look?

hanars:
You can not merge PRs to this repo without Mike's approval. My approval was very much conditional on him approving. Revert this immediately

hanars:
actually I reverted it, new PR is here: https://github.com/broadinstitute/hail-elasticsearch-pipelines/pull/314

PR #308, opened 2022-02-14
bump minimum required hail version

hanars:
The requirements.txt is used to build a docker image for local seqr users to run this pipeline, and the pinned version of hail is pretty old and is now causing bugs (https://github.com/broadinstitute/seqr/issues/2460)


mike-w-wilson (approved):
LGTM

PR #310, opened 2022-02-23
Mitochondrial data loading pipeline

ShifaSZ:
This PR loads mitochondrial data from a Hail Matrix Table, annotates with the reference data and high constraint region data and exports the data to an Elasticsearch index.

### DIFF
--- a/luigi_pipeline/lib/model/mito_mt_schema.py
+++ b/luigi_pipeline/lib/model/mito_mt_schema.py
@@ -1,0 +1,257 @@
+import hail as hl
+
+from lib.model.base_mt_schema import BaseMTSchema, row_annotation, RowAnnotationOmit
+from hail_scripts.computed_fields import variant_id
+from hail_scripts.computed_fields import vep
+
+
+class SeqrMitoSchema(BaseMTSchema):
+
+    def __init__(self, *args, ref_data, **kwargs):
+        self._ref_data = ref_data
+
+        # See _selected_ref_data
+        self._selected_ref_data_cache = None
+
+        super().__init__(*args, **kwargs)
+
+    def set_mt(self, mt):
+        super().set_mt(mt)
+        # set this to None, and the @property _selected_ref_data
+        # can populate it if it gets used after each MT update.
+        self._selected_ref_data_cache = None
+
+    @property
+    def _selected_ref_data(self):
+        """
+        Reuse `self._ref_data[self.mt.row_key]` for all annotations.
+        We'll use the @property _selected_ref_data to access this
+        value, and lazily cache it (so as not to compute it if it
+        doesn't get get accessed after an update to self.mt).
+
+        Returns: self._ref_data[self.mt.row_key]
+        """
+        if not self._selected_ref_data_cache:
+            self._selected_ref_data_cache = self._ref_data[self.mt.row_key]
+        return self._selected_ref_data_cache
+
+    @row_annotation()
+    def vep(self):
+        return self.mt.vep
+
+    @row_annotation()
+    def rsid(self):
+        return self.mt.rsid
+
+    @row_annotation()
+    def filters(self):
+        return self.mt.filters
+
+    @row_annotation()
+    def common_low_heteroplasmy(self):
+        return self.mt.common_low_heteroplasmy
+
+    # @row_annotation()
+    # def originalAltAlleles(self):
+        # TODO: This assumes we annotate `locus_old` in this code because `split_multi_hts` drops the proper `old_locus`.
+        # If we can get it to not drop it, we should revert this to `old_locus`
+    #     return variant_id.get_expr_for_variant_ids(self.mt.locus_old, self.mt.alleles_old)
+
+    @row_annotation(name='sortedTranscriptConsequences', fn_require=vep)
+    def sorted_transcript_consequences(self):
+        return vep.get_expr_for_vep_sorted_transcript_consequences_array(self.mt.vep)
+
+    @row_annotation(name='docId')
+    def doc_id(self, length=512):
+        return variant_id.get_expr_for_variant_id(self.mt, length)
+
+    @row_annotation(name='variantId')
+    def variant_id(self):
+        return variant_id.get_expr_for_variant_id(self.mt)
+
+    @row_annotation()
+    def contig(self):
+        return variant_id.get_expr_for_contig(self.mt.locus)
+
+    @row_annotation()
+    def pos(self):
+        return variant_id.get_expr_for_start_pos(self.mt)
+
+    @row_annotation()
+    def start(self):
+        return variant_id.get_expr_for_start_pos(self.mt)
+
+    @row_annotation()
+    def end(self):
+        return variant_id.get_expr_for_end_pos(self.mt)
+
+    @row_annotation()
+    def ref(self):
+        return variant_id.get_expr_for_ref_allele(self.mt)
+
+    @row_annotation()
+    def alt(self):
+        return variant_id.get_expr_for_alt_allele(self.mt)
+
+    @row_annotation()
+    def xpos(self):
+        return variant_id.get_expr_for_xpos(self.mt.locus)
+
+    @row_annotation()
+    def xstart(self):
+        return variant_id.get_expr_for_xpos(self.mt.locus)
+
+    @row_annotation()
+    def xstop(self):
+        return variant_id.get_expr_for_xpos(self.mt.locus) + hl.len(variant_id.get_expr_for_ref_allele(self.mt)) - 1
+
+    @row_annotation(fn_require=sorted_transcript_consequences)
+    def domains(self):
+        return vep.get_expr_for_vep_protein_domains_set_from_sorted(
+            self.mt.sortedTranscriptConsequences)
+
+    @row_annotation(name='transcriptConsequenceTerms', fn_require=sorted_transcript_consequences)
+    def transcript_consequence_terms(self):
+        return vep.get_expr_for_vep_consequence_terms_set(self.mt.sortedTranscriptConsequences)
+
+    @row_annotation(name='transcriptIds', fn_require=sorted_transcript_consequences)
+    def transcript_ids(self):
+        return vep.get_expr_for_vep_transcript_ids_set(self.mt.sortedTranscriptConsequences)
+
+    @row_annotation(name='mainTranscript', fn_require=sorted_transcript_consequences)
+    def main_transcript(self):
+        return vep.get_expr_for_worst_transcript_consequence_annotations_struct(
+            self.mt.sortedTranscriptConsequences)
+
+    @row_annotation(name='geneIds', fn_require=sorted_transcript_consequences)
+    def gene_ids(self):
+        return vep.get_expr_for_vep_gene_ids_set(self.mt.sortedTranscriptConsequences)
+
+    @row_annotation(name='codingGeneIds', fn_require=sorted_transcript_consequences)
+    def coding_gene_ids(self):
+        return vep.get_expr_for_vep_gene_ids_set(self.mt.sortedTranscriptConsequences, only_coding_genes=True)
+
+    @row_annotation()
+    def gnomad(self):
+        return self._selected_ref_data.gnomad_mito
+
+    @row_annotation()
+    def mitomap(self):
+        return self._selected_ref_data.mitomap
+
+    @row_annotation(name='APOGEE_score')
+    def mitimpact(self):
+        return self._selected_ref_data.mitimpact
+
+    @row_annotation()
+    def hmtvar(self):
+        return self._selected_ref_data.hmtvar
+
+    @row_annotation()
+    def helix(self):
+        return self._selected_ref_data.helix_mito
+
+    @row_annotation()
+    def clinvar(self):
+        return self._selected_ref_data.clinvar_mito
+
+    @row_annotation()
+    def dbnsfp(self):
+        return self._selected_ref_data.dbnsfp_mito
+
+
+class SeqrMitoVariantSchema(SeqrMitoSchema):
+
+    @row_annotation(name='AC')
+    def ac(self):
+        return self.mt.AC_hom
+
+    @row_annotation(name='AF')
+    def af(self):
+        return self.mt.AF_hom
+
+    @row_annotation(name='AN')
+    def an(self):
+        return self.mt.AN
+
+
+class SeqrMitoGenotypesSchema(BaseMTSchema):
+
+    @row_annotation()
+    def genotypes(self):
+        return hl.agg.collect(hl.struct(**self._genotype_fields()))
+
+    @row_annotation(fn_require=genotypes)
+    def samples_no_call(self):
+        return self._genotype_filter_samples(lambda g: g.num_alt == -1)
+
+    @row_annotation(fn_require=genotypes)
+    def samples_num_alt(self, start=1, end=3, step=1):
+        return hl.struct(**{
+            '%i' % i: self._genotype_filter_samples(lambda g: g.num_alt == i)
+            for i in range(start, end, step)
+        })
+
+    @row_annotation(fn_require=genotypes)
+    def samples_gq(self, start=0, end=95, step=5):
+        # struct of x_to_y to a set of samples in range of x and y for gq.
+        return hl.struct(**{
+            '%i_to_%i' % (i, i+step): self._genotype_filter_samples(lambda g: ((g.gq >= i) & (g.gq < i+step)))
+            for i in range(start, end, step)
+        })
+
+    @row_annotation(fn_require=genotypes)
+    def samples_hl(self, start=0, end=45, step=5):
+        # struct of x_to_y to a set of samples in range of x and y for ab.
+        return hl.struct(**{
+            '%i_to_%i' % (i, i+step): self._genotype_filter_samples(
+                lambda g: ((g.num_alt == 1) & ((g.hl*100) >= i) & ((g.hl*100) < i+step))
+            )
+            for i in range(start, end, step)
+        })
+
+    def _genotype_filter_samples(self, filter):
+        # Filter on the genotypes.
+        return hl.set(self.mt.genotypes.filter(filter).map(lambda g: g.sample_id))
+
+    def _genotype_fields(self):
+        # Convert the mt genotype entries into num_alt, gq, ab, dp, and sample_id.
+        is_called = hl.is_defined(self.mt.GT)
+        return {
+            'num_alt': hl.cond(is_called, hl.cond(self.mt.HL>=0.95, 2, hl.cond(self.mt.HL>=0.01, 1, 0)), -1),
+            'gq': hl.cond(is_called, self.mt.MQ, 0),
+            'hl': hl.cond(is_called, self.mt.HL, 0),
+            'mito_cn': self.mt.mito_cn,
+            'contamination': self.mt.contamination,
+            'dp': hl.cond(is_called, hl.int(hl.min(self.mt.DP, 32000)), hl.null(hl.tfloat)),
+            'sample_id': self.mt.s
+        }
+
+
+class SeqrMitoVariantsAndGenotypesSchema(SeqrMitoVariantSchema, SeqrMitoGenotypesSchema):
+    """
+    Combined variant and genotypes.
+    """
+
+    @staticmethod
+    def elasticsearch_row(ds):
+        """
+        Prepares the mt to export using ElasticsearchClient V02.
+        - Flattens nested structs
+        - drops locus and alleles key
+
+        TODO:
+        - Call validate
+        - when all annotations are here, whitelist fields to send instead of blacklisting.
+        :return:
+        """
+        # Converts a mt to the row equivalent.
+        if isinstance(ds, hl.MatrixTable):
+            ds = ds.rows()
+        # Converts nested structs into one field, e.g. {a: {b: 1}} => a.b: 1
+        table = ds.drop('vep').flatten()
+        # When flattening, the table is unkeyed, which causes problems because our locus and alleles should not
+        # be normal fields. We can also re-key, but I believe this is computational?
+        table = table.drop(table.locus, table.alleles)
+
+        return table
--- a/luigi_pipeline/lib/model/mito_mt_schema.py
+++ b/luigi_pipeline/lib/model/mito_mt_schema.py
@@ -1,6 +1,6 @@
 import hail as hl
 
-from lib.model.base_mt_schema import BaseMTSchema, row_annotation, RowAnnotationOmit
+from lib.model.base_mt_schema import BaseMTSchema, row_annotation
 from hail_scripts.computed_fields import variant_id
 from hail_scripts.computed_fields import vep
 
@@ -254,10 +254,6 @@ class SeqrMitoVariantsAndGenotypesSchema(SeqrMitoVariantSchema, SeqrMitoGenotype
         Prepares the mt to export using ElasticsearchClient V02.
         - Flattens nested structs
         - drops locus and alleles key
-
-        TODO:
-        - Call validate
-        - when all annotations are here, whitelist fields to send instead of blacklisting.
         :return:
         """
         # Converts a mt to the row equivalent.
@@ -266,7 +262,7 @@ class SeqrMitoVariantsAndGenotypesSchema(SeqrMitoVariantSchema, SeqrMitoGenotype
         # Converts nested structs into one field, e.g. {a: {b: 1}} => a.b: 1
         table = ds.drop('vep').flatten()
         # When flattening, the table is unkeyed, which causes problems because our locus and alleles should not
-        # be normal fields. We can also re-key, but I believe this is computational?
+        # be normal fields.
         table = table.drop(table.locus, table.alleles)
 
         return table
--- a/luigi_pipeline/lib/model/mito_mt_schema.py
+++ b/luigi_pipeline/lib/model/mito_mt_schema.py
@@ -7,8 +7,9 @@ from hail_scripts.computed_fields import vep
 
 class SeqrMitoSchema(BaseMTSchema):
 
-    def __init__(self, *args, ref_data, **kwargs):
+    def __init__(self, *args, ref_data, high_constraint_region, **kwargs):
         self._ref_data = ref_data
+        self._high_constraint_region = high_constraint_region
 
         # See _selected_ref_data
         self._selected_ref_data_cache = None
@@ -46,12 +47,6 @@ class SeqrMitoSchema(BaseMTSchema):
     @row_annotation()
     def filters(self):
         return self.mt.filters
-
-    # @row_annotation()
-    # def originalAltAlleles(self):
-        # TODO: This assumes we annotate `locus_old` in this code because `split_multi_hts` drops the proper `old_locus`.
-        # If we can get it to not drop it, we should revert this to `old_locus`
-    #     return variant_id.get_expr_for_variant_ids(self.mt.locus_old, self.mt.alleles_old)
 
     @row_annotation(name='sortedTranscriptConsequences', fn_require=vep)
     def sorted_transcript_consequences(self):
@@ -189,6 +184,10 @@ class SeqrMitoVariantSchema(SeqrMitoSchema):
     @row_annotation()
     def mitotip_trna_prediction(self):
         return self.mt.mitotip_trna_prediction
+
+    @row_annotation()
+    def high_constraint_region(self):
+        return hl.is_defined(self._high_constraint_region[self.mt.locus])
 
 
 class SeqrMitoGenotypesSchema(BaseMTSchema):
--- a/luigi_pipeline/lib/model/mito_mt_schema.py
+++ b/luigi_pipeline/lib/model/mito_mt_schema.py
@@ -47,10 +47,6 @@ class SeqrMitoSchema(BaseMTSchema):
     def filters(self):
         return self.mt.filters
 
-    @row_annotation()
-    def common_low_heteroplasmy(self):
-        return self.mt.common_low_heteroplasmy
-
     # @row_annotation()
     # def originalAltAlleles(self):
         # TODO: This assumes we annotate `locus_old` in this code because `split_multi_hts` drops the proper `old_locus`.
@@ -139,11 +135,11 @@ class SeqrMitoSchema(BaseMTSchema):
     def mitomap(self):
         return self._selected_ref_data.mitomap
 
-    @row_annotation(name='APOGEE_score')
+    @row_annotation(name='APOGEE')
     def mitimpact(self):
         return self._selected_ref_data.mitimpact
 
-    @row_annotation()
+    @row_annotation(name='HmtVar')
     def hmtvar(self):
         return self._selected_ref_data.hmtvar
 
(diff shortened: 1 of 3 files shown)


luigi_pipeline/lib/model/mito_mt_schema.py
```diff
@@ -0,0 +1,268 @@
...
+        doesn't get get accessed after an update to self.mt).
+
+        Returns: self._ref_data[self.mt.row_key]
+        """
+        if not self._selected_ref_data_cache:
+            self._selected_ref_data_cache = self._ref_data[self.mt.row_key]
+        return self._selected_ref_data_cache
+
+    @row_annotation()
+    def vep(self):
```
hanars, line 40: most of the function here look identical to those defined in ` SeqrSchema`. I think it would be better to, in the file where `SeqrSchema` is defined, split the class into `SeqrCoreSchema` which inherits from `BaseMTSchema` and has everything thats shared with mito and then have `SeqrSchema` inherit from `SeqrCoreSchema` and only have the fields not used by mito

luigi_pipeline/lib/model/mito_mt_schema.py
```diff
@@ -0,0 +1,268 @@
...
+    @row_annotation()
+    def mitotip_trna_prediction(self):
+        return self.mt.mitotip_trna_prediction
+
+    @row_annotation()
+    def high_constraint_region(self):
+        return hl.is_defined(self._high_constraint_region[self.mt.locus])
+
+
+class SeqrMitoGenotypesSchema(BaseMTSchema):
```
hanars, line 193: This can probably inherit from `SeqrGenotypesSchema` instead and just override the methods like `_genotype_fields` that are actually different

luigi_pipeline/seqr_mito_loading.py
```diff
@@ -0,0 +1,109 @@
...
+import hail as hl
+
+from lib.hail_tasks import HailMatrixTableTask, HailElasticSearchTask
+from lib.model.mito_mt_schema import SeqrMitoVariantSchema, SeqrMitoGenotypesSchema, SeqrMitoVariantsAndGenotypesSchema
+import seqr_loading
+
+logger = logging.getLogger(__name__)
+
+
+class SeqrMitoVariantMTTask(seqr_loading.SeqrVCFToMTTask):
```
hanars, line 16: what is the purpose of this unoptimized task? We have the unoptimized regular loadnig task for local installs but thats out of scope for mito

luigi_pipeline/seqr_mito_loading.py
```diff
@@ -0,0 +1,109 @@
...
+        if self.subset_path:
+            mt = self.subset_samples_and_variants(mt, self.subset_path)
+
+        mt = SeqrMitoGenotypesSchema(mt).annotate_all(overwrite=True).select_annotated_mt()
+
+        mt.describe()
+        mt.write(self.output().path, stage_locally=True, overwrite=True)
+
+
+class SeqrMTToESOptimizedTask(HailElasticSearchTask):
```
hanars, line 84: this looks identical to `SeqrMTToESOptimizedTask` except for a couple different class names. Whenever you find yourself copy-pasting large amounts of code, it almost certainly means you should be using inheritance instead. Define a base class and inherit as needed:
```
class BaseMTToESOptimizedTask(HailElasticSearchTask):

   ...
    def requires(self):
        return [self.VCF_TO_VARIANT_TASK(), self.VCF_TO_GENOTYPES_TASK()]
    ...
    def run(self):
        ...
        row_ht = self.VARIANT_GENOTYPES_SCHEMA.elasticsearch_row(row_ht)

...
class SeqrMTToESOptimizedTask(BaseMTToESOptimizedTask):
    VCF_TO_VARIANT_TASK = SeqrMitoVariantMTTask
    VCF_TO_GENOTYPES_TASK = SeqrMitoGenotypesMTTask
    VARIANT_GENOTYPES_SCHEMA = SeqrMitoVariantsAndGenotypesSchema
```

hanars (changes requested):

luigi_pipeline/lib/model/mito_mt_schema.py
ShifaSZ: After studying the `BaseMTSchema` class, I found that I could remove unwanted fields from the base class by redefining the annotation function. So, I could use the `SeqrSchema` as a base class without changing it. Do you think it is a good approach?

hanars: yes I think so

luigi_pipeline/seqr_mito_loading.py
ShifaSZ: To my understanding, the purpose of the optimization is to separate the variant annotations and the genotype annotations into two tasks. It could improve performance by more parallel.
I'm not sure the unoptimized way still works or not. But I will try.

hanars: just don't make this task at all. its not used and theres no point

hanars: you misunderstood me. Remove the UN-optimized task - keep ONLY the optimized task

Bust also, do what I said about inheriting classes before that

luigi_pipeline/lib/model/mito_mt_schema.py
```diff
@@ -0,0 +1,183 @@
...
+
+    @row_annotation(name='AC_het')
+    def ac_het(self):
+        return self.mt.AC_het
+
+    @row_annotation(name='AF_het')
+    def af_het(self):
+        return self.mt.AF_het
+
+    # Remove the inherited unwanted annotation function
```
hanars, line 79: this is not a good approach for only inheriting some fields from a base class - its very hard to read and very hard to make sure new fields get added to the appropriate classes. Please go with the approach I originally asked for with a `SeqrCoreSchema`  containing only the overlapping fields that is used both here and by `SeqrVariantSchema`

luigi_pipeline/lib/model/mito_mt_schema.py
```diff
@@ -0,0 +1,183 @@
...
+
+    def hgmd(self):
+        pass
+
+
+class SeqrMitoGenotypesSchema(SeqrGenotypesSchema):
+
+    @row_annotation(fn_require=SeqrGenotypesSchema.genotypes)
+    def samples_ab(self, start=0, end=45, step=5):
+        # struct of x_to_y to a set of samples in range of x and y for ab.
```
hanars, line 139: should say `hl` in the comment instead of `ab`

luigi_pipeline/lib/model/mito_mt_schema.py
```diff
@@ -0,0 +1,183 @@
...
+        return hl.struct(**{
+            '%i_to_%i' % (i, i+step): self._genotype_filter_samples(
+                lambda g: ((g.num_alt == 1) & ((g.hl*100) >= i) & ((g.hl*100) < i+step))
+            )
+            for i in range(start, end, step)
+        })
+
+    def _genotype_fields(self):
+        # Convert the mt genotype entries into num_alt, gq, ab, dp, and sample_id.
```
hanars, line 148: this comment is inaccurate

luigi_pipeline/lib/model/mito_mt_schema.py
```diff
@@ -0,0 +1,183 @@
...
+        }
+
+
+class SeqrMitoVariantsAndGenotypesSchema(SeqrMitoVariantSchema, SeqrMitoGenotypesSchema):
+    """
+    Combined variant and genotypes.
+    """
+
+    @staticmethod
+    def elasticsearch_row(ds):
```
hanars, line 167: is this at all different from the method in `SeqrVariantsAndGenotypesSchema`? If not, since its a static method, you could just do 
```
@staticmethod
def elasticsearch_row(ds):
    return SeqrVariantsAndGenotypesSchema.elasticsearch_row(ds)
```

luigi_pipeline/seqr_mito_loading.py
```diff
@@ -0,0 +1,79 @@
...
+
+from lib.hail_tasks import HailElasticSearchTask
+from lib.model.mito_mt_schema import SeqrMitoVariantsAndGenotypesSchema
+from sv_pipeline.utils.common import get_es_index_name
+import seqr_loading
+
+logger = logging.getLogger(__name__)
+
+
+class SeqrMitoMTTask(seqr_loading.SeqrVCFToMTTask):
```
hanars, line 17: it would be cleaner if you imported `SeqrVCFToMTTask` and used it here, instead of referencing it on the module

hanars (changes requested):
you have still not done what I asked regarding creating a base class for the task - please see this comment and use it as the template for what you should be implementing and then ask me if you have any questions https://github.com/broadinstitute/hail-elasticsearch-pipelines/pull/310#discussion_r814023749 . You seem very unwilling to make any changes to any existing files, which is making the implementation here very clunky and hard to maintain in the long run. If done properly, you will have to make changes to the existing code to make it possible to share what functionality should be shared and keep seperate the functionality that shouldn't

ShifaSZ:
> 

Your suggestion is fine for me. I was just working on that. I pushed the inheritance version of the tasks.

luigi_pipeline/lib/model/mito_mt_schema.py
ShifaSZ: Done

luigi_pipeline/lib/model/mito_mt_schema.py
```diff
@@ -0,0 +1,105 @@
...
+
+class SeqrMitoVariantSchema(BaseSeqrSchema):
+
+    def __init__(self, *args, high_constraint_region, **kwargs):
+        super().__init__(*args, **kwargs)
+        self._high_constraint_region = high_constraint_region
+
+    # Mitochondrial only fields
+    @row_annotation()
+    def gnomad(self):
```
hanars, line 15: depending on how Anne responds on the document, we may want to put this in a new `gnomad_mito` field, or we may want it to be the existing `gnomad_genomes` field, but until she responds we can't know which

luigi_pipeline/lib/model/seqr_mt_schema.py
```diff
@@ -111,12 +97,6 @@ def xstart(self):
     def xstop(self):
         return variant_id.get_expr_for_xpos(self.mt.locus) + hl.len(variant_id.get_expr_for_ref_allele(self.mt)) - 1
 
-    @row_annotation()
-    def rg37_locus(self):
```
hanars, line 115: I think we actually do want this for mito variants as well

luigi_pipeline/seqr_mito_loading.py
```diff
@@ -0,0 +1,60 @@
...
+import hail as hl
+
+from lib.model.mito_mt_schema import SeqrMitoVariantsAndGenotypesSchema, SeqrMitoVariantSchema, SeqrMitoGenotypesSchema
+from luigi_pipeline.seqr_loading_optimized import BaseVCFToGenotypesMTTask, BaseMTToESOptimizedTask
+from seqr_loading import SeqrVCFToMTTask
+
+logger = logging.getLogger(__name__)
+
+
+class SeqrMitoVariantMTTask(SeqrVCFToMTTask):
```
hanars, line 16: I don't think we need this task at all and its kind of clunky to read/ maintain. Lets remove it

hanars (changes requested):

luigi_pipeline/seqr_mito_loading.py
hanars, line 16: In seqr_loading_optimized we currently have the following:
```
class SeqrVCFToVariantMTTask(seqr_loading.SeqrVCFToMTTask):
    """
    Loads all annotations for the variants of a VCF into a Hail Table (parent class of MT is a misnomer).
    """

    def run(self):
        # We only want to use the Variant Schema.
        self.read_vcf_write_mt(schema_cls=SeqrVariantSchema)
```
In that file we should change it to 
```
class SeqrVCFToVariantMTTask(seqr_loading.SeqrVCFToMTTask):
   SCHEMA_CLASS = SeqrVariantSchema

    def run(self):
        self.read_vcf_write_mt(schema_cls=self.SCHEMA_CLASS)
```

and then this class should be
```
class SeqrMitoVariantMTTask(SeqrVCFToVariantMTTask):
   SCHEMA_CLASS = SeqrMitoVariantSchema
```

luigi_pipeline/seqr_loading.py
```diff
@@ -85,11 +85,7 @@ def run(self):
 
         self.read_vcf_write_mt()
 
-    def read_vcf_write_mt(self, schema_cls=SeqrVariantsAndGenotypesSchema):
```
hanars, line 88: rename this function to `read_input_write_mt` so it can be used both here and by the mito class

luigi_pipeline/seqr_loading.py
```diff
@@ -85,11 +85,7 @@ def run(self):
...
-        logger.info("Args:")
-        pprint.pprint(self.__dict__)
-
-        mt = self.import_vcf()
```
hanars, line 92: rename this to `import_vcf` so it can be used both here and by mito

luigi_pipeline/seqr_loading.py
```diff
@@ -106,14 +102,24 @@ def read_vcf_write_mt(self, schema_cls=SeqrVariantsAndGenotypesSchema):
         # hgmd is optional.
         hgmd = hl.read_table(self.hgmd_ht_path) if self.hgmd_ht_path else None
 
-        mt = schema_cls(mt, ref_data=ref_data, clinvar_data=clinvar, hgmd_data=hgmd).annotate_all(
+        mt = schema_cls(mt, ref_data=ref_data, clinvar_data=clinvar, hgmd_data=hgmd, **kwargs).annotate_all(
```
hanars, line 105: instead of passing through kwargs, which can get messy, use a helper method like `get_schema_class_kwargs` which here would be
```
def get_schema_class_kwargs(self):
        ref_data = hl.read_table(self.reference_ht_path)
        clinvar = hl.read_table(self.clinvar_ht_path)
        # hgmd is optional.
        hgmd = hl.read_table(self.hgmd_ht_path) if self.hgmd_ht_path else None
        return {'ref_data': ref_data, 'clinvar': clinvar, 'hgmd_data': hgmd}
```

and for the mito class it would be 
```
def get_schema_class_kwargs(self):
   kwargs = super().get_schema_class_kwargs()
   high_constraint_region = hl.import_locus_intervals(self.high_constraint_interval_path,
                                                           reference_genome='GRCh38')
   kwargs['high_constraint_region'] = high_constraint_region
```

luigi_pipeline/seqr_mito_loading.py
ShifaSZ: I updated the pipeline tasks per your suggestions. Please have a look.

luigi_pipeline/lib/model/seqr_mt_schema.py
hanars: blocks approval

luigi_pipeline/seqr_loading_optimized.py
```diff
@@ -1,36 +1,31 @@
...
 logger = logging.getLogger(__name__)
 
 
-class SeqrVCFToVariantMTTask(seqr_loading.SeqrVCFToMTTask):
+class SeqrVCFToVariantMTTask(SeqrVCFToMTTask):
     """
     Loads all annotations for the variants of a VCF into a Hail Table (parent class of MT is a misnomer).
     """
-
-    def run(self):
```
hanars, line 21: this is still needed, the optimized class absolutely needs a different `run`, it should just now be 
```
def run(self):
    # We only want to use the Variant Schema.
    self.read_input_write_mt()
```

luigi_pipeline/seqr_mito_loading.py
```diff
@@ -0,0 +1,52 @@
...
+
+    def get_schema_class_kwargs(self):
+        kwargs = super().get_schema_class_kwargs()
+        kwargs['high_constraint_region'] = hl.import_locus_intervals(self.high_constraint_interval_path,
+                                                                     reference_genome='GRCh38')
+        return kwargs
+
+    def import_dataset(self):
+        mt = hl.read_matrix_table(self.source_paths[0])
+        return mt.drop('dbsnp_version',
```
hanars, line 30: why is this necessary? The annotation step should ignore unused fields and we then call `select_annotated_mt` which selects only the needed rows for export

luigi_pipeline/seqr_mito_loading.py
```diff
@@ -0,0 +1,52 @@
...
+import hail as hl
+
+from lib.model.mito_mt_schema import SeqrMitoVariantsAndGenotypesSchema, SeqrMitoVariantSchema, SeqrMitoGenotypesSchema
+from luigi_pipeline.seqr_loading_optimized import BaseVCFToGenotypesMTTask, BaseMTToESOptimizedTask
+from seqr_loading import SeqrVCFToMTTask
+
+logger = logging.getLogger(__name__)
+
+
+class SeqrMitoVariantMTTask(SeqrVCFToMTTask):
```
hanars, line 14: this needs to inherit from  seqr_loading_optimized. As I have been saying, we only want mito to use the optimized and not the unoptimized

hanars (changes requested):

luigi_pipeline/lib/model/seqr_mt_schema.py
ShifaSZ: Okay. I didn't add it back because the field `rg37_locus` wasn't annotated with the old code.

luigi_pipeline/seqr_mito_loading.py
ShifaSZ: The `select_annotated_mt` doesn't select the global fields. I can add a selection for them. Then, we don't need to drop extra fields here.

luigi_pipeline/seqr_loading_optimized.py
ShifaSZ: The difference compared with the inherited `run()` is that it doesn't check the input files exist. Why don't we need those checks?

luigi_pipeline/seqr_mito_loading.py
ShifaSZ: Okay. I'm changing it to inherit the `SeqrVCFToVariantMTTask` task.

luigi_pipeline/seqr_mito_loading.py
ShifaSZ: It is hard to select the global fields when calling the `select_annotated_mt` because we don't know which of them we need. Maybe, it is better for us to select them we need here.

hanars: It is better for us to share code with the SNP pipeline, which successfully outputs all the fields it needs and no more. We export globals just fine for SNPs so I don't see why that would be a problem here. Figure out a way so that the existing call of `self.SCHEMA_CLASS(mt, **kwargs).annotate_all(overwrite=True).select_annotated_mt()` handles all of the field selection we need. If we need to make modifications to what we are annotating or what we are selecting, make those changes inside the helper functions that do the annotation or selection, not here. Remeber, partially overriding `annotate_all` or `select_annotated_mt` or adding additional conditional logic in those functions is allowed

hanars: > It is hard to select the global fields when calling the select_annotated_mt because we don't know which of them we need. 

We absolutely do know what fields we need in the outputted index, you wrote a detailed document listing all of them. Moreover, for SNPs (and now this pipeline since you are using shared code) the call the `annotate_globals` is run after the call to `select_annotated_mt` so the necessary globals are explicitly set rather than accidentally being carried over from the input. This is the behavior we want for mito. If there are any additional globals you think we need for mito, we should add a helper function that generates the globals for an index and override it for mito

ShifaSZ: There are no additional global fields for mito. So, it is weird to override `select_annotated_mt` for mito. I also found that the `_meta` fields in the existing Elasticsearch indices contain only those fields explicitly annotated in the pipeline. The `select_globals` method can do annotation and selection of the global fields at the same time.  How about we changing `annotate_globals` to `select_globals`?

hanars: I don't understand why anything needs to change at all

luigi_pipeline/seqr_loading_optimized.py
hanars: please do not change the behavior of the optimized task we use for all or data loading, we made it different on purpose

luigi_pipeline/seqr_mito_loading.py
hanars: sorry github was showing this comment really weirdly that mad it hard to see what you had actually changed, this looks fine now

luigi_pipeline/lib/model/mito_mt_schema.py
hanars: approval blocked on confirmation from Anne how we want to do this, will figure that out by tomorrow

luigi_pipeline/lib/model/mito_mt_schema.py
```diff
@@ -0,0 +1,105 @@
...
+    @row_annotation(name='AF_het')
+    def af_het(self):
+        return self.mt.AF_het
+
+
+class SeqrMitoGenotypesSchema(SeqrGenotypesSchema):
+
+    @row_annotation(fn_require=SeqrGenotypesSchema.genotypes)
+    def samples_ab(self, start=0, end=45, step=5):
+        # struct of x_to_y to a set of samples in range of x and y for hl.
```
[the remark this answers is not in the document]
hanars: approval blocked on confirmation from Anne how we want to do this, will figure that out by tomorrow (may need to be `samples_hl` if we do want separate sliders)

luigi_pipeline/seqr_loading_optimized.py
hanars: maintaining this functionality blocks approval

luigi_pipeline/seqr_mito_loading.py
```diff
@@ -0,0 +1,45 @@
...
+
+    def get_schema_class_kwargs(self):
+        kwargs = super().get_schema_class_kwargs()
+        kwargs['high_constraint_region'] = hl.import_locus_intervals(self.high_constraint_interval_path,
+                                                                     reference_genome='GRCh38')
+        return kwargs
+
+    def import_dataset(self):
+        mt = hl.read_matrix_table(self.source_paths[0])
+        return mt.select_globals('vep_version')
```
hanars, line 29: what does this do?

ShifaSZ: I thought we might need the `vep_version` global field. I will remove it after changing `annotate_globals` to `select_globals`.

luigi_pipeline/seqr_loading.py
```diff
@@ -118,7 +118,7 @@ def read_input_write_mt(self):
         kwargs = self.get_schema_class_kwargs()
         mt = self.SCHEMA_CLASS(mt, **kwargs).annotate_all(overwrite=True).select_annotated_mt()
 
-        mt = mt.annotate_globals(sourceFilePath=','.join(self.source_paths),
+        mt = mt.select_globals(sourceFilePath=','.join(self.source_paths),
```
hanars, line 121: why is this necessary? The point is to share code where possible, NOT to change the current SNP pipeline behavior. Unless there is an incredibly compelling reason to make this change, the existing pipeline needs to stay as is. If there is a very compelling reason that the mito pipeline needs this to be a different function, then you need to make a helper function and override it for mito. But I really do think we want them to share code as much as possible, so if the existing functionality works as is you shouldn't change it

hanars (changes requested):

hanars:
A functionality change we do need to make: we need to add `datasetType` to the globals. Currently for SVs we set it and then in seqr we assume if its missing it has type "VARIANTS". We should now explicitly set datasetType to be "VARIANTS" for the main pipeline and "MITO" for the mito pipeline. We can do this by adding DATASET_TYPE as a class variable and overriding it as needed

luigi_pipeline/seqr_loading.py
hanars: the SV pipeline is also using annotate_globals so I'm even more confused by this change

ShifaSZ: The change is to get rid of the extra global fields in the mitochondrial Matrix Table. The 'select_globals` does the same annotations as `annotate_globals` does. And it also does the global field selections. 
Let's see how a helper function works.

ShifaSZ:
> 

Okay. I will specify it in the document.

luigi_pipeline/lib/model/mito_mt_schema.py
```diff
@@ -0,0 +1,113 @@
...
+        return self.mt.mitotip_trna_prediction
+
+    @row_annotation()
+    def high_constraint_region(self):
+        return hl.is_defined(self._high_constraint_region[self.mt.locus])
+
+    # Fields with the same names but annotated differently
+    @row_annotation()
+    def rsid(self):
+        return self.mt.rsid.find(lambda x: hl.is_defined(x))
```
ShifaSZ, line 53: The `rsid` field has to be annotated differently because the `rsid` field is an array of strings in the mitochondrial MT while we just need a string.

hanars: what are some examples of the rsids in the VCF? Do any of the rows have multiple rsids, or is it always a list of 1 element?

luigi_pipeline/lib/model/mito_mt_schema.py
```diff
@@ -0,0 +1,113 @@
...
+    @row_annotation(name='AF')
+    def af(self):
+        return self.mt.AF_hom
+
+    @row_annotation(name='AN')
+    def an(self):
+        return self.mt.AN
+
+    @row_annotation(name='AC_het')
+    def ac_het(self):
```
hanars, line 68: these fields are mito specific, not "Fields with the same names but annotated differently" so you should move them above that comment

luigi_pipeline/lib/model/mito_mt_schema.py
```diff
@@ -0,0 +1,113 @@
...
+        # struct of x_to_y to a set of samples in range of x and y for heteroplasmy level.
+        return hl.struct(**{
+            '%i_to_%i' % (i, i+step): self._genotype_filter_samples(
+                lambda g: ((g.num_alt == 1) & ((g.hl*100) >= i) & ((g.hl*100) < i+step))
+            )
+            for i in range(start, end, step)
+        })
+
+    # Override the samples_ab annotation
+    def samples_ab(self, start=0, end=45, step=5):
```
hanars, line 89: probably better to do `def samples_ab(self, **kwargs)` so we don't need to keep the kwargs in sync

hanars (changes requested):
you need to update the base of this branch now that the mito-ref PR was merged

luigi_pipeline/lib/model/mito_mt_schema.py
ShifaSZ: Here are some examples of the rsid field:
image

luigi_pipeline/lib/model/mito_mt_schema.py
ShifaSZ: I thought that Python just overrode the method with the same arguments. It is not true. I'll remove all of them.

hanars:
This overall looks good to me! I would like to hold off on merging it until @mike-w-wilson gets back from his leave and has time to take a look, just because it does refactor main variant loading a bit and I want to make sure he's okay with these changes. But overall I really like it

mike-w-wilson (commented):
A couple small things and I would like to test the SNV pipeline changes before we merge to make sure I did not miss anything. I should have time to do that on Thursday

luigi_pipeline/seqr_loading_optimized.py
```diff
@@ -40,33 +39,45 @@ def run(self):
...
         super().__init__(*args, **kwargs)
 
     def requires(self):
-        return [SeqrVCFToVariantMTTask(), SeqrVCFToGenotypesMTTask()]
+        return [self.VariantTask(), self.GenotypesTask()]
 
     def run(self):
         variants_mt = hl.read_matrix_table(self.input()[0].path)
         genotypes_mt = hl.read_matrix_table(self.input()[1].path)
+        genotypes_mt = genotypes_mt.drop(*[k for k in genotypes_mt.globals.keys()])
```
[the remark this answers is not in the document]
ShifaSZ: Good catch! The following statement `row_ht = genotypes_mt.rows().join(variants_mt.rows())` joins two tables with the same global fields. Hail renames the global fields from one table to avoid conflicts and puts them all into the new table. Hail has the below messages:
```
2022-07-12 10:04:57 Hail: INFO: Table.join: renamed the following fields on the right to avoid name conflicts:
    'sourceFilePath' -> 'sourceFilePath_1'
    'hail_version' -> 'hail_version_1'
    'datasetType' -> 'datasetType_1'
    'sampleType' -> 'sampleType_1'
    'genomeVersion' -> 'genomeVersion_1'
```
Then, there are duplications in the global fields. All these fields go to the `_meta` field of the ES index while exporting the ES. It seems to me that we never use the fields "*_1" from the ES index. But I'm not sure whether they are useful or not.

hanars:
The mito-ref branch has laready been merged, this needs to be PRd against master now

luigi_pipeline/seqr_loading_optimized.py
```diff
@@ -42,37 +41,49 @@ def run(self):
...
         row_ht = genotypes_mt.rows().join(variants_mt.rows())
 
-        row_ht = SeqrVariantsAndGenotypesSchema.elasticsearch_row(row_ht)
+        row_ht = self.VariantsAndGenotypesSchema.elasticsearch_row(row_ht)
         es_shards = self._mt_num_shards(genotypes_mt)
 
         # Initialize an empty SeqrVariantsAndGenotypesSchema to access class properties
-        disabled_fields = SeqrVariantsAndGenotypesSchema(None, ref_data=defaultdict(dict), clinvar_data=None).get_disable_index_field()
-        
+        disabled_fields = self.VariantsAndGenotypesSchema(None, ref_data=defaultdict(dict), clinvar_data=None).get_disable_index_field()
```
ShifaSZ, line 74: This is a change to the new code for adding the `disable_index` option.

hanars: I don't see how this is a change? This seems like just the change you make everywhere in this class to use the correct base class

luigi_pipeline/lib/model/mito_mt_schema.py
```diff
@@ -0,0 +1,113 @@
...
+        super().__init__(*args, **kwargs)
+        self._high_constraint_region = high_constraint_region
+
+    # Mitochondrial only fields
+    @row_annotation()
+    def gnomad_mito(self):
+        return self._selected_ref_data.gnomad_mito
+
+    @row_annotation(disable_index=True)
+    def mitomap(self):
```
hanars, line 19: So we do not currently allow you to filter on the mito-specific pathogenicity and predictor scores, but in the future we might and if we disable indexing for these fields we would need to reload all the data before we could do that. I think that, although we have no short-term plans to add filtering on those fields, the overhead of reloading data is so much worse than the overhead of a slightly larger index, especially given that the mito index is relatively small, that it makes sense for us to enable indexing for fields we think we may someday want. Therefore, we should enable indexing for all the mito-specific fields. The core seqr fields with indexing disabled should still be disabled here

luigi_pipeline/seqr_loading_optimized.py
ShifaSZ: `SeqrVariantsAndGenotypesSchema` is changed to `self.VariantsAndGenotypesSchema`. The git merged the code and reported no conflicts. I made the change manually.

mike-w-wilson:
I've completed testing on the class rewrite. I compared the variant MT and project MT to previous versions and they are identical so I am fine with these changes. From my perspective, this looks good but I think @hanars should give the final stamp of approval since there have been updates since June 23rd when she requested my review.

hanars:
@mike-w-wilson if its good with you it is good with me

mike-w-wilson (approved):

PR #311, opened 2022-02-28
Update contig recoding param for importing VCF to include Clinvar Mitochondrial variants

ShifaSZ:
Clinvar uses a slightly different contig name for the mitochondrial variants. It uses 'MT' instead of the 'M' in the GRCh38. That's why the mitochondrial variant data are missing in the downloaded Clinvar hail table. This PR adds an entry for `"MT": "chrM"` in the `contig_recoding` to include mitochondrial data.

### DIFF
--- a/hail_scripts/utils/hail_utils.py
+++ b/hail_scripts/utils/hail_utils.py
@@ -60,6 +60,10 @@ def import_vcf(
     contig_recoding = {
         **{ref_contig.replace("chr", ""): ref_contig for ref_contig in ref.contigs if "chr" in ref_contig},
         **{f"chr{ref_contig}": ref_contig for ref_contig in ref.contigs if "chr" not in ref_contig}}
+    if genome_version == 38:
+        contig_recoding['MT'] = 'chrM'
+    else:
+        contig_recoding['chrM'] = 'MT'
 
     mt = hl.import_vcf(
         vcf_path,


hail_scripts/utils/hail_utils.py
```diff
@@ -60,6 +60,10 @@ def import_vcf(
     contig_recoding = {
         **{ref_contig.replace("chr", ""): ref_contig for ref_contig in ref.contigs if "chr" in ref_contig},
         **{f"chr{ref_contig}": ref_contig for ref_contig in ref.contigs if "chr" not in ref_contig}}
+    if genome_version == '38':
```
hanars, line 63: this seems like a weird way to do the recoding - do we have any reason to believe that all build 38 VCFs that we would want to import would always have a different notation than build 37 (remember this is a generic utility function used all over the place)? Why would we want different contig representation for different builds? I will defer to @mike-w-wilson but I think this is way too fragile an approach, we should either always provide the same contig recoding for mito variants, or should allow this function to take an optional parameter with overriding contigs so we can do it on a per vcf basis and only make this change for clinvar

hanars (changes requested):

hail_scripts/utils/hail_utils.py
hanars: right, but his code is not dependent on genome build version??

ShifaSZ: I learned that gnomAD has changed to import Clinvar data in XML format instead of VCF format. So they don't need to recode the contig while importing. Clinvar uses 'MT' for mitochondrial contig for both 37 and 38. Therefore, gnomAD can use a single `normalized_contig` function to convert 'MT' to 'M' for both versions.

ShifaSZ: It is safer to add an optional parameter for the mito contig recoding. Because we don't know what are the cases for the other datasets.

mike-w-wilson: This new approach is much better. We never know how new databases code their contigs so this gives us the flexibility to address differences as we add more reference datasets.

hanars (approved):
looks good to me, if @mike-w-wilson approves

mike-w-wilson (approved):
LGTM

PR #321, opened 2022-03-14
Remove BOTHSIDES_SUPPORT from filters field and add a bothsides_support_filter field

ShifaSZ:
Changes for https://github.com/broadinstitute/seqr/issues/2463


sv_pipeline/genome/load_data.py
```diff
@@ -38,7 +38,8 @@
                                              hl.missing(hl.tstr)),
     'gnomad_svs_AF': lambda rows: rows.info.gnomAD_V2_AF,
     'pos': lambda rows: rows.locus.position,
-    'filters': lambda rows: hl.array(rows.filters.filter(lambda x: x != 'PASS')),
+    'filters': lambda rows: hl.array(rows.filters.filter(lambda x: (x != 'PASS') & (x != 'BOTHSIDES_SUPPORT'))),
+    'bothsides_support_filter': lambda rows: rows.filters.any(lambda x: x == 'BOTHSIDES_SUPPORT'),
```
hanars, line 42: so its not actually a filter (which is part of why its so confusing that it was in the filters field to begin with) so just name it `bothsides_support`

sv_pipeline/genome/load_data.py
```diff
@@ -38,7 +38,8 @@
                                              hl.missing(hl.tstr)),
     'gnomad_svs_AF': lambda rows: rows.info.gnomAD_V2_AF,
     'pos': lambda rows: rows.locus.position,
-    'filters': lambda rows: hl.array(rows.filters.filter(lambda x: x != 'PASS')),
+    'filters': lambda rows: hl.array(rows.filters.filter(lambda x: (x != 'PASS') & (x != 'BOTHSIDES_SUPPORT'))),
```
hanars, line 41: since you use the string `'BOTHSIDES_SUPPORT'` twice it is better to make a constant

hanars (changes requested):

hanars (approved):

PR #324, opened 2022-03-18
Stop exporting empty fields to ES.

ShifaSZ:
This change may require some seqr code changes. One of the examples is as the following.
```
hit[SORTED_TRANSCRIPTS_FIELD_KEY] or []
```
Needs to be changed to:
```
hit.get(SORTED_TRANSCRIPTS_FIELD_KEY, [])
```


hanars:
We want to maintain current behavior. Please in the future do not make pipeline changes until I actually make tickets and assign them. This is out of scope of current work and lower priority than everything on your plate, with a very high risk of breaking seqr

PR #326, opened 2022-03-22
Add the 'rg37_locus' and 'rg37_locus_end' fields.

ShifaSZ:
Add 'rg37_locus_end' for the translocation variants only because some CPX variants have END2 which exceeds the range of the GRCh38 reference genome.


hanars (approved):
My understanding of the `rg37_locus_end` logic is as follows: for translocations with `END2` defined it does the liftover for the `CHR2` and `END2` and for everything else it does the liftover for `contig` and `END`. If that is correct, then this is good. If I am not understanding that correctly though please let me know whats actually happening here

PR #334, opened 2022-03-24
Change the `rg37_locus_end` field to work for all SV types.

ShifaSZ:


sv_pipeline/genome/load_data.py
```diff
@@ -175,6 +178,10 @@ def annotate_fields(mt, gencode_release, gencode_path):
                                                                  hl.missing(hl.dtype('array<str>')))})
     rows = rows.annotate(**{k: v(rows) for k, v in DERIVED_FIELDS.items()})
 
+    rows = rows.annotate(rg37_locus_end=hl.liftover(
```
hanars, line 181: I don't like adding an annotation in `DERIVED_FIELDS` thats just completely wrong so you can override it here. `exp_contig_lengths` is something you could actually derive inline instead of passing it around like a variable and it would be much neater. Also, I told you if it is out of range we should just skip liftover. Also, I think we want to share some logic between `xstop` and `rg37_locus_end` to prevent this sort of issue in the future. Here is my proposal: 
```
# In CORE_FIELDS
'end_locus': lambda rows:  hl.if_else(hl.is_defined(rows.info.END2),
                   hl.locus(rows.info.CHR2, rows.info.END2, reference_genome='GRCh38'),
                   hl.locus(rows.locus.contig, rows.info.END, reference_genome='GRCh38'))
...
# In  DERIVED_FIELDS
'xstop': lambda rows: get_xpos(rows.end_locus.contig, rows.end_locus.position),
'rg37_locus_end': lambda rows: hl.if_else(
    rows.end_locus.position < hl.literal(hl.get_reference('GRCh38').lengths)[rows.end_locus.contig], 
    hl.liftover(rows.end_locus, 'GRCh37'), 
    hl.missing(hl.dtype('locus'))),
...
# update export FIELDS list to not export the intermediate value
FIELDS.remove('end_locus')
```

hanars (changes requested):

sv_pipeline/genome/load_data.py
ShifaSZ: Adding `end_locus` is fine. I'm not sure the cost for `hl.literal(hl.get_reference('GRCh38').lengths)`. Hope the `hl.literal()` and `hl.get_reference()` won't be executed repeatedly for each row. Using a missing value for the out of range is fine. I didn't know that was an option. The unlifted locus is not possible when the position is out of range.

hanars: hail should be pretty optimized not to be re-computing literals in annotation statements, I wouldn't worry about that. I'm not sure what you mean by "The unlifted locus is not possible when the position is out of range."

ShifaSZ: Another problem is that when END2 is out of range, it is not possible to create a locus for it. So the `end_locus` can fail for some END2. I'll use a hl.struct() to replace the hl.locus().

ShifaSZ: Performance of the code of `hl.literal(hl.get_reference('GRCh38').lengths)` looks good. It's almost the same as before.

hanars (approved):
looks good!

PR #338, opened 2022-04-05
Adding init.py to sv_pipeline

karynne7:
I needed to have these files to make my load_data.py script run.


hanars (approved):
Thanks for adding, this is very helpful

PR #349, opened 2022-07-28
Add ignore-missing-samples option.

ShifaSZ:
Add an ignore-missing-samples option to allow testing on partial mitochondrial callset.

### DIFF
--- a/luigi_pipeline/lib/hail_tasks.py
+++ b/luigi_pipeline/lib/hail_tasks.py
@@ -48,6 +48,7 @@ class HailMatrixTableTask(luigi.Task):
     genome_version = luigi.Parameter(description='Reference Genome Version (37 or 38)')
     vep_runner = luigi.ChoiceParameter(choices=['VEP', 'DUMMY'], default='VEP', description='Choice of which vep runner'
                                                                                             'to annotate vep.')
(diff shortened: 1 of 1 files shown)


luigi_pipeline/lib/hail_tasks.py
```diff
@@ -123,8 +124,7 @@ def run_vep(mt, genome_version, runner='VEP', vep_config_json_path=None):
 
         return runners[runner]().run(mt, genome_version, vep_config_json_path=vep_config_json_path)
 
-    @staticmethod
-    def subset_samples_and_variants(mt, subset_path):
+    def subset_samples_and_variants(self, mt, subset_path):
```
ShifaSZ, line 127: Change the method type to access the `ignore_missing_samples` parameter.

hanars:
We only want this to be an option for the mitpo loading so we should not be changing the base class at all

ShifaSZ:
> We only want this to be an option for the mito loading so we should not be changing the base class at all

I could change the new Luigi parameter `ignore-missing-sample` to the mito class. But I have to change the base class method to use the parameter. Or should I overwrite the method in the mito class?

hanars:
override the method in the mito class

PR #355, opened 2022-08-19
Add noncoding region based reference datasets

mike-w-wilson:
This PR adds in two noncoding region reference datasets: Siwei's noncoding constraint z score and SCREEN. It also adds a general function for adding any region based reference dataset should we get one in the future. The SCREEN dataset required a little bit of manipulation so I wrote a script for writing that HT. Siwei's is a direct hl.import_bed -> write ht situation so I did not write a script for that but can add it if you think it is necessary.


hanars (commented):
So as a high level thing, it looks like variants that otherwise exist in our reference data will be annotated with the regions, but during loading any new variant that we add thats not otherwise in the reference data will not be annotated. Is that correct?

mike-w-wilson:
You're right, I thought CADD got us around this but thinking more, that is not the case...I'm going to move this a schema class annotation

download_and_create_reference_datasets/v02/hail_scripts/write_combined_reference_data_ht.py
```diff
@@ -217,6 +229,28 @@
...
+    :param reference_genome: Reference genome, e.g. GRCh38
+    :return: hail Table annotated with references target annotation.
+    """
+    region_ht = hl.read_table(CONFIG[region_dataset][reference_genome]["path"])
+    region_select = {
+        region_dataset: hl.struct(
+            **{
+                CONFIG[region_dataset][reference_genome]["select"]["target"]: region_ht[
+                    ht.locus
+                ].target
```
ShifaSZ, line 246: I'm curious why we split these into multiple lines.

download_and_create_reference_datasets/v02/hail_scripts/write_combined_reference_data_ht.py
```diff
@@ -217,6 +229,28 @@
 CONFIG['dbnsfp_mito']['38']['filter'] = lambda ht: ht.locus.contig == 'chrM'
 
 
+def annotate_ref_regions(ht, region_dataset, reference_genome):
+    """
+    Annotate combined reference data with region based reference data.
+    :param ht: Combined reference hail Table.
+    :param region_dataset: Imported region/bed file with 3 or 4 columns
+    :param reference_genome: Reference genome, e.g. GRCh38
```
ShifaSZ, line 237: The `e.g. GRCh38` is a little confusing. It could be more clear to change it to `e.g. '38' for GRCh38`

ShifaSZ (commented):
Some nitpicking comments.

mike-w-wilson:
@hanars , this has been tested using the new DAG code in https://github.com/broadinstitute/seqr-pipeline-airflow/pull/88 which produced the ES index r000_test_case__wes__grch38__variants__v02__vcfv16__20220919. The two new ES fields are `gnomad_non_coding_constraint_z_score`, a float,  and `screen_region_type`, an array of strings. The test case is a small random region of the exome and is quite small.

luigi_pipeline/seqr_loading.py
```diff
@@ -205,6 +208,7 @@ class SeqrMTToESTask(HailElasticSearchTask):
     vep_runner = luigi.ChoiceParameter(choices=['VEP', 'DUMMY'], default='VEP', description='Choice of which vep runner to annotate vep.')
 
     reference_ht_path = luigi.Parameter(default=None, description='Path to the Hail table storing the reference variants.')
+    interval_ref_ht_path = luigi.Parameter(description='Path to the Hail Table storing interval-keyed reference data.')
```
hanars, line 211: all the other params have `default=None`, you probably want to add that here too

luigi_pipeline/lib/model/seqr_mt_schema.py
```diff
@@ -215,6 +216,31 @@ def hgmd(self):
...
+                    self.mt.locus, all_matches=True
+                )
+                .filter(
+                    lambda x: hl.is_defined(x.gnomad_non_coding_constraint["z_score"])
+                )
+                .gnomad_non_coding_constraint.z_score
+            }
+        )
+
+    @row_annotation(disable_index=True)
```
hanars, line 234: in the seqr ticket for this it says we want the ability to filter on these fields, so I think we should not disable indexing

hanars (changes requested):
The overall approach looks good although I have a few small comments. Do we have build 37 data to add, or will this only be available in 38?

download_and_create_reference_datasets/v02/hail_scripts/write_combined_reference_data_ht.py
mike-w-wilson: Ah in the gnomAD repos we format using the tool 'black' which only allows lines to be a certain length. I think I formatted this file the same way. I can make this a single line but I do plan on formatting the entire repo using black at some point as part of a standardization process.

mike-w-wilson:
@hanars, we only have 38 right now. We could lift over the regions to generate the resource for 37 but I think that should be another ticket which makes me think we need to change the behavior here to act more like hgmd?

luigi_pipeline/lib/model/seqr_mt_schema.py
mike-w-wilson: Thanks for the reminder, Anne confirmed she wants to search on zscore as well so I enabled indexing there as well.

hanars (approved):

PR #357, opened 2022-09-01
Add an allowing missing samples option for mitochondria dataset

ShifaSZ:


luigi_pipeline/configs/seqr-mito-loading-local.cfg
```diff
@@ -4,19 +4,33 @@
 sample_type = WGS
 genome_version = 38
 dont_validate=True
-source_paths = ["test_ten_cmg_mito_combined.mt"]
+source_paths = ["../test_ten_cmg_mito_combined.mt"]
+; source_paths = ["gs://seqr-datasets/cmg_mito/test_ten_cmg_mito_combined.mt"]
```
hanars, line 8: There is a ton of commented out code in this file. We should either use those values instead, add helpful comments explaining why we left them in, or take them out

luigi_pipeline/seqr_mito_loading.py
```diff
@@ -35,9 +36,37 @@ def annotate_globals(self, mt):
...
+        anti_join_ht = subset_ht.anti_join(mt.cols())
+        anti_join_ht_count = anti_join_ht.count()
+
+        if anti_join_ht_count != 0:
+            missing_samples = anti_join_ht.s.collect()
+            message = f'Only {subset_count - anti_join_ht_count} out of {subset_count} ' \
+                      f'subsetting-table IDs matched IDs in the variant callset.\n' \
+                      f'IDs that aren\'t in the callset: {missing_samples}\n' \
+                      f'All callset sample IDs:{mt.s.collect()}'
+            if (subset_count > anti_join_ht_count) and self.ignore_missing_samples:
```
hanars, line 58: this is almost entirely copy-pasted from the base function, this line is basically the only part that changed. When I said you needed to override it, I meant you should override the very small part that is different, not just copy-paste it. Shared behavior needs to actually be shared, when you copy it out it makes future maitenance basically impossible. You can achieve this by adding a keyword argument to the base class and then use super to set that keyword properly:
```
def subset_samples_and_variants(self, *args):
    return super().subset_samples_and_variants(*args, ignore_missing_samples=self.ignore_missing_samples)
```

hanars (changes requested):

luigi_pipeline/seqr_mito_loading.py
hanars: yes and we change the base class method to take a keyword arg where the default behavior is the current behavior

hanars (commented):
looks good to me, would want @mike-w-wilson to approve

luigi_pipeline/configs/seqr-mito-loading-local.cfg
```diff
@@ -4,19 +4,22 @@
...
 high_constraint_interval_path = gs://seqr-reference-data/GRCh38/mitochondrial/Helix high constraint intervals Feb-15-2022.tsv
 
 [SeqrMitoGenotypesMTTask]
 sample_type=WGS
 genome_version=38
-source_paths = ["test_ten_cmg_mito_combined.mt"]
-remap_path = gs://seqr-datasets/v02/GRCh38/RDG_WGS_Broad_Internal/base/projects/R0486_cmg_gcnv/R0486_cmg_gcnv_remap.tsv
-subset_path = gs://seqr-datasets/v02/GRCh38/RDG_WGS_Broad_Internal/base/projects/R0486_cmg_gcnv/R0486_cmg_gcnv_ids.txt
+source_paths = ["gs://seqr-datasets/cmg_mito/test_ten_cmg_mito_combined.mt"]
 dest_path = genotypes.mt
```
mike-w-wilson, line 17: Looks like everything got moved to buckets except the output, is that the desired behavior?

mike-w-wilson (approved):
One small question in the test config but otherwise LGTM!

luigi_pipeline/configs/seqr-mito-loading-local.cfg
ShifaSZ: It is an intermediate output. So it is acceptable to be local.

PR #366, opened 2022-10-07
add fasta file for local install hg37

hanars:
The updated docker image has already been build and pushed, addresses request from https://github.com/broadinstitute/seqr/discussions/2695#discussioncomment-3790576


mike-w-wilson (approved):
LGTM

PR #367, opened 2022-10-12
Grab first element of zscore list

mike-w-wilson:
The way we have to grab all annotations when joining locus and allele keyed tables to interval key tables generates a list of structs containing the annotations. The original implementation maintained this list for gnomAD noncoding constraint zscore however the list contained a single element so this change just grabs that element or returns missing for empty lists.


mike-w-wilson:
r000_test_case__wes__grch38__variants__v02__vcfv17__20221012, I confirmed its a double in kibana but please double check!

PR #369, opened 2022-10-17
Update code to make int ref optional

mike-w-wilson:
We do not have interval reference datasets for GRCh37 right now so this updates the code to make int ref optional. I also updated the rg37_locus indexing as the pipeline currently does not work for GRCh37 projects. I figured indexing this field rather than building a bunch of logic around which fields to index based on build was a simpler and cheap solution since that is a small field. These changes are needed to allow GRCh37 projects to load.

### DIFF
--- a/luigi_pipeline/lib/model/seqr_mt_schema.py
+++ b/luigi_pipeline/lib/model/seqr_mt_schema.py
@@ -98,7 +98,7 @@ class BaseSeqrSchema(BaseMTSchema):
     def xstop(self):
         return variant_id.get_expr_for_xpos(self.mt.locus) + hl.len(variant_id.get_expr_for_ref_allele(self.mt)) - 1
 
-    @row_annotation(disable_index=True)
+    @row_annotation()
     def rg37_locus(self):
         if self.mt.locus.dtype.reference_genome.name != "GRCh38":
             raise RowAnnotationOmit
@@ -219,6 +219,8 @@ class SeqrSchema(BaseSeqrSchema):
 
     @row_annotation()
     def gnomad_non_coding_constraint(self):
+        if self._interval_ref_data is None:
+            raise RowAnnotationOmit
         return hl.struct(
             **{
                 "z_score": self._interval_ref_data.index(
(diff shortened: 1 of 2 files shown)


luigi_pipeline/lib/model/seqr_mt_schema.py
```diff
@@ -98,7 +98,7 @@ def xstart(self):
     def xstop(self):
         return variant_id.get_expr_for_xpos(self.mt.locus) + hl.len(variant_id.get_expr_for_ref_allele(self.mt)) - 1
 
-    @row_annotation(disable_index=True)
+    @row_annotation()
```
hanars, line 101: I don't see why we need to remove `disable_index` here though? It doesn't look like you were able to clean up any build-specific index field logic so I feel like it doesn't hurt to leave this in and maintain current behavior?

hanars (changes requested):

luigi_pipeline/lib/model/seqr_mt_schema.py
mike-w-wilson: Sorry -- I did not explain this bit well. When we create the disable index fields list https://github.com/broadinstitute/seqr-loading-pipelines/blob/0602427fe3daf8078189a373ba44fa10edd55596/luigi_pipeline/seqr_loading_optimized.py#L74 `rg37_locus` is part of the list but that is not an annotation that our GRCh37 MTs have so the pipeline throws an error saying it cannot find the annotation to disable it. Re-enabling the index here prevents it from being part of the list and was just a cheap way to get around the error without adding any reference build logic.

hanars (approved):

luigi_pipeline/lib/model/seqr_mt_schema.py
mike-w-wilson: I could just add something like if build=="GRCh37": disabled_fields.remove('rg37_locus') ... I dont know off the top of my head where to pull build from but thats an alternative if we want to keep this disabled

hanars: nah this way is easier and like you said its not very big data

PR #373, opened 2022-11-10
Use hail's agg.call_stats to generate AC,AN,AF, hom alt counts

mike-w-wilson:
This update replaces DSP's AC,AN,AF within the INFO struct with hails callstats. This resolves a bug found by Stephanie where low qual variants were no longer being counted towards DSPs AC which is not the behavior we want. This works for all callsets, not just the WGS and is a change that will allow us to no longer require AC,AN, or AF from AnVIL users. Hail's methods generates AC,AF, and hom alt for both ref and alt so we grab the alt from this fields with [1]. We also get homozygote count for free here which was a feature request. I'm not sure if the callstats function should have a @ staticmethod decorator though? I tested this on a shard of the recent callset and the ES index is r0652_pipeline_test__wgs__grch38__variants__v02__vcfv32__20221110

### DIFF
--- a/luigi_pipeline/lib/hail_tasks.py
+++ b/luigi_pipeline/lib/hail_tasks.py
@@ -87,7 +87,7 @@ class HailMatrixTableTask(luigi.Task):
                              reference_genome='GRCh' + self.genome_version,
                              skip_invalid_loci=True,
                              contig_recoding=recode,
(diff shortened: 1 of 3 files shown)


hanars (approved):
looks good!

PR #374, opened 2022-11-16
fix docker build platform and script path

hanars:
Fixes the issues discussed in https://github.com/broadinstitute/seqr/discussions/3069 and https://github.com/broadinstitute/seqr/discussions/3031

I have already built and released the image


mike-w-wilson (approved):
LGTM

PR #375, opened 2022-11-23
Add merging-sample-call support for gCNV loading pipeline.

ShifaSZ:
See this [issue](https://github.com/broadinstitute/seqr/issues/3065).


sv_pipeline/exome/load_data.py
```diff
@@ -106,16 +107,25 @@ def _parse_genes(genes):
...
+}
+
+NEW_JOINT_CALL_SAMPLE_COLS = {
     PREV_IDENTICAL_COL: {'field_name': 'prev_call', 'format': bool},
     PREV_OVERLAP_COL: {'field_name': 'prev_overlap', 'format': bool},
     PREV_MISSING_COL: {'field_name': NEW_CALL_FIELD, 'format': lambda call: call not in {'NA', 'FALSE'}},
 }
+
+MERGING_CALL_SAMPLE_COLS = {
+    IS_LATEST: {'field_name': 'prev_call', 'format': lambda val: not BOOL_MAP[val.strip()]},
```
hanars, line 119: this is incorrect, this should be `PREV_IDENTICAL_COL`. The column names are the same regardless of how we are doing the loading, its just how they are computed that is different

sv_pipeline/exome/load_data.py
```diff
@@ -106,16 +107,25 @@ def _parse_genes(genes):
...
+    PREV_OVERLAP_COL: {'field_name': 'prev_overlap', 'format': lambda val: False},
+    PREV_MISSING_COL: {'field_name': NEW_CALL_FIELD, 'format': lambda val: False},
+}
+
 COL_CONFIGS.update({col: {'format': _parse_genes} for col in GENE_CONSEQUENCE_COLS.keys()})
 
 CORE_COLUMNS = [CHR_COL, SC_COL, SF_COL, CALL_COL, IN_SILICO_COL]
 SAMPLE_COLUMNS = [
```
hanars, line 127: the columns are the same regardless of how we do loading, so this should stay the same

sv_pipeline/exome/load_data.py
```diff
@@ -439,6 +449,7 @@ def main():
     p.add_argument('--skip-sample-subset', action='store_true')
     p.add_argument('--write-subsetted-bed', action='store_true')
     p.add_argument('--ignore-missing-samples', action='store_true')
+    p.add_argument('--is-merging-sample-call', action='store_true')
```
hanars, line 452: per my comment on the ticket, we need to have the default behavior change and we want the `--is-new-joint-call` flag to be something you have to set explicitly

hanars (changes requested):

sv_pipeline/exome/load_data.py
ShifaSZ: The `PREV_IDENTICAL_COL` is the column in the input file. The output column `prev_call` doesn't change.

sv_pipeline/exome/load_data.py
ShifaSZ: The columns do not change. The formatting functions are changed. I could update the formatting functions if there is no option of the `is_new_join_call`.

ShifaSZ:
The column configurations are still to be confirmed. But it is ready for a new round of review.

sv_pipeline/exome/load_data.py
hanars: I see. In that case you need to make constants for `prev_call` and `prev_overlap`, you shouldn't be copy-pasting strings

sv_pipeline/exome/load_data.py
```diff
@@ -450,6 +469,12 @@ def main():
     if not es_password:
         es_password = getpass(prompt='Enter ES password: ')
 
+    sample_col_configs = NEW_JOINT_CALL_SAMPLE_COLS if args.is_new_joint_call else MERGING_CALL_SAMPLE_COLS
+    COL_CONFIGS.update(sample_col_configs)
```
hanars, line 473: you can't update constants conditionally, its really unstable because you assume a constant is whatever it is defined as but in fact the constant could be one of many things depending on what calls were made to this function. You need to define 2 different configs at the top level, `COL_CONFIGS` and `JOINT_CALLED_COL_CONFIGS` and then conditionally use the correct configuration

hanars (changes requested):

sv_pipeline/exome/load_data.py
```diff
@@ -273,14 +306,14 @@ def _parse_row(row, header_indices):
...
-            parse_sv_row(row, parsed_svs_by_name, header_indices, sample_id)
+            parse_sv_row(row, parsed_svs_by_name, header_indices, sample_id, is_new_joint_call)
             found_samples.add(sample_id)
             return True
         else:
             skipped_samples.add(sample_id)
             return False
 
-    load_file(input_dataset, _parse_row, out_file_path=out_file_path)
+    load_file(input_dataset, _parse_row, out_file_path=out_file_path, columns=None if is_new_joint_call else MERGED_CALL_COLUMNS)
```
hanars, line 316: the `columns=None` as the way of handling the default is confusing, I think this would be better if you made columns required and explicitly set it in both cases

sv_pipeline/exome/load_data.py
```diff
@@ -106,18 +118,31 @@ def _parse_genes(genes):
...
+    PREV_MISSING_COL: {'field_name': NEW_CALL_FIELD, 'format': lambda val: False},
+})
+
 CORE_COLUMNS = [CHR_COL, SC_COL, SF_COL, CALL_COL, IN_SILICO_COL]
 SAMPLE_COLUMNS = [
     START_COL, END_COL, QS_COL, CN_COL, NUM_EXON_COL, GENES_COL, DEFRAGGED_COL, PREV_IDENTICAL_COL, PREV_OVERLAP_COL,
     PREV_MISSING_COL,
 ] + list(GENE_CONSEQUENCE_COLS.keys())
+
+MERGED_CALL_SAMPLE_COLUMNS = SAMPLE_COLUMNS + [IS_LATEST]
```
hanars, line 141: the default configuration should be the one thats in the main config, and since merged is now the default you should update the main `COL_CONFIGS`/ `SAMPLE_COLUMNS ` and maintain the joint called configs as separate

sv_pipeline/exome/load_data.py
```diff
@@ -198,7 +227,9 @@ def parse_sv_row(row, parsed_svs_by_id, header_indices, sample_id):
         parsed_svs_by_id[variant_id][GENOTYPES_FIELD] = []
         parsed_svs_by_id[variant_id][GENES_FIELD] = set()
 
-    sample_info = get_parsed_column_values(row, header_indices, SAMPLE_COLUMNS)
+    sample_info = get_parsed_column_values(row, header_indices, SAMPLE_COLUMNS) if is_new_joint_call else \
+        get_parsed_column_values(row, header_indices, MERGED_CALL_SAMPLE_COLUMNS, col_configs=MERGED_CALL_COL_CONFIGS)
```
hanars, line 231: this is pretty hard to read/ maintain, calling the same function with half the args different like this. I think it might be clearer if you made both `columns` and `col_configs` optional in `get_parsed_column_values`.  If `col_configs` is passed in it loads all of those columns, otherwise it uses the subset of columns from the `columns` args. Then this call would be
```
kwargs = {'columns': SAMPLE_COLUMNS} if is_new_joint_call else {'col_configs': MERGED_CALL_COL_CONFIGS}
sample_info = get_parsed_column_values(row, header_indices, **kwargs) 
```

hanars (changes requested):

sv_pipeline/exome/load_data.py
ShifaSZ: Okay

sv_pipeline/exome/load_data.py
ShifaSZ: We don't have a sample column config yet. If we load all columns in the col_configs, it will include more columns than the sample ones.

hanars (changes requested):

ShifaSZ:
Should I also add some tests to ensure it doesn't break any existing functions?

hanars:
> Should I also add some tests to ensure it doesn't break any existing functions?

thats out of scope, since we don't really have *any* tests set up in this part of the repo. But if you really want to you can

PR #381, opened 2022-12-22
Update the input data format for genome SVs

ShifaSZ:
See ticket https://github.com/broadinstitute/seqr/issues/3064


hanars (changes requested):
The idea behind having a separate PR for the tests and then branching off of it for this PR was to make it really clear which data is changing because of function change. You do not achieve that here. You add a whole bunch of new rows here in the new format, and then test them. You need to add those rows to the test fixtures in the test change PR and update the tests accordingly, so in this PR I'm not seeing changed to, say, the GQ of samples when there are no code changes related to GQ

ShifaSZ:
Okay. I will use the same data rows with updated prediction fields.

sv_pipeline/genome/load_data_tests.py
```diff
@@ -157,7 +154,7 @@
 EXPECTED_DATA = [
     hl.Struct(
         contig='1', sc=7, sf=0.910684, sn=8, start=789481, end=789481, sv_callset_Het=520, sv_callset_Hom=2391,
-        gnomad_svs_ID=None, gnomad_svs_AF=None, pos=789481,
+        gnomad_svs_ID=None, gnomad_svs_AF=None, gnomad_svs_AC=None, gnomad_svs_AN=None, pos=789481,
```
hanars, line 157: I'd prefer to test one row where we can test that the new fields are populated

hanars (changes requested):
Overall looks good, just one small request to add a test for the new fields

hanars (approved):
Approved, but we should first merge the test branch and then merge this branch to master, don't merge this into the test branch

ShifaSZ:
The gnomAD ID format was changed. I think you may want to look it over again.

hanars (approved):

PR #382, opened 2023-01-19
Remove travis file

mike-w-wilson:
KC brought up that BITs notified us that we should rotate the PAT in the travis config. Steve recommends we delete the file as it doesnt appear to be used. See relevant slack thread here: https://the-tgg.slack.com/archives/CG83SEN1Z/p1671211900455049.


bpblanken (approved):
I approve of this.  

We'll get to GitHub actions shortly with some functioning tests!

PR #386, opened 2023-02-17
Just trying to get the existing tests to pass!

bpblanken:
This was definitely more than anticipated... but it got me up to my waist in how this code works!

Next steps from me will be to port over the two patch file tickets I did then get rolling on the SV pipeline.


luigi_pipeline/lib/model/base_mt_schema.py
```diff
@@ -28,6 +28,18 @@ def __repr__(self):
             requires = f' (requires: {", ".join(self.requirements)})'
         return f"{self.name}{requires}"
 
+    def __call__(self: "RowAnnotation", schema: "BaseMTSchema", overwrite: bool = False):
```
bpblanken, line 31: Several of the tests expected these to be callable... at this point I'd vote to just remove the tests but that also kind of seems like a bad idea 🤷.

luigi_pipeline/seqr_loading.py
```diff
@@ -148,6 +124,30 @@ def annotate_old_and_split_multi_hts(self, mt):
         # Named `locus_old` instead of `old_locus` because split_multi_hts drops `old_locus`.
         return hl.split_multi_hts(mt.annotate_rows(locus_old=mt.locus, alleles_old=mt.alleles))
 
+    @staticmethod
+    def contig_check(mt, standard_contigs, threshold):
```
bpblanken, line 128: I moved this method onto the class so that the mocking looked similar to the existing mocking.

luigi_pipeline/tests/test_seqr_loading_tasks.py
```diff
@@ -1,52 +1,66 @@
...
+        mock_sample_type_stats.return_value = None
+        self.assertRaises(SeqrValidationError, SeqrVCFToMTTask.validate_mt, None, '37', None)
+        self.assertRaises(SeqrValidationError, SeqrVCFToMTTask.validate_mt, 12345, '37', None)
+        self.assertRaises(SeqrValidationError, SeqrVCFToMTTask.validate_mt, {}, '37', None)
+
+    @patch('lib.hail_tasks.HailMatrixTableTask.sample_type_stats')
+    def test_seqr_loading_validate_match_none(self, mock_sample_type_stats, mock_contig_check):
         # Matched none should fail.
         mock_sample_type_stats.return_value = self._sample_type_stats_return_value(0, 0, False, 0, 0, False)
-        self.assertRaises(SeqrValidationError, SeqrVCFToMTTask.validate_mt, None, '37', None)
```
bpblanken, line 22: Calling `validate_mt` on a `None` `mt` was 😬.

luigi_pipeline/lib/model/base_mt_schema.py
ShifaSZ: Those tests look ugly. It may be better if the tests are done at a higher level (e.g., calling `annotate_all()` rather than each annotation function). But I like your change that makes the annotation callable.

bpblanken: Yep!  Agreed.  My assumption when starting this was that we should try to keep the tests and manipulate the code so that the tests passed... but I think the better answer here is probably to just get rid of the tests that called annotations individually (which I think was most of them :() .

luigi_pipeline/lib/model/base_mt_schema.py
```diff
@@ -57,7 +69,10 @@ def mt_prop_wrapper(func):
             fn_requirements = fn_require if isinstance(fn_require, list) else [fn_require]
             for fn in fn_requirements:
                 if not isinstance(fn, RowAnnotation):
-                    raise ValueError('Schema: dependency %s is not a row annotation method.' % fn_require.__name__)
+                    raise ValueError(
+                        'Schema: dependency %s is not a row annotation method.' 
+                        % (fn_require.__name__ if hasattr(fn_require, "__name__") else str(fn_require))
```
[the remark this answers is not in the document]
bpblanken: 👍

luigi_pipeline/tests/test_seqr_loading_tasks.py
```diff
@@ -1,52 +1,66 @@
...
 
+import hail as hl
+
 from seqr_loading import SeqrVCFToMTTask, SeqrValidationError
 from tests.data.sample_vep import VEP_DATA, DERIVED_DATA
 
 TEST_DATA_MT_1KG = 'tests/data/1kg_30variants.vcf.bgz'
 
 
+@patch('seqr_loading.SeqrVCFToMTTask.contig_check', return_value={})
```
hanars, line 12: are we losing test coverage for contig_check, or was there never any testing for it in the first place?

bpblanken: We had no coverage... and it was breaking the tests (it doesn't handle a `None` matrix table and all the tests used an empty matrix table).   I made the decision to just disable running the function because it was actually altering the behavior of some of the existing tests, which I assume were written before `contig_check` existed.

ShifaSZ (approved):
It looks good to me.

PR #387, opened 2023-02-17
Fix validate mt error messages.

### LINKED ISSUE #256: Fix validate MT error messages for WES and WGS
In this function: https://github.com/broadinstitute/hail-elasticsearch-pipelines/blob/e2a7e08b356ce70af314507908c7df4a9f128d7b/luigi_pipeline/seqr_loading.py#L99 The text in the errors raised for when a VCF has or does not have coding and non_coding variants is wrong. The third check should say `...appears to be WES...` while the fourth should be `...appears to be WGS...`

bpblanken:
Resolves #256.

```
PYTHONPATH=.. nosetests tests/test_seqr_loading_tasks.py

LOGGING: writing to /Users/bblanken/Desktop/code/seqr-loading-pipelines/luigi_pipeline/hail-20230217-1129-0.2.109-b71b065e4bb6.log
.......
Ran 7 tests in 1.865s

OK
```

### DIFF
--- a/luigi_pipeline/seqr_loading.py
+++ b/luigi_pipeline/seqr_loading.py
@@ -196,15 +196,19 @@ class SeqrVCFToMTTask(HailMatrixTableTask):
             # Only coding should be WES.
             if sample_type != 'WES':
                 raise SeqrValidationError(
-                    'Sample type validation error: dataset sample-type is specified as {} but appears to be '
-                    'WGS because it contains many common coding variants'.format(sample_type)
+                    '{} because it contains many common non-coding variants'.format(
+                        "WGS",
+                        "WES",
+                    )
                 )
         elif has_noncoding and has_coding:
             # Both should be WGS.
             if sample_type != 'WGS':
                 raise SeqrValidationError(
-                    'Sample type validation error: dataset sample-type is specified as {} but appears to be '
-                    'WES because it contains many common non-coding variants'.format(sample_type)
+                    '{} because it contains many common non-coding variants'.format(
+                        "WES",
+                        "WGS",
+                    )
                 )
         return True
 
--- a/luigi_pipeline/seqr_loading.py
+++ b/luigi_pipeline/seqr_loading.py
@@ -196,7 +196,8 @@ class SeqrVCFToMTTask(HailMatrixTableTask):
             # Only coding should be WES.
             if sample_type != 'WES':
                 raise SeqrValidationError(
-                    '{} because it contains many common non-coding variants'.format(
+                    'Sample type validation error: dataset sample-type is specified as {} but appears to be '
+                    '{} because it contains many common coding variants'.format(
                         "WGS",
                         "WES",
                     )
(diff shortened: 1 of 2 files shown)


luigi_pipeline/seqr_loading.py
```diff
@@ -200,14 +200,20 @@ def validate_mt(mt, genome_version, sample_type):
             if sample_type != 'WES':
                 raise SeqrValidationError(
                     'Sample type validation error: dataset sample-type is specified as {} but appears to be '
-                    'WGS because it contains many common coding variants'.format(sample_type)
+                    '{} because it contains many common coding variants'.format(
+                        "WGS",
+                        "WES",
```
mike-w-wilson, line 205: Thoughts on just filling in the data type and making this a normal string we don't need to format? Not sure if there is a best practice around errors and formatting string but feel a basic string "...sample-type is specified as WGS  but appears to be WES...." is more readable?

luigi_pipeline/seqr_loading.py
```diff
@@ -200,14 +200,20 @@ def validate_mt(mt, genome_version, sample_type):
...
         elif has_noncoding and has_coding:
             # Both should be WGS.
             if sample_type != 'WGS':
                 raise SeqrValidationError(
                     'Sample type validation error: dataset sample-type is specified as {} but appears to be '
-                    'WES because it contains many common non-coding variants'.format(sample_type)
+                    '{} because it contains many common non-coding variants'.format(
+                        "WES",
+                        "WGS",
+                    )
```
mike-w-wilson, line 216: Same comment as above about this being a basic string.

luigi_pipeline/tests/test_seqr_loading_tasks.py
```diff
@@ -57,10 +57,11 @@ def test_seqr_loading_validate_match_noncoding_only(self, mock_sample_type_stats
     def test_seqr_loading_validate_wes_mismatch(self, mock_sample_type_stats, mock_contig_check):
         # Supposed to be WES but we report as WGS.
         mock_sample_type_stats.return_value = self._sample_type_stats_return_value(0, 0, False, 0, 0, True)
-        self.assertRaises(SeqrValidationError, SeqrVCFToMTTask.validate_mt, self.test_mt, '37', 'WGS')
+        self.assertRaisesRegex(SeqrValidationError, "specified as WGS but appears to be WES", SeqrVCFToMTTask.validate_mt, self.test_mt, '37', 'WGS')
```
mike-w-wilson, line 60: Oh does the above formatting have to do with this test?

mike-w-wilson (commented):
One thing I'd like clarification on FMK regarding the data type used in the error.

luigi_pipeline/seqr_loading.py
bpblanken: No real reason as is!  I think ideally we'd have one error function that would take the sample types as arguments (which is what I attempted and then backtracked to this...) but just plopping in the strings is fine!

luigi_pipeline/tests/test_seqr_loading_tasks.py
bpblanken: Nope!  It should work fine as is.

mike-w-wilson (approved):
LGTM!

hanars (approved):

PR #389, opened 2023-02-17
Delete `SeqrMTToESTask`, which isn't used?

bpblanken:
Leaving this as a draft because this is more of a question to ensure I'm following the abstractions more than anything else.  

My understanding is that `SeqrMTToESOptimizedTask` inside `seqr_loading_optimized` is the main entry point and that that task (which does eventually wrap the `SeqrVCFToMTTask` inside `seqr_loading.py`) replaced `SeqrMTToESTask`.


hanars:
this un-optimized task is still used in the regular `seqr_loading` script which is what we package in our docker image and recommend for local installations of seqr: https://github.com/broadinstitute/seqr-loading-pipelines/blob/master/docker/bin/load_data.sh#L25

bpblanken:
Awesome!  Thank you for clarifying!

PR #390, opened 2023-02-17
I think this should be a variant schema... though there's no functional…

bpblanken:
… impact because it's immediately overridden

@ShifaSZ Tagged you on this because it's small and should be non-functional!  Just wanting to make sure I'm understanding this!


bpblanken:
I don't think this is right after reading more of this... closing!

PR #391, opened 2023-02-21
First pass at SV Schema.

bpblanken:
This is close enough that I'd like 👀 to make sure I haven't rocketed off in the wrong direction.  Going to work a bit on porting the old `load_data_tests.py` over.


luigi_pipeline/seqr_sv_loading.py
```diff
@@ -0,0 +1,49 @@
...
+    gencode_path = luigi.OptionalParameter(default="", description="Path for downloaded gencode data")
+    RUN_VEP = False
+    SCHEMA_CLASS = SeqrSVVariantSchema
+
+    # NB: electing not to override import_vcf here eventhough the inherited args are slightly different
+    # than from the old pipeline.
+
+    def get_schema_class_kwargs(self):
+        return {
+            "gene_id_mapping" : hl.literal(load_gencode(self.gencode_release, self.gencode_path))
```
bpblanken, line 34: Opted to just use the existing `load_gencode` function for now... though I imagine we'll want to port it to something that looks like what we have in `download_and_create_reference_datas`.

sv_pipeline/genome/utils/download_utils.py
```diff
@@ -4,7 +4,12 @@
 import tempfile
 from contextlib import contextmanager
 import hail as hl
-from tqdm import tqdm
+
+try:
```
bpblanken, line 8: This is gross.  Other options:
1) Just add `tqdm` to the requirements (at the time this seemed worse to me but 🤷 ).
2) Just remove it.

luigi_pipeline/lib/model/sv_mt_schema.py
```diff
@@ -0,0 +1,241 @@
...
+    def sn(self):
+        return self.mt.info.AN
+
+    @row_annotation()
+    def start(self):
+        # 'start': lambda rows: rows.locus.position,
+        return variant_id.get_expr_for_start_pos(self.mt)
+
+    @row_annotation()
+    def end(self):
```
bpblanken, line 57: This deviates from the seqr_mt_schema which uses:

```
@row_annotation(disable_index=True)
    def end(self):
        return variant_id.get_expr_for_end_pos(self.mt)
```

just curious about why?

luigi_pipeline/lib/model/sv_mt_schema.py
```diff
@@ -0,0 +1,240 @@
...
+            )
+        )
+
+    @row_annotation(name='variantId')
+    def variant_id(self):
+        return self.mt.rsid
+
+    # NB: This is the "elasticsearch_mapping_id" used inside of export_table_to_elasticsearch.
+    @row_annotation(name='docId', disable_index=True)
+    def doc_id(self, max_length=512):
```
bpblanken, line 214: I added this to make [this line](https://github.com/broadinstitute/seqr-loading-pipelines/blob/e2109adfc508112be0c613ad2efb68e884b861ce/luigi_pipeline/lib/hail_tasks.py#L256) work in the elastic search export.  But the field itself is not present in the `load_data` export.  Was this the right thing to do?

sv_pipeline/genome/utils/download_utils.py
bpblanken: I went with option 1 after re-considering.

luigi_pipeline/lib/model/sv_mt_schema.py
```diff
@@ -0,0 +1,240 @@
...
+
+def get_cpx_interval(x):
+    # an example format of CPX_INTERVALS is "DUP_chr1:1499897-1499974"
+    type_chr = x.split('_chr')
+    chr_pos = type_chr[1].split(':')
+    pos = chr_pos[1].split('-')
+    return hl.struct(type=type_chr[0], chrom=chr_pos[0], start=hl.int32(pos[0]), end=hl.int32(pos[1]))
+
+
+class SeqrSVVariantSchema(BaseMTSchema):
```
hanars, line 28: It seems like several of these are direct copy-pastes from `BaseSeqrSchema`. Might it be petter to have this inherit from that instead, or to move some of those annotations down form `BaseSeqrSchema` to `BaseMTSchema`?

bpblanken: I think my thought there was that a couple fields were (`contig` and `start`) but that most were not, and that there were more things on `BaseSeqrSchema` that I did not want to inherit.  I'm ok with moving those few things to `BaseMTSchema` though!

luigi_pipeline/lib/model/sv_mt_schema.py
```diff
@@ -0,0 +1,240 @@
...
+
+    def __init__(self, *args, gene_id_mapping, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.gene_id_mapping = gene_id_mapping
```
bpblanken, line 32: Noticed I didn't make this private.  Will fix.

luigi_pipeline/tests/test_seqr_sv_loading.py
```diff
@@ -0,0 +1,305 @@
...
+        self._variant_mt_file = tempfile.mkstemp(dir=self._temp_dir.name, suffix='.mt')[1]
+        self._genotypes_mt_file = tempfile.mkstemp(dir=self._temp_dir.name, suffix='.mt')[1]
+        with open(self._vcf_file, 'w') as f:
+            f.writelines('\n'.join(VCF_DATA))
+
+    def tearDown(self):
+        shutil.rmtree(self._temp_dir.name)
+
+    @mock.patch('seqr_sv_loading.load_gencode', return_value=GENE_ID_MAPPING)
+    def test_run_task(self, load_gencode_mock):
```
ShifaSZ, line 258: I'm curious about what happened to the Elasticsearch exporting. It must have been mocked, but I don't see the mocking.

ShifaSZ (changes requested):

ShifaSZ:
Here is the [doc](https://docs.google.com/document/d/1_FyERQB0vxcUo1yKm0C_fgQCl8sFuIHfInNYgoOTy08/edit#) about the data fields of the genome SVs  for your reference.

luigi_pipeline/lib/hail_tasks.py
```diff
@@ -48,6 +48,10 @@ class HailMatrixTableTask(luigi.Task):
     genome_version = luigi.Parameter(description='Reference Genome Version (37 or 38)')
     vep_runner = luigi.ChoiceParameter(choices=['VEP', 'DUMMY'], default='VEP', description='Choice of which vep runner'
                                                                                             'to annotate vep.')
+    ignore_missing_samples = luigi.BoolParameter(default=False, description="Allow missing samples in the callset.")
+    skip_sample_subset = luigi.BoolParameter(default=False, description=
+        "Skip subsetting samples... only subset on present variants."
```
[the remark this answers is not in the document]
bpblanken: I don't think there's any reason no!  I think even python linters are pretty lenient on single/double quotes since both are supported.  I will change these to single quotes.

I saw some double quotes floating around: https://github.com/broadinstitute/seqr-loading-pipelines/blob/master/luigi_pipeline/seqr_loading.py#L70-L73, so didn't try hard to be consistent but do think it's important!

luigi_pipeline/lib/hail_tasks.py
```diff
@@ -48,6 +48,10 @@ class HailMatrixTableTask(luigi.Task):
...
+    )
```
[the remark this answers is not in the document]
bpblanken: Good call out!  I was generally trying to mimic the behavior of the existing `load_data`, which used a `skip_sample_subset` param.  But `load_data` was fetching the sample ids from GCS rather than using the path as an input param.  I will fix!

luigi_pipeline/lib/hail_tasks.py
```diff
@@ -124,36 +128,34 @@ def run_vep(mt, genome_version, runner='VEP', vep_config_json_path=None):
...
-            else:
-                raise MatrixTableSampleSetError(message, missing_samples)
-
-        mt = mt.semi_join_cols(subset_ht)
-        mt = mt.filter_rows(hl.agg.any(mt.GT.is_non_ref()))
-
-        logger.info(f'Finished subsetting samples. Kept {subset_count} '
-                    f'out of {mt.count()} samples in vds')
-        return mt
+        if not self.skip_sample_subset:
```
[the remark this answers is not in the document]
bpblanken: Good suggestion.  Note that the last line of this function filters to only variants that have a non_ref genotype.  Do we need that?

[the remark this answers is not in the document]
bpblanken: I believe I was mirroring the `load_data` behavior from [here](https://github.com/broadinstitute/seqr-loading-pipelines/blob/master/sv_pipeline/genome/load_data.py#L131-L159) where that line is called regardless of whether or not we skip the sample subset.

luigi_pipeline/lib/model/sv_mt_schema.py
```diff
@@ -0,0 +1,219 @@
...
+        conseq_predicted_gene_cols = [
+            gene_col for gene_col in self.mt.info if gene_col.startswith(CONSEQ_PREDICTED_PREFIX)
+            and gene_col not in NON_GENE_PREDICTIONS
+        ]
+        mapped_genes = [
+            self.mt.info[gene_col].map(
+                lambda gene: hl.struct(**{
+                    'gene_symbol': gene,
+                    'gene_id': self._gene_id_mapping.get(gene, hl.missing(hl.tstr)),
+                    'major_consequence': gene_col.replace(CONSEQ_PREDICTED_PREFIX, '', 1)
```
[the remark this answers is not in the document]
bpblanken: 👍

luigi_pipeline/seqr_loading.py
```diff
@@ -32,13 +32,13 @@ class SeqrVCFToMTTask(HailMatrixTableTask):
...
-    hgmd_ht_path = luigi.Parameter(default=None,
+    hgmd_ht_path = luigi.OptionalParameter(default=None,
                                    description='Path to the Hail table storing the hgmd variants.')
     sample_type = luigi.ChoiceParameter(choices=['WGS', 'WES'], description='Sample type, WGS or WES', var_type=str)
     dont_validate = luigi.BoolParameter(description='Disable checking whether the dataset matches the specified '
                                                     'genome version and WGS vs. WES sample type.')
-    dataset_type = luigi.ChoiceParameter(choices=['VARIANTS', 'SV'], default='VARIANTS',
-                                         description='VARIANTS or SV.')
+    dataset_type = luigi.ChoiceParameter(choices=['VARIANTS', 'SV', 'MITO'], default='VARIANTS',
+                                         description='VARIANTS or SV or MITO.')
```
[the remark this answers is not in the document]
bpblanken: 😜

luigi_pipeline/lib/model/sv_mt_schema.py
```diff
@@ -0,0 +1,219 @@
...
+            lambda x: (x != PASS) & (x != BOTHSIDES_SUPPORT)
+        ))
+        return hl.if_else(
+            hl.len(filters) > 0,
+            filters,
+            hl.missing(hl.dtype('array<str>')),
+        )
+
+    @row_annotation()
+    def bothsides_support(self):
```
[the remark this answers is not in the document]
bpblanken: Added for those fields!

luigi_pipeline/seqr_loading_optimized.py
```diff
@@ -65,6 +65,7 @@ def run(self):
         variants_mt = hl.read_matrix_table(self.input()[0].path)
         genotypes_mt = hl.read_matrix_table(self.input()[1].path)
         genotypes_mt = genotypes_mt.drop(*[k for k in genotypes_mt.globals.keys()])
+        genotypes_mt = genotypes_mt.drop(*[k for k in genotypes_mt.row.keys() if k in variants_mt.row.keys() and k not in variant_mt.row_key])
```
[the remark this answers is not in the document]
bpblanken: I ended up re-doing this... have a class called `BaseVariantsSchema` which accomplishes the shared fields between SNV, MITO, and SV...

luigi_pipeline/lib/model/base_mt_schema.py
```diff
@@ -4,6 +4,8 @@
 from inspect import getmembers
 from collections import defaultdict
 
+from hail_scripts.computed_fields import variant_id
+
```
ShifaSZ, line 8: Unused.

ShifaSZ (changes requested):
I think this looks good to me. But there is an unused code that is better to be removed.

luigi_pipeline/lib/model/base_mt_schema.py
bpblanken: yep!  good catch.

luigi_pipeline/seqr_loading.py
```diff
@@ -103,15 +106,16 @@ def read_input_write_mt(self):
             mt = self.subset_samples_and_variants(mt, self.subset_path)
```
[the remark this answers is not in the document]
ShifaSZ: Agree. There should be at least one GT that is non-reference when there is no subsetting.

ShifaSZ (approved):

luigi_pipeline/tests/test_seqr_sv_loading.py
```diff
@@ -0,0 +1,303 @@
...
+                   hl.Struct(sample_id='SAMPLE-3', gq=0, cn=2, num_alt=1),
+                   hl.Struct(sample_id='SAMPLE-4', gq=99, cn=3, num_alt=0),
+                   hl.Struct(sample_id='SAMPLE-5', gq=99, cn=1, num_alt=0)],
+        **{"samples_num_alt.1": {'SAMPLE-2', 'SAMPLE-3'}, "samples_num_alt.2": None},
+        **{key: EXPECTED_SAMPLE_GQ[2].get(key) for key in SAMPLES_GQ_SV_FIELDS}
+    )
+]
+
+
+class SeqrSvLoadingTest(unittest.TestCase):
```
mike-w-wilson, line 244: Nitpick but maybe 
```suggestion
class SeqrSVLoadingTest(unittest.TestCase):
```
for consistency?

mike-w-wilson (commented):
This is looking good to me! I have a few points/nitpicks like using types in some missing statements, e.g. hl.tint vs hl.dtype('int'), and also an alternative approach in some if_else statements. I am starting to think we need to rename the classes using `Variant` from the "SNV" loading pipeline but that's certainly out of scope here.

luigi_pipeline/lib/model/sv_mt_schema.py
```diff
@@ -0,0 +1,225 @@
...
+    def __init__(self, *args, gene_id_mapping=None, **kwargs):
+        super().__init__(*args, **kwargs)
+        self._gene_id_mapping = gene_id_mapping
+
+    def sv_types(self):
+        return self.mt.alleles[1].replace('[<>]', '').split(':', 2)
+
+    @row_annotation()
+    def sc(self):
+        return self.mt.info.AC[0]
```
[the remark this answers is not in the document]
mike-w-wilson: Sorry I forgot hail just imports AC as an array. I got confused because hl.agg.callstats produces an array with AC = [ref AC, alt AC] so wanted to make sure.

luigi_pipeline/lib/model/seqr_mt_schema.py
```diff
@@ -4,8 +4,29 @@
 from hail_scripts.computed_fields import variant_id
 from hail_scripts.computed_fields import vep
 
+class BaseVariantsSchema(BaseMTSchema):
```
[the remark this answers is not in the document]
bpblanken: I'll just go with `BaseVariantSchema`!  We do have `SeqrGenotypes` and `SeqrVariantsandGenotypes` though and we can revisit later!

luigi_pipeline/lib/model/sv_mt_schema.py
```diff
@@ -0,0 +1,225 @@
...
+
+    @row_annotation()
+    def filters(self):
+        filters = hl.array(self.mt.filters.filter(
+            lambda x: (x != PASS) & (x != BOTHSIDES_SUPPORT)
+        ))
+        return hl.if_else(
+            hl.len(filters) > 0,
+            filters,
+            hl.missing(hl.dtype('array<str>')),
```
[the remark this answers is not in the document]
bpblanken: I think we can maybe just do `hl.or_missing(hl.len(filters) > 0, filters)`?  I didn't know that function existed at the time!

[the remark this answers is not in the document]
bpblanken: I also (I think), ripped this logic right from this line: https://github.com/broadinstitute/seqr-loading-pipelines/blob/master/sv_pipeline/genome/load_data.py#L181

luigi_pipeline/lib/model/sv_mt_schema.py
```diff
@@ -0,0 +1,217 @@
...
+        return {
+            'sample_id': self.mt.s,
+            'gq': self.mt.GQ,
+            'cn': self.mt.RD_CN,
+            'num_alt': hl.if_else(is_called, self.mt.GT.n_alt_alleles(), -1)
+        }
+
+    def _genotype_filter_samples(self, filter):
+        # NB: override this function here to mimic the existing null handling behavior.
+        samples = hl.set(self.mt.genotypes.filter(filter).map(lambda g: g.sample_id))
```
mike-w-wilson, line 201: Similar to above, just wondering why this is converted to a set? An alternative is to keep it as an arrayexpression and call .size() below.

mike-w-wilson (approved):
Sorry two things but then LGTM!

luigi_pipeline/lib/model/sv_mt_schema.py
bpblanken: Good question!  I'm actually just inheriting from `SeqrGenotypesSchema` here with copy-paste inheritance.  I _think_ we can just use `super()` here (I had a bit of trouble inheriting functions on these classes because of the `row_annotation` decorator but that doesn't apply on this function!).

As for why `set` is present in general, I'm less sure.  Maybe the sample can end up in a bucket twice and we want to remove?

mike-w-wilson: Ah sorry -- I misread L202 --I was thinking that it's only checking if samples > 0 so removing a dup shouldnt matter since we'd still get 1 in the set but it actually returns the sample set 🤦 . I cant think off the top of my head where wed see dup sample IDs in the genotype field on a variant..but makes sense to keep it and we can always revisit.

luigi_pipeline/lib/model/sv_mt_schema.py
```diff
@@ -0,0 +1,217 @@
...
+    @row_annotation(name='gnomad_svs_AN')
+    def gnomad_svs_an(self):
+        return self.mt.info.gnomAD_V2_AN_AF
+
+    @row_annotation()
+    def filters(self):
+        filters = hl.array(self.mt.filters.filter(
+            lambda x: (x != PASS) & (x != BOTHSIDES_SUPPORT)
+        ))
+        return hl.or_missing(hl.len(filters) > 0, filters)
```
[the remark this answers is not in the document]
bpblanken: 👍

luigi_pipeline/lib/model/sv_mt_schema.py
```diff
@@ -0,0 +1,217 @@
...
+
+class SeqrSVGenotypesSchema(SeqrGenotypesSchema):
+
+    def _genotype_fields(self):
+        is_called = hl.is_defined(self.mt.GT)
+        return {
+            'sample_id': self.mt.s,
+            'gq': self.mt.GQ,
+            'cn': self.mt.RD_CN,
+            'num_alt': hl.if_else(is_called, self.mt.GT.n_alt_alleles(), -1)
```
hanars, line 196: not very important, but it would be nice if this could inherit from the SeqrGenotypesSchema definition of num_alt

luigi_pipeline/lib/model/sv_mt_schema.py
```diff
@@ -0,0 +1,217 @@
...
+    # NB: override this function here to mimic the existing null handling behavior.
+    def _genotype_filter_samples(self, filter_):
+        samples = super()._genotype_filter_samples(filter_)
+        return hl.or_missing(hl.len(samples) > 0, samples)            
+
+    @row_annotation(name="samples_gq_sv", fn_require=SeqrGenotypesSchema.genotypes)
+    def samples_gq(self):
+        # NB: super().samples_gq is a RowAnnotation... so we call the method under the hood.
+        # ew it is gross.
+        return super().samples_gq.fn(self, start=0, end=1000, step=10)
```
hanars, line 208: would the change you made to make RowAnnotation callable mean you could do `super().samples_gq(self, start=0, end=1000, step=10)` or no?

luigi_pipeline/requirements.txt
```diff
@@ -2,3 +2,4 @@ elasticsearch==7.9.1
 hail>=0.2.77
 luigi==2.8.11
 google-api-python-client>=1.8.0
+tqdm==4.64.1
```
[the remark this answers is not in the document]
hanars: I think we can just take tqdm out of the gencode downloader if we prefer

luigi_pipeline/seqr_sv_loading.py
hanars: I agree that downloading all of genocode every single time we run the pipeline is overkill just to get a gene symbol to id mapping. It probably makes the *most* sense to ask the team that delivers the file to do so with ensemble IDs, which is what we asked them to do for gCNV data. But in the interim this is fine

hanars (changes requested):

hanars:
It look slike this is stil missing the `StrVCTVRE_score` annotation. Its okay to assume that the input to the pipeline will already have strvctre run on it and available as `mt.info.StrVCTVRE`

luigi_pipeline/lib/model/sv_mt_schema.py
bpblanken: I think it would work if we changed the `__call__` to take in **kwargs which would then be passed to the function.  That might be cleaner?

bpblanken: Tried it and it doesn't work.  Calling the RowAnnotation doesn't return the function result but returns the instance of the schema class... which is very unintuitive.

There were a couple of tests doing things like:

```
test_schema.a().b().c_1()
```
and 
```
mt = test_schema.d().select_annotated_mt()
```

I'm going to leave this for now and we can revisit.

luigi_pipeline/lib/model/sv_mt_schema.py
bpblanken: 👍

luigi_pipeline/requirements.txt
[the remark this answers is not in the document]
bpblanken: I've elected to rip out tqdm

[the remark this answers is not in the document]
bpblanken: `load_gencode` still seems to work!

```
bblanken@wmbbe-67c seqr-loading-pipelines % ls -alth gencode.v42.annotation.gtf.*
-rw-r--r--@ 1 bblanken  staff   1.8M Mar 10 17:06 gencode.v42.annotation.gtf.pickle
-rw-r--r--@ 1 bblanken  staff    47M Mar 10 17:06 gencode.v42.annotation.gtf.gz
```

hanars (approved):

PR #398, opened 2023-03-01
GCNV Schema and Tests

bpblanken:
Resolves https://github.com/broadinstitute/seqr-loading-pipelines/issues/394

```
Ran 4 tests in 85.164s

OK
```


luigi_pipeline/lib/model/gcnv_mt_schema.py
```diff
@@ -0,0 +1,212 @@
...
+        return hl.if_else(hl.len(samples) > 0, samples, hl.missing(hl.dtype('array<str>')))
+    
+    def _genotype_fields(self):
+        return {
+            'sample_id': self.mt.sample_fix.first_match_in(SAMPLE_ID_REGEX)[0],
+            'qs': self.mt.QS,
+            'cn': self.mt.CN,
+            'defragged': self.mt.defragmented,
+            # Hail expression is to bool-ify a string value.
+            'prev_call': hl.if_else(hl.len(self.mt.identical_ovl) > 0, True, False) if self._is_new_joint_call else ~self.mt.is_latest,
```
bpblanken, line 199: This was the gnarliest logic to reproduce (or at least caused me the most headaches!). Here's the old code for reference: https://github.com/broadinstitute/seqr-loading-pipelines/blob/master/sv_pipeline/exome/load_data.py#L121-L133

luigi_pipeline/lib/model/gcnv_mt_schema.py
```diff
@@ -0,0 +1,212 @@
...
+
+    def samples_ab(self):
+        pass
+
+    @row_annotation(fn_require=SeqrGenotypesSchema.genotypes)
+    def samples_qs(self, start=0, end=1000, step=10):
+        return hl.struct(**{
+            '%i_to_%i' % (i, i+step): self._genotype_filter_samples(lambda g: ((g.qs >= i) & (g.qs < i+step)))
+            for i in range(start, end, step)
+        }, **{
```
bpblanken, line 175: This code (AND the old code) doesn't handle `qs == 1000` exactly well.  I think that's ok?

luigi_pipeline/lib/model/gcnv_mt_schema.py
```diff
@@ -0,0 +1,212 @@
...
+        return {
+            'sample_id': self.mt.sample_fix.first_match_in(SAMPLE_ID_REGEX)[0],
+            'qs': self.mt.QS,
+            'cn': self.mt.CN,
+            'defragged': self.mt.defragmented,
+            # Hail expression is to bool-ify a string value.
+            'prev_call': hl.if_else(hl.len(self.mt.identical_ovl) > 0, True, False) if self._is_new_joint_call else ~self.mt.is_latest,
+            'prev_overlap': hl.if_else(hl.len(self.mt.any_ovl) > 0, True, False)  if self._is_new_joint_call else False,
+            # NB: previous implementation also falsified NA, but hail treats NA as an empty value.
+            'new_call': self.mt.no_ovl if self._is_new_joint_call else False,
```
bpblanken, line 202: I tried to default to hail's `NA` handling in this implementation... but the behavior might be slightly different on this one: https://github.com/broadinstitute/seqr-loading-pipelines/blob/master/sv_pipeline/exome/load_data.py#L132

luigi_pipeline/lib/model/gcnv_mt_schema.py
```diff
@@ -0,0 +1,212 @@
...
+    def strvctvre(self):
+       return self.mt.strvctvre_score
+
+    @row_annotation(name='variantId')
+    def variant_id(self):
+        return hl.format(f"%s_%s_{datetime.date.today():%m%d%Y}", self.mt.variant_name, self.mt.svtype)
+
+    @row_annotation()
+    def start(self):
+        return hl.agg.min(self.mt.sample_start)
```
bpblanken, line 71: These are a little weird but duplicate the existing behavior.  Not sure we ever see different `start` and `end` over different samples?

luigi_pipeline/lib/model/gcnv_mt_schema.py
```diff
@@ -0,0 +1,212 @@
...
+
+    @row_annotation(name='geneIds')
+    def gene_ids(self):
+        return hl.array(hl_agg_collect_set_union(parse_genes(self.mt.genes_any_overlap_Ensemble_ID)))
+
+    @row_annotation(name='transcriptConsequenceTerms', fn_require=sv_type)
+    def transcript_consequence_terms(self):
+        formatted_gcnv_consequence = hl.format('gCNV_%s', self.mt.svType)
+        lof = hl.len(hl_agg_collect_set_union(parse_genes(self.mt.genes_LOF_Ensemble_ID))) > 0
+        copy_gain = hl.len(hl_agg_collect_set_union(parse_genes(self.mt.genes_CG_Ensemble_ID))) > 0
```
hanars, line 89: You are already computing these consequence lists below in `sorted_transcript_consequences`. It would be neater to make this function depend on that function and then compute this list based on the values of `major_consequence` in that field

luigi_pipeline/lib/model/gcnv_mt_schema.py
hanars: oh good catch, lets fix the bug and have if `qs` is exactly 1000 its in the `samples_qs_gt_1000` bin

luigi_pipeline/lib/model/gcnv_mt_schema.py
```diff
@@ -0,0 +1,212 @@
...
+    def samples_gq(self):
+        pass
+
+    def samples_ab(self):
+        pass
+
+    @row_annotation(fn_require=SeqrGenotypesSchema.genotypes)
+    def samples_qs(self, start=0, end=1000, step=10):
+        return hl.struct(**{
+            '%i_to_%i' % (i, i+step): self._genotype_filter_samples(lambda g: ((g.qs >= i) & (g.qs < i+step)))
```
hanars, line 173: I think we try to use f-string for string formatting, but I defer to Mike on style for this repo

luigi_pipeline/lib/model/gcnv_mt_schema.py
hanars: You don't need to do `hl.if_else(condition, True, False)`, you can just do `condition`.  Also, the string "FALSE" is truthy so I don't think `~self.mt.is_latest` will work the way we want here

luigi_pipeline/lib/model/gcnv_mt_schema.py
hanars: It might actually  be easier to read if you define a whole dictionary based on whether or not its a new join call, instead of having al the fields be `if else` inline.
 So something like
```
fields = {...}
if self._is_new_joint_call:
  fields.update({'prev_overlap': ....})
else:
  fields.update({'prev_overlap': ....})
```

hanars: I don't think hail has any special `NA` handling here,  because theres an actual hail missing value and then theres a string that says "NA" and the latter is what we have in our data and hail will just treat that as a string

luigi_pipeline/lib/model/gcnv_mt_schema.py
```diff
@@ -0,0 +1,212 @@
...
+        }
+
+class SeqrGCNVVariantsAndGenotypesSchema(SeqrGCNVVariantSchema, SeqrGCNVGenotypesSchema):
+    
+    # NB: we override this method because the row keys are different.
+    @staticmethod
+    def elasticsearch_row(ds):
+        if isinstance(ds, hl.MatrixTable):
+            ds = ds.rows()
+        return table.drop(table.variant_name, table.svtype)
```
hanars, line 212: you should be able to get `row_key = table.row_key` before you do the flattening and then do `table = table.drop(*row_key)` here, and that should work for both schemas and then you could just update the base class method

hanars (changes requested):

hanars:
Overall this looks really good to me, mostly just feedback on syntax. I really like the MT schema you've chose to use for this data

luigi_pipeline/lib/model/gcnv_mt_schema.py
```diff
@@ -0,0 +1,212 @@
...
+    """
+    return hl.set(hl.map(
+        lambda gene: gene.split(r'\.')[0],
+        hl.filter(
+            lambda gene: ~hl.set({'None', 'null', 'NA', ''}).contains(gene),
+            gene_col.split(','),
+        )
+    ))
+
+def hl_agg_collect_set_union(gene_col: hl.expr.SetExpression) -> hl.expr.SetExpression:
```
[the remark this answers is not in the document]
bpblanken: I think nope!

`collect_as_set` was taking the Set expression column and collecting into a `Set` of `Sets`.  

```
In [1]: import hail as hl

In [2]: t1 = hl.Table.parallelize(
   ...:     [
   ...:         {"genes": hl.set(["abc"])},
   ...:         {"genes": hl.set(["123"])},
   ...:         {"genes": hl.set(["123"])},
   ...:         {"genes": hl.set(["123"])},
   ...:         {"genes": hl.set(["123"])},
   ...:         {"genes": hl.set(["123", "456"])},
   ...:     ],
   ...:     hl.tstruct(genes=hl.dtype('set<str>')),
   ...:     key="genes"
   ...: )

In [3]: t1.aggregate(hl.agg.collect_as_set(t1.genes))
Out[3]: {frozenset({'123'}), frozenset({'abc'}), frozenset({'123', '456'})}
```

There's probably a way to do this with `flatMap` though!

luigi_pipeline/lib/model/gcnv_mt_schema.py
bpblanken: Got it!  I did see the sample specific starts in the genotype but I didn't implement that!  Good catch!

luigi_pipeline/lib/model/gcnv_mt_schema.py
bpblanken: Yep that makes sense!  I've been a bit wary about adding fields that we don't have (want?) in elastic search.  Is that generally ok?  Is `disable_index` useful here?

luigi_pipeline/lib/model/gcnv_mt_schema.py
bpblanken: Yep I agree!  All the existing bucketer functions look like this so I kept as is for now.

luigi_pipeline/lib/model/gcnv_mt_schema.py
```diff
@@ -0,0 +1,212 @@
...
+
+SAMPLE_ID_REGEX = r'(.+)_v\d+_Exome_(C|RP-)\d+$'
+
+def parse_genes(gene_col: hl.expr.StringExpression) -> hl.expr.SetExpression:
+    """
+    Convert a string-ified gene list to a set()
+    """
+    return hl.set(hl.map(
+        lambda gene: gene.split(r'\.')[0],
+        hl.filter(
```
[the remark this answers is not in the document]
bpblanken: Agreed and resolved!

luigi_pipeline/lib/model/gcnv_mt_schema.py
bpblanken: Resolved!

luigi_pipeline/lib/model/gcnv_mt_schema.py
bpblanken: Fixed the `if_else` bit!

`is_latest` has been getting type-handled by hail correctly ("TRUE" => `True` and "FALSE" => `False`) however, so at this point `mt.is_latest` is mapped appropriately.  it might be a bad assumption to let hail do this for us though... I'm not sure!

luigi_pipeline/lib/model/gcnv_mt_schema.py
bpblanken: Hail has been parsing `NA` to the actual hail missing value for us here... though as above... maybe we want to be defensive.

luigi_pipeline/lib/model/gcnv_mt_schema.py
bpblanken: Awesome.  Great suggestion!

luigi_pipeline/seqr_gcnv_loading.py
```diff
@@ -0,0 +1,74 @@
...
+        mt = ht.to_matrix_table(
+            row_key=['variant_name', 'svtype'], col_key=['sample_fix'],
+            # Analagous to CORE_COLUMNS = [CHR_COL, SC_COL, SF_COL, CALL_COL, IN_SILICO_COL] in the old implementation
+            row_fields=['chr', 'vac', 'vaf', 'strvctvre_score'],
+        )
+        # This rename helps disambiguate between the 'start' & 'end' that are aggregations
+        # over samples and the start and end of each sample.
+        mt = mt.rename({'start': 'sample_start', 'end': 'sample_end'})
+
+        if self.remap_path:
```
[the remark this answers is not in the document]
bpblanken: 👍 

My first pass went that direction and I ended up disabling `annotate_old_and_split_multi_hts`, `add_37_coordinates`, and `generate_callstats`.  That makes sense right?

luigi_pipeline/lib/model/gcnv_mt_schema.py
hanars: I'm a bit confused by what would be added to elasticsearch here - we want both the `sorted_transcript_consequences` and  `transcript_consequence_terms` to run and populate in elsaticsearch. The way you have the code now runs these functions in parallel, and I am proposing you run these functions in sequence so that one can be computed from the other to prevent duplicating logic

luigi_pipeline/lib/model/gcnv_mt_schema.py
hanars: Oh I didn't realize hail was imputing the type as a boolean, although that makes sense. We may want to either add some type checking on the matrix table, or actually add some specific typing when we import the table. My concern would be if a new value got added to the input data, we would not throw any error we would just have some undefined behavior. As I think about it, it probably is best to have explicit type definitions or the columns so the import fails if the data does not conform to what we expect

luigi_pipeline/lib/model/gcnv_mt_schema.py
hanars: I didn;tknow it did that, thats cool. I don't think we need to be defensive if so

luigi_pipeline/seqr_gcnv_loading.py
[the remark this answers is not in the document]
hanars: actually `add_37_coordinates` would be nice if it works, I think we have it for genome SVS

luigi_pipeline/lib/model/gcnv_mt_schema.py
bpblanken: Ha.  I was answering a different question than you asked sorry.  My original implementation of this had `hl_agg_collect_set_union(parse_genes(self.mt.genes_LOF_Ensemble_ID)` and `hl_agg_collect_set_union(parse_genes(self.mt.genes_CG_Ensemble_ID))` as subfields just used those on a second round of annotations.  I backed away because I wasn't sure how additional fields are handled upstream.

I'm following what you're saying though and have implemented `transcript_consequence_terms` to depend on `sorted_transcript_consequences`

luigi_pipeline/seqr_gcnv_loading.py
```diff
@@ -0,0 +1,74 @@
...
+    dont_validate = True
+    dataset_type = "SV"
+
+    RUN_VEP = False
+    SCHEMA_CLASS = SeqrGCNVVariantSchema
+
+    # The run function is overriden here to avoid inheriting the default vcf chaining and allele splitting functionality.
+    def run(self):
+        ht = hl.import_table(self.source_paths[0], impute=True, min_partitions=500)
+        mt = ht.to_matrix_table(
```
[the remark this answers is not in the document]
bpblanken: I believe it should be one row... but on my first pass through I thought that keys had to be unique... which I don't think is the case!

Here's an example happy path:

```
In [11]: t1 = hl.Table.parallelize(
    ...:     [
    ...:         {"variant": "1", "sv_type":"2", "sample_fix": "1", "data" : 5},
    ...:         {"variant": "1", "sv_type":"2", "sample_fix": "2", "data" : 6},
    ...:         {"variant": "1", "sv_type":"2", "sample_fix": "3", "data" : 7},
    ...:         {"variant": "1", "sv_type":"2", "sample_fix": "4", "data" : 8},
    ...:     ],
    ...:     hl.tstruct(variant=hl.dtype('str'), sv_type=hl.dtype('str'), sample_fix=hl.dtype('str'), data=hl.dtype('int32')),
    ...: )

In [12]: mt = t1.to_matrix_table(row_key=["variant", "sv_type"], col_key=["sample_fix"])

In [13]: mt.show()
2023-03-01 17:19:01.847 Hail: INFO: Coerced sorted dataset
+---------+---------+----------+----------+----------+----------+
| variant | sv_type | '1'.data | '2'.data | '3'.data | '4'.data |
+---------+---------+----------+----------+----------+----------+
| str     | str     |    int32 |    int32 |    int32 |    int32 |
+---------+---------+----------+----------+----------+----------+
| "1"     | "2"     |        5 |        6 |        7 |        8 |
+---------+---------+----------+----------+----------+----------+

```

and here's a pathological case:

```
In [14]: t1 = hl.Table.parallelize(
    ...:     [
    ...:         {"variant": "1", "sv_type":"2", "sample_fix": "1", "data" : 5},
    ...:         {"variant": "1", "sv_type":"2", "sample_fix": "1", "data" : 6},
    ...:         {"variant": "1", "sv_type":"2", "sample_fix": "1", "data" : 7},
    ...:         {"variant": "1", "sv_type":"2", "sample_fix": "2", "data" : 8},
    ...:     ],
    ...:     hl.tstruct(variant=hl.dtype('str'), sv_type=hl.dtype('str'), sample_fix=hl.dtype('str'), data=hl.dtype('int32')),
    ...: )

In [15]: mt = t1.to_matrix_table(row_key=["variant", "sv_type"], col_key=["sample_fix"])

In [16]: mt.show()
2023-03-01 17:20:13.413 Hail: INFO: Coerced sorted dataset
+---------+---------+----------+----------+
| variant | sv_type | '1'.data | '2'.data |
+---------+---------+----------+----------+
| str     | str     |    int32 |    int32 |
+---------+---------+----------+----------+
| "1"     | "2"     |        5 |        8 |
+---------+---------+----------+----------+

```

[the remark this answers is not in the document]
bpblanken: Honestly have no idea if this is even relevant to your question 🤣.

bpblanken:
@hanars Believe I've addressed everything!

I made a couple of tickets (#399 and #400) for stuff that I thought might be better as their own PRS.

If you approve this I'm going to make a new PR against `master` I think because I've stacked upon non-reviewed stuff 😞.

hanars:
> I've made a couple of tickets (https://github.com/broadinstitute/seqr-loading-pipelines/issues/399 and https://github.com/broadinstitute/seqr-loading-pipelines/issues/400) for stuff that I thought might be better as their own PRS.

So this PR as is is not safe to merge to master, the outstanding comment about type checking is not optional for me. If you create these other ticket and this PR gets approved, we would see that as a sign that this ticket is done and this PR is releasable, despite the fact that its not. I would prefer we get rid of #400 and implement that directly in this PR, or  you could branch off of this branch and do a PR against it if you want that code reviewed separately while keeping it associated with the existing ticket, and then once its merged to this branch we can merge everything to master

For the build 37 work in #399 its up to you if you want a separate ticket for that or not, as its not critical. What I would say though is if you aren't doing that as part of the initial PR for this, that ticket should be put in the backlog and not worked on for a while, its really lower priority than getting the airflow DAGs done for this and other work as well. My comment on it was more if you were going to be sharing functionality between the SV and gCNV tasks it seemed just as easy to share that functionality as to not, but if you think it will actually be its own reviewable chunk of code and effort its probably not worth doing now

Also, I recognize that branching/ticketing strategies can vary, so if you do have any questions or concerns about how we do things or want to discuss offline please let me know

hanars:
> If you approve this I'm going to make a new PR against master I think because I've stacked upon non-reviewed stuff 

Our general approach is to just re-review the same PR multiple times until it is done so I don't mind reading through this all again, and I don't think Mike and Shifa even looked at this yet. It has the advantage of being able to keep the comment history for the whole review process. But if you would feel better about it you can make a new PR and close this one out

luigi_pipeline/seqr_gcnv_loading.py
```diff
@@ -0,0 +1,76 @@
...
+            # Analagous to CORE_COLUMNS = [CHR_COL, SC_COL, SF_COL, CALL_COL, IN_SILICO_COL] in the old implementation
+            row_fields=['chr', 'vac', 'vaf', 'strvctvre_score'],
+        )
+
+        # rename the sample id column before the sample subset happens
+        mt = mt.transmute_cols(s = mt.sample_fix.first_match_in(SAMPLE_ID_REGEX)[0])
+
+        # This rename helps disambiguate between the 'start' & 'end' that are aggregations
+        # over samples and the start and end of each sample.
+        return mt.rename({'start': 'sample_start', 'end': 'sample_end'})
```
ShifaSZ, line 51: I think `start` and `end` are for variants, not samples. But I'm unsure about gCNV.

ShifaSZ (commented):
Some comments at a glance. Maybe more when I finish reviewing.

luigi_pipeline/tests/test_seqr_gcnv_loading.py
```diff
@@ -0,0 +1,386 @@
...
+        genotypes_mt = hl.read_matrix_table(self._genotypes_mt_file)
+        self.assertEqual(genotypes_mt.count(), (2, 3))
+
+        # Now mimic the join in BaseMTToESOptimizedTask
+        genotypes_mt = genotypes_mt.drop(*[k for k in genotypes_mt.globals.keys()])
+        row_ht = genotypes_mt.rows().join(variant_mt.rows()).flatten().drop("variant_name", "svtype")
+        data = row_ht.collect()
+        for i, row in enumerate(data):
+            data[i] = prune_empties(row)
+        self.assertListEqual(data, MERGED_EXPECTED_VARIANT_AND_GENOTYPES_DATA)
```
ShifaSZ, line 386: You may also want to test the `disabled` fields.

luigi_pipeline/tests/test_seqr_sv_loading.py
```diff
@@ -258,14 +258,17 @@ def test_run_task(self, load_gencode_mock):
...
+            source_paths=self._vcf_file,
+            dest_path=self._variant_mt_file,
+            grch38_to_grch37_ref_chain=REFERENCE_CHAIN,
+        ) 
         genotype_task = SeqrSVGenotypesMTTask(
             genome_version="38",
             source_paths="i am completely ignored",
             dest_path=self._genotypes_mt_file
         )
+        SeqrSVGenotypesMTTask.requires = lambda self: [variant_task]
```
ShifaSZ, line 271: It looks nicer. I didn't know it could be done like this.

luigi_pipeline/tests/test_seqr_gcnv_loading.py
```diff
@@ -0,0 +1,386 @@
...
+        genotypes_mt = hl.read_matrix_table(self._genotypes_mt_file)
+        self.assertEqual(genotypes_mt.count(), (3, 4))
+
+        # Now mimic the join in BaseMTToESOptimizedTask
+        genotypes_mt = genotypes_mt.drop(*[k for k in genotypes_mt.globals.keys()])
+        row_ht = genotypes_mt.rows().join(variant_mt.rows()).flatten().drop("variant_name", "svtype")
+        data = row_ht.collect()
+        for i, row in enumerate(data):
+            data[i] = prune_empties(row)
+        self.assertListEqual(data, NEW_JOINT_CALLED_EXPECTED_VARIANT_AND_GENOTYPES_DATA)
```
ShifaSZ, line 349: I'm curious if it is better to mock the `HailElasticsearchClient` and test the parameters for calling `export_table_to_elasticsearch`.

ShifaSZ (changes requested):

luigi_pipeline/lib/model/seqr_mt_schema.py
```diff
@@ -344,7 +342,9 @@ def elasticsearch_row(ds):
         table = ds.drop('vep').flatten()
         # When flattening, the table is unkeyed, which causes problems because our locus and alleles should not
         # be normal fields. We can also re-key, but I believe this is computational?
-        table = table.drop(table.locus, table.alleles)
-
+        #
+        # PS: locus and alleles used to be the only possible row keys... however we're extending to drop any row key!
+        row_key = table.row_key
+        table = table.drop(*row_key)
```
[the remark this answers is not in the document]
bpblanken: The keys here are `row_key=['variant_name', 'svtype']`, which is set explicitly during data load on [this step](https://github.com/broadinstitute/seqr-loading-pipelines/pull/398/files#diff-e8711afd284aabadd3dcdc1e1fbbfc1c96dc461aff255765515602a0515529f3R41.) 

We won't actually need them because we've re-annotated to `variantId` and `svType` in [this block](https://github.com/broadinstitute/seqr-loading-pipelines/pull/398/files#diff-d6fdf9e5cbf606552685459442b0989ded7b0f8e4ad7421ef720e73ff69aeca5R60-R69)

luigi_pipeline/seqr_gcnv_loading.py
bpblanken: 👍 

yep!  I think the situation here is we're getting a tsv of values where each row corresponds to the `start` and `end` of a variant & sample combination.  We then aggregate over the samples to get the variant `start` and `end` when we do the row annotations:

```
    @row_annotation()
    def start(self):
        return hl.agg.min(self.mt.sample_start)

    @row_annotation()
    def end(self):
        return hl.agg.max(self.mt.sample_end)
```

I was initially fine leaving the `start` and `end` named what they were, but hail was very unhappy with something like this:

```
    @row_annotation()
    def start(self):
        return hl.agg.min(self.mt.start)

    @row_annotation()
    def end(self):
        return hl.agg.max(self.mt.end)
```
because the field names already exist.

luigi_pipeline/lib/model/seqr_mt_schema.py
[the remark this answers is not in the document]
ShifaSZ: Gotcha. There is a comment in the function docstring about the dropping locus and alleles key. It's better to update them.

luigi_pipeline/lib/model/gcnv_mt_schema.py
```diff
@@ -0,0 +1,213 @@
...
+    return hl.fold(
+        lambda i, j: i | j,
+        hl.empty_set(hl.tstr),
+        hl.agg.collect(gene_col),
+    )
+
+class SeqrGCNVVariantSchema(BaseMTSchema):
+
+    @row_annotation()
+    def contig(self):
```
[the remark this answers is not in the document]
bpblanken: Hrm... I think the implementations are actually pretty different here... `start` and `pos` are aggregations and we don't have an underlying `locus` to derive the other fields from.  We could maybe pre-annotate with `locus` to make some of them work... but that might be more effort than it's worth (`BaseVariantSchema` doesn't have any `fn_requires`).

luigi_pipeline/lib/model/gcnv_mt_schema.py
```diff
@@ -0,0 +1,213 @@
...
+    @row_annotation(name='transcriptConsequenceTerms', fn_require=[
+        sv_type,
+        sorted_transcript_consequences,
+    ])
+    def transcript_consequence_terms(self):
+        default_consequences = hl.set([hl.format('gCNV_%s', self.mt.svType)])
+        gene_major_consequences = hl.set(self.mt.sortedTranscriptConsequences.map(lambda x: x[MAJOR_CONSEQUENCE]))
+        return hl.array(
+            default_consequences.union(gene_major_consequences)
+        )
```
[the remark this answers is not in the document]
bpblanken: 👍 this function is modified on an upstream pr so will leave it here and address on the modified implementation.

luigi_pipeline/lib/model/gcnv_mt_schema.py
```diff
@@ -0,0 +1,213 @@
...
+            '%i_to_%i' % (i, i+step): self._genotype_filter_samples(lambda g: ((g.qs >= i) & (g.qs < i+step)))
+            for i in range(start, end, step)
+        }, **{
+            "samples_qs_gt_1000": self._genotype_filter_samples(lambda g: g.qs >= 1000)
+        })
+
+    @row_annotation(name="samples_cn", fn_require=SeqrGenotypesSchema.genotypes)
+    def samples_cn(self, start=0, end=4, step=1):
+        return hl.struct(**{
+            '%i' % i: self._genotype_filter_samples(lambda g: g.cn == i)
```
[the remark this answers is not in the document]
bpblanken: we don't use `f-strings` on any of these bucketing functions... though I agree it makes sense.  We should probably address all at once?

luigi_pipeline/lib/model/seqr_mt_schema.py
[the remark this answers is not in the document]
bpblanken: Agreed!

luigi_pipeline/seqr_gcnv_loading.py
```diff
@@ -0,0 +1,76 @@
...
+
+    def generate_callstats(self, mt, *args, **kwargs):
+        return mt
+
+    def import_dataset(self):
+        ht = hl.import_table(self.source_paths[0], impute=True, min_partitions=500)
+        mt = ht.to_matrix_table(
+            row_key=['variant_name', 'svtype'], col_key=['sample_fix'],
+            # Analagous to CORE_COLUMNS = [CHR_COL, SC_COL, SF_COL, CALL_COL, IN_SILICO_COL] in the old implementation
+            row_fields=['chr', 'vac', 'vaf', 'strvctvre_score'],
```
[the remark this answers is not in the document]
bpblanken: Yep... the idea here is that every other field becomes an entry field.  These are the fields that are unique to a variant.

luigi_pipeline/lib/model/gcnv_mt_schema.py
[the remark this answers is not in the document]
bpblanken: I tweaked `parse_genes` to return an array and this to use `flatten()`, which I think makes this more readable!

luigi_pipeline/lib/model/gcnv_mt_schema.py
```diff
@@ -0,0 +1,210 @@
...
+    )
+
+def hl_agg_collect_set_union(gene_col: hl.expr.SetExpression) -> hl.expr.SetExpression:
+    """
+    aggregate with the set union operator
+    """
+
+    return hl.set(
+        hl.flatten(
+            hl.agg.collect(gene_col)
```
hanars, line 31: would ` hl.flatten(hl.agg.collect_as_set` work?

luigi_pipeline/lib/model/gcnv_mt_schema.py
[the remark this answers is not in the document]
hanars: But you could still overwrite `start` and `pos` while inheriting `contig`. I agree that pre-annotating locus is more effort than it is worth. Or at a minimum define a helper function for this expression so its not copy-pasted and duplicated

luigi_pipeline/lib/model/seqr_mt_schema.py
```diff
@@ -348,9 +346,8 @@ def elasticsearch_row(ds):
...
         # Converts nested structs into one field, e.g. {a: {b: 1}} => a.b: 1
         table = ds.drop('vep').flatten()
-        # When flattening, the table is unkeyed, which causes problems because our locus and alleles should not
+        # When flattening, the table is unkeyed, which causes problems because our row keys should not
         # be normal fields. We can also re-key, but I believe this is computational?
-        table = table.drop(table.locus, table.alleles)
-
-
-        return table
+        # PS: locus and alleles used to be the only possible row keys... however we're extending to drop any row key!
```
hanars, line 351: I don't love comments that will live in perpetuity that say things like "this is what the behavior used to be". As a new user to a repo usually find it really confusing. Maybe a comment like "row key is often locus and allele, but does not have to be"

luigi_pipeline/seqr_gcnv_loading.py
```diff
@@ -0,0 +1,76 @@
...
+    dataset_type = "SV"
+
+    RUN_VEP = False
+    SCHEMA_CLASS = SeqrGCNVVariantSchema
+
+    def annotate_old_and_split_multi_hts(self, mt, *args, **kwargs):
+        return mt
+
+    def add_37_coordinates(self, mt, *args, **kwargs):
+        # TODO.. implement this (it was out of scope in the port but we want it eventually)
```
hanars, line 32: did we make a separate ticket for this? Just want to make sure it doesn't get forgotten

luigi_pipeline/lib/model/gcnv_mt_schema.py
bpblanken: As is... `hl.flatten(hl.agg.collect_as_set` throws an error for me (I'm not fully sure why...)

I think we would actually still need the outer `hl.set` as well as an inner `hl.array` cast to make that work.  

For example:

```
t1 = hl.Table.parallelize(
    ...:      [
    ...:          {"variant": ["1", "2"], "sv_type":"2", "sample_fix": "1", "data" : 5},
    ...:          {"variant": ["1", "4"], "sv_type":"2", "sample_fix": "2", "data" : 6},
    ...:          {"variant": ["2", "5", "4"], "sv_type":"2", "sample_fix": "3", "data" : 7},
    ...:          {"variant": ["1", "3", "4"], "sv_type":"2", "sample_fix": "4", "data" : 8},
    ...:      ],
    ...:      hl.tstruct(variant=hl.dtype('array<str>'), sv_type=hl.dtype('str'), sample_fix=hl.dtype('str'), data=hl.dtype('int32')),
    ...:  )
```

```
In [14]: t1.aggregate(hl.flatten(hl.array(hl.agg.collect_as_set(t1.variant))))
Out[14]: ['1', '2', '1', '3', '4', '1', '4', '2', '5', '4']

In [15]: t1.aggregate(hl.set(hl.flatten(hl.array(hl.agg.collect_as_set(t1.variant)))))
Out[15]: {'1', '2', '3', '4', '5'}
```

versus

```
In [12]: t1.aggregate(hl.flatten(hl.agg.collect(t1.variant)))
Out[12]: ['1', '2', '1', '4', '2', '5', '4', '1', '3', '4']
In [12]: t1.aggregate(hl.flatten(hl.agg.collect(t1.variant)))
Out[12]: ['1', '2', '1', '4', '2', '5', '4', '1', '3', '4']

In [13]: t1.aggregate(hl.set(hl.flatten(hl.agg.collect(t1.variant))))
Out[13]: {'1', '2', '3', '4', '5'}
```

luigi_pipeline/lib/model/seqr_mt_schema.py
bpblanken: 👍

luigi_pipeline/seqr_gcnv_loading.py
bpblanken: yep! https://github.com/broadinstitute/seqr-loading-pipelines/issues/399

luigi_pipeline/lib/model/gcnv_mt_schema.py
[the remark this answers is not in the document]
bpblanken: I made a ticket: https://github.com/broadinstitute/seqr-loading-pipelines/issues/403.  We can just do them all at once!

luigi_pipeline/lib/model/gcnv_mt_schema.py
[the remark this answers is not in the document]
bpblanken: Do you mean writing a helper function to do `.replace("^chr", "")`?

luigi_pipeline/tests/test_seqr_gcnv_loading.py
```diff
@@ -0,0 +1,384 @@
+import datetime
+import shutil
+import tempfile
+import unittest
+from unittest import mock
+
+import hail as hl
+import luigi.worker
+
+from seqr_gcnv_loading import SeqrGCNVVariantMTTask, SeqrGCNVGenotypesMTTask, SeqrGCNVMTToESTask
```
ShifaSZ, line 10: The `SeqrGCNVMTToESTask` is unused.

luigi_pipeline/tests/test_seqr_gcnv_loading.py
ShifaSZ: You don't test the `SeqrGCNVMTToESTask`. That's why the [disabled fields](https://github.com/broadinstitute/seqr-loading-pipelines/blob/c794aff1f9f54094416db16b6c4fb92e328d7d3b/luigi_pipeline/seqr_loading_optimized.py#L77-L78) can't be tested.

luigi_pipeline/lib/model/gcnv_mt_schema.py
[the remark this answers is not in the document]
bpblanken: Should be fixed!

luigi_pipeline/tests/test_seqr_gcnv_loading.py
bpblanken: Yep, I'm going to take a shot at doing what you suggested regarding mocking closer to elastic search.  I initially left testing the elastic search export as out of scope because I don't really touch anything there too closely, but there's definitely major holes in these tests if we don't test `SeqrGCNVMTToESTask`.

hanars: I will say, we are going to be sunsetting elasticsearch over the next few months. So while it is probably be a good idea to update the mocking to better test some elasticsearch behavior such as disable_index fields, I would recommend you not put too much time into it

bpblanken: ☝️ This was my initial thought.  I _think_ it's just a couple of lines.  If it takes more than that I'll reverse course.

bpblanken: Got this working!  It was a _little_ more painful than expected (two tests pointing at the same mock 🤦‍♂️) but was definitely a worthy exercise since it found two nasty bugs.

luigi_pipeline/tests/test_seqr_gcnv_loading.py
```diff
@@ -0,0 +1,394 @@
+import datetime
+import shutil
+import tempfile
+import unittest
+from unittest import mock
+
+import hail as hl
+import luigi.worker
+
+from hail_scripts.elasticsearch.hail_elasticsearch_client import HailElasticsearchClient
```
ShifaSZ, line 10: You don't need this line.

luigi_pipeline/tests/test_seqr_gcnv_loading.py
```diff
@@ -0,0 +1,394 @@
...
+            dest_path=self._variant_mt_file,
+        )
+        genotype_task = SeqrGCNVGenotypesMTTask(
+            genome_version="38",
+            source_paths="i am completely ignored",
+            dest_path=self._genotypes_mt_file,
+            is_new_joint_call=True,
+        )
+        export_task = SeqrGCNVMTToESTask()
+        export_task._es.reset_mock()
```
ShifaSZ, line 331: I know export_task._es has been used when calling `SeqrGCNVMTToESTask()` above. But is it necessary to do `reset_mock()` here? To me, you can remove this line.

luigi_pipeline/tests/test_seqr_gcnv_loading.py
```diff
@@ -0,0 +1,394 @@
...
+        genotypes_mt = hl.read_matrix_table(self._genotypes_mt_file)
+        self.assertEqual(genotypes_mt.count(), (3, 4))
+
+        export_task._es.export_table_to_elasticsearch.assert_called_once()
+        args, kwargs = export_task._es.export_table_to_elasticsearch.call_args
+        row_ht = args[0].collect()
+        for i, row in enumerate(row_ht):
+            row_ht[i] = prune_empties(row)
+        self.assertListEqual(row_ht, NEW_JOINT_CALLED_EXPECTED_VARIANT_AND_GENOTYPES_DATA)
+        self.assertListEqual(kwargs["disable_index_for_fields"], EXPECTED_DISABLED_INDEX_FIELDS)
```
ShifaSZ, line 354: I like these better than those before.

ShifaSZ (changes requested):

luigi_pipeline/lib/model/gcnv_mt_schema.py
hanars: its weird that its throwing an error for you - the documentation for `hl.flatten` explicitly states that  it can accept either a SetExpression or an ArrayExpression

hanars: > this can work if I go back to having the parsed genes be sets instead of an array.

I think thats better in some cases, and you can always cast it back to an array if you really need to in one or two places

luigi_pipeline/tests/test_seqr_gcnv_loading.py
bpblanken: I fixed this!  The source of the bug was luigi re-using instances of `SeqrGCNVMTToESTask`.  I didn't know that task instances with the same parameters would actually be the same in-memory instance.  I disabled the cache in the tests which forces disjoint mock objects in the different tests.

luigi_pipeline/lib/model/gcnv_mt_schema.py
bpblanken: 👍 and fixed!

hanars (approved):
Looks good to me! Only comment I would have is that this PR is branched off a branch off a branch and it might be easiest if you merge the branches its branched on and then merge this directly to master, instead of having this merged to a massive branch

luigi_pipeline/lib/model/gcnv_mt_schema.py
```diff
@@ -0,0 +1,202 @@
...
+
+from lib.model.base_mt_schema import row_annotation
+from lib.model.seqr_mt_schema import BaseVariantSchema, SeqrGenotypesSchema, SeqrVariantsAndGenotypesSchema
+
+from hail_scripts.computed_fields import variant_id
+
+GENE_ID = "gene_id"
+MAJOR_CONSEQUENCE = "major_consequence"
+
+def parse_genes(gene_col: hl.expr.StringExpression) -> hl.expr.SetExpression:
```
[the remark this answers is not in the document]
mike-w-wilson: ^^^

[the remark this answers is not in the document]
bpblanken: 👍 Yep, I think we can get there... gradually!

PR #401, opened 2023-03-02
Type check the gcnv columns.

bpblanken:


luigi_pipeline/seqr_gcnv_loading.py
```diff
@@ -25,6 +76,16 @@ class SeqrGCNVVariantMTTask(SeqrVCFToVariantMTTask):
     RUN_VEP = False
     SCHEMA_CLASS = SeqrGCNVVariantSchema
 
+    is_new_joint_call = luigi.BoolParameter(default=False, description='Is this a fully joint-called callset.')
+
+    @staticmethod
+    def validate_ht_types(ht, expected_types=EXPECTED_TYPES):
```
bpblanken, line 82: This throws if there are extra columns or incorrectly typed columns but not if there are missing columns.  If those arise, the pipeline will fail anyway.  Might be better to be explicit here though?

luigi_pipeline/seqr_gcnv_loading.py
```diff
@@ -13,6 +14,56 @@
...
+    "HIGH_QUALITY": hl.tbool, 
+    "genes_any_overlap": hl.tstr,
+    "genes_any_overlap_exonsPerGene": hl.tstr, 
+    "genes_any_overlap_totalExons": hl.tint32, 
+    "genes_strict_overlap": hl.tstr,
+    "genes_strict_overlap_exonsPerGene": hl.tint32,
+    "genes_strict_overlap_totalExons": hl.tint32,
+    "genes_CG": hl.tstr, 
+    "genes_LOF": hl.tstr, 
+    "genes_any_overlap_Ensemble_ID": hl.tstr,
```
bpblanken, line 54: Many of these are columns that are unused upstream. Maybe not worth throwing if an unused column is mis-typed?

luigi_pipeline/seqr_gcnv_loading.py
```diff
@@ -37,6 +98,7 @@ def generate_callstats(self, mt, *args, **kwargs):
 
     def import_dataset(self):
         ht = hl.import_table(self.source_paths[0], impute=True, min_partitions=500)
```
hanars, line 100: I think it would be better to pass `types` in here instead of imputing types and then validating. For anything you don't specify, hail will default them to strings, which for fields we never use is acceptable behavior so you will only need to pass in the fields we care about

hanars (changes requested):

luigi_pipeline/seqr_gcnv_loading.py
bpblanken: 👍 makes sense!

bpblanken: From what I'm seeing hail does the validation for us.  If hail can't parse as the type specified it'll throw an exception.

luigi_pipeline/seqr_gcnv_loading.py
```diff
@@ -36,7 +60,7 @@ def generate_callstats(self, mt, *args, **kwargs):
         return mt
 
     def import_dataset(self):
-        ht = hl.import_table(self.source_paths[0], impute=True, min_partitions=500)
+        ht = hl.import_table(self.source_paths[0], types=EXPECTED_TYPES, min_partitions=500)
```
bpblanken, line 63: Validation happens lazily here, which I think is fine?  There's a very clear error message:

```
HailUserError: Error summary: HailException: cannot parse boolean from input string 't'
```

hanars: most things in hail are lazy, so thats fine with me

luigi_pipeline/seqr_gcnv_loading.py
hanars: I think things default ot strings, so its also okay to leave out all the string columns from this mapping too

bpblanken: 👍

luigi_pipeline/seqr_gcnv_loading.py
```diff
@@ -13,6 +13,21 @@
...
+    "CN": hl.tint32,
+    "QS": hl.tint32, 
+    "defragmented": hl.tbool, 
+    "vaf": hl.tfloat64, 
+    "vac": hl.tint32,
+    "genes_any_overlap_totalExons": hl.tint32,
+    "genes_strict_overlap_totalExons": hl.tint32,
+    "no_ovl": hl.tbool, 
+    "strvctvre_score": hl.tfloat64,
+    "is_latest": hl.tbool
```
hanars, line 28: If you pass in array fields for the gene list fields, will hail properly parse those as lists or no?

luigi_pipeline/seqr_gcnv_loading.py
```diff
@@ -13,6 +13,21 @@
...
+    "start": hl.tint32, 
+    "end": hl.tint32, 
+    "CN": hl.tint32,
+    "QS": hl.tint32, 
+    "defragmented": hl.tbool, 
+    "vaf": hl.tfloat64, 
+    "vac": hl.tint32,
+    "genes_any_overlap_totalExons": hl.tint32,
+    "genes_strict_overlap_totalExons": hl.tint32,
+    "no_ovl": hl.tbool,
```
hanars, line 26: what about `identical_ovl` and `any_ovl`

hanars (changes requested):

luigi_pipeline/seqr_gcnv_loading.py
```diff
@@ -13,6 +13,21 @@
 
 SAMPLE_ID_REGEX = r'(.+)_v\d+_Exome_(C|RP-)\d+$'
 
+EXPECTED_TYPES = {
```
ShifaSZ, line 16: The `type` parameter of the `import_table` is a field type assignment. So you may want to change it to something like `FIELD_TYPES`.

luigi_pipeline/seqr_gcnv_loading.py
bpblanken: Looks like no... but almost!

```
Error summary: JsonParseException: Unrecognized token 'ENSG00000283761': was expecting (JSON String, Number, Array, Object or token 'null', 'true' or 'false')
 at [Source: UNKNOWN; line: 1, column: 16]
```

I imagine if the fields were proper JSON arrays rather than csv strings we'd be fine!  This was a super idea though.

luigi_pipeline/seqr_gcnv_loading.py
hanars: no that is right, I just forgot

luigi_pipeline/seqr_gcnv_loading.py
hanars: oh well, worth a try

hanars (approved):

PR #404, opened 2023-03-14
Prefer f-string to c-style

### LINKED ISSUE #403: Update genotype bucketing functions to use `f-string`
They're using `%` styling, and `f-string` is preferred.

bpblanken:
Resolves #403

### DIFF
--- a/luigi_pipeline/lib/model/gcnv_mt_schema.py
+++ b/luigi_pipeline/lib/model/gcnv_mt_schema.py
@@ -164,7 +164,7 @@ class SeqrGCNVGenotypesSchema(SeqrGenotypesSchema):
     @row_annotation(fn_require=SeqrGenotypesSchema.genotypes)
     def samples_qs(self, start=0, end=1000, step=10):
         return hl.struct(**{
-            '%i_to_%i' % (i, i+step): self._genotype_filter_samples(lambda g: ((g.qs >= i) & (g.qs < i+step)))
+            f'{i}_to_{i + step}': self._genotype_filter_samples(lambda g: ((g.qs >= i) & (g.qs < i+step)))
             for i in range(start, end, step)
         }, **{
             "samples_qs_gt_1000": self._genotype_filter_samples(lambda g: g.qs >= 1000)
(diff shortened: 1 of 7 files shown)


luigi_pipeline/lib/model/gcnv_mt_schema.py
```diff
@@ -173,7 +173,7 @@ def samples_qs(self, start=0, end=1000, step=10):
     @row_annotation(name="samples_cn", fn_require=SeqrGenotypesSchema.genotypes)
     def samples_cn(self, start=0, end=4, step=1):
         return hl.struct(**{
-            '%i' % i: self._genotype_filter_samples(lambda g: g.cn == i)
+            f'{i}_to_{i + step}': self._genotype_filter_samples(lambda g: g.cn == i)
```
hanars, line 176: this should just be `f'{i}'` I think

hanars (changes requested):

luigi_pipeline/lib/model/seqr_mt_schema.py
```diff
@@ -278,23 +278,23 @@ def samples_no_call(self):
     @row_annotation(fn_require=genotypes)
     def samples_num_alt(self, start=1, end=3, step=1):
         return hl.struct(**{
-            '%i' % i: self._genotype_filter_samples(lambda g: g.num_alt == i)
+            f'{i}': self._genotype_filter_samples(lambda g: g.num_alt == i)
```
bpblanken, line 281: oof and this one also flipped.

bpblanken: oh never mind

luigi_pipeline/lib/model/gcnv_mt_schema.py
bpblanken: this is what I get for not running the tests :(. This is even covered!

ShifaSZ:
The changes look good. You may also want to change the logger string formats in these files. The test files also contain c-string formats which need to be updated.

ShifaSZ (approved):

PR #405, opened 2023-03-15
Move `load_gencode` to `hail_utils` and `rm` the existing pipeline.  

bpblanken:
`hail_scripts` has been consistently deployed into the `pyscripts.zip` folder for airflow, so this seemed like a reasonable place to put the gencode download.

The tests in this directory still pass, which is nice!

```
bblanken@wmbbe-67c seqr-loading-pipelines % nosetests hail_scripts/utils
...
Ran 3 tests in 0.013s

OK
```


hail_scripts/utils/download_utils.py
```diff
@@ -1,14 +1,32 @@
...
-from contextlib import contextmanager
-import hail as hl
 
-from sv_pipeline.utils.common import parse_gs_path_to_bucket
+import hail as hl
+from google.cloud import storage
 
 logger = logging.getLogger(__name__)
 
+def parse_gs_path_to_bucket(gs_path):
```
bpblanken, line 12: I opted to copy a couple functions that were used from the `sv_pipeline.utils.common` module, which I believe can be deleted after the new SV and GCNV pipelines are merged?

sv_pipeline/genome/load_data.py
```diff
@@ -1,277 +0,0 @@
-# A simple usage guide:
```
bpblanken, line 1: I might be being aggressive/presumptuous with removing these, but I don't think we'll be going backwards at this point?

ShifaSZ (approved):
LGTM

PR #408, opened 2023-03-17
Add `isort` config and run it on `luigi_pipeline`

bpblanken:


luigi_pipeline/lib/model/base_mt_schema.py
```diff
@@ -1,8 +1,7 @@
+from collections import defaultdict
```
bpblanken, line 1: There's a configuration option (`force_sort_within_sections`) that enables full alphabetical rather than splitting the `import x` imports from the `from an import b` imports.  With the option disabled (the default), this block would be:
```
import logging
from collections import defaultdict
from inspect import getmembers
from typing import List
```

rather than what it is here.  

I'm extremely unopinionated about which option is better, though I was slightly surprised by the default!  Does anyone else have thoughts?

luigi_pipeline/pyproject.toml
```diff
@@ -0,0 +1,6 @@
+[tool.isort]
+force_sort_within_sections = "True"
+include_trailing_comma = "True"
+known_external_package = ['hail_scripts']
```
bpblanken, line 4: This is a mini-hack to lift `hail_scripts` into its own section of imports.  By default, `isort` is treating it as a `third party`.  Setting `hail_scripts` as a `first party` module seemed a bit off since those lines were getting interspersed with `lib` and `seqr_loading`.

luigi_pipeline/lib/model/base_mt_schema.py
ShifaSZ: I didn't know much about the `isort` before. Both options look good to me.

ShifaSZ (approved):
I don't see any issue with that.

luigi_pipeline/lib/model/base_mt_schema.py
bpblanken: 👍 Slight preference > no preference!  Also... I was looking at a block elsewhere that was:

```
import a
from b import c
import d
from e import f
import h
```

and thought it looked a bit gross.  So I think I agree!

bpblanken:
Accidentally closed this by deleting the underlying branch, which should have already been deleted.

PR #414, opened 2023-03-22
Update deployment process.

bpblanken:
Idea here is to run a script rather than have a few manual steps.  Long term we can do something else?


pyproject.toml
```diff
@@ -0,0 +1,8 @@
+[project]
+name = "seqr_luigi"
+requires-python = ">=3.8"
+version = "1.9.0"
+
+[tool.setuptools]
+packages = ["luigi_pipeline.lib", "hail_scripts"]
```
bpblanken, line 7: Idea here is to package our code using a standard packaging tool (`setuptools`) rather than zipping manually with a curated list of files.  

The current `pyscripts.zip` has this structure:
```
lib/
hail_scripts/
seqr_loading.py
```
which was very tricky to 1-to-1 duplicate because it's a bit different from what we have in the repo.  

There's a nice utility for lifting a nested package up a level (`package-dir` = `{"lib": "luigi_pipeline/lib"}`), but the same utility doesn't exist for single modules like `seqr_loading` and `seqr_loading_optimized` (which we need because we're also using it as a library module in mito/sv/gcnv).

Another option is to lift `luigi_pipeline` itself up a level (`package-dir` = `{"": "lib"}`) which does work... but `hail_scripts` stopped working... I think because the empty string package cannot be combined with other packages. 

The clearest path forwards for me was to just make `luigi_pipeline` a package and make all of our `import lib` statements absolute by making them `import luigi_pipeline.lib`.  We already had that sneaking in (in the `mito` and `sv` loading scripts) in ways that would break in the airflow deployment.

bin/deploy.py
```diff
@@ -0,0 +1,74 @@
...
+        help="Relative path to the source directory containing executable mains",
+    )
+    parser.add_argument(
+        "--gcs-project",
+        default="seqr-project",
+        help="Name of the GCS project",
+    )
+    parser.add_argument(
+        "--gcs-bucket-name",
+        default="seqr-luigi",
```
bpblanken, line 60: Open question for me about whether this script even belongs inside this repo (because it's public) and this process is specific to our own deployment.

mike-w-wilson: We could make a `seqr-loading-pipelines-private` private repo? Mimicking what seqr did?

PR #427, opened 2023-03-28
Implement SV Concordance

### LINKED ISSUE #424: Add prev call and genotype annotations to SV loading
From original seqr ticket https://github.com/broadinstitute/seqr/issues/3294 :

> In addition to SV data for samples on which we did not previously have SV calls, the v3 callset includes many samples included in v2, which was previously loaded to seqr. For the samples included in both v2 and v3, we anticipate ~10% of calls to be discordant. We'd like to be able to flag a subset of these discordant calls to make them easier for the analysts to review.

> Variants that were NOT called in the sample in v2 but are called in v3 - These would be tagged as New calls.
Variants that were called in the sample in both v2 and v3, but genotype differs (eg het vs hom) - These would be tagged as New genotype. It would be great if hovering over this tag could display the quality score and genotype from the previous callset.
For filtering behavior, we'd like filtering for "New" calls to return both of the categories listed above.

We will be adding `prev_call` and `new_call` fields to the genotypes to correspond with the gCNV fields, and adding a new `prev_num_alt` field derived from the concordance code. We will also add a `samples_new_call` field similar to what we have for gCNV

bpblanken:
I know we're not all the way there yet on deciding the business logic, so just putting this up as a placeholder.
Resolves #424

EDIT: adding reviewers now!


luigi_pipeline/lib/model/sv_mt_schema.py
```diff
@@ -196,17 +211,37 @@ class SeqrSVGenotypesSchema(SeqrGenotypesSchema):
...
+        novel_genotype = (num_alt != prev_num_alt) & (prev_num_alt == 0)
         return {
             'sample_id': self.mt.s,
             'gq': self.mt.GQ,
             'cn': self.mt.RD_CN,
-            'num_alt': self._num_alt(is_called)
+            'num_alt': num_alt,
+            'prev_num_alt': prev_num_alt,
+            'new_genotype': hl.or_missing(is_called, discordant_genotype),
+            'new_call': hl.or_missing(is_called, ~was_previously_called | novel_genotype),
```
bpblanken, line 230: This is a implementation of this truth table:

image

bpblanken: whether that's correct or not is up in the air.

luigi_pipeline/lib/model/sv_mt_schema.py
```diff
@@ -196,17 +211,37 @@ class SeqrSVGenotypesSchema(SeqrGenotypesSchema):
 
     def _genotype_fields(self):
         is_called = hl.is_defined(self.mt.GT)
+        was_previously_called = hl.is_defined(self.mt.CONC_ST) & ~self.mt.CONC_ST.contains("EMPTY")
```
bpblanken, line 214: Occasionally `CONC_ST` is a populated array with the single value of `EMPTY`.  I asked Alba about these.

luigi_pipeline/lib/model/sv_mt_schema.py
```diff
@@ -19,9 +19,24 @@
 CONSEQ_PREDICTED_PREFIX = 'PREDICTED_'
 NON_GENE_PREDICTIONS = {'PREDICTED_INTERGENIC', 'PREDICTED_NONCODING_BREAKPOINT', 'PREDICTED_NONCODING_SPAN'}
 
+PREVIOUS_GENOTYPE_N_ALT_ALLELES = hl.dict({
+    # Map of concordance string -> previous n_alt_alleles()
 
-INTERVAL_TYPE = 'array<struct{type: str, chrom: str, start: int32, end: int32}>'
+    # Concordant
+    frozenset(["TN"]): 0,       # 0/0 -> 0/0
```
bpblanken, line 26: I opted for a `frozenset` here because it's not clear to me whether the array of classifications is guaranteed to be ordered.  In practice, I think it _is_ ordered... the VCF always has `TP,TN` and never `TN,TP`.  In the slideshow though... it's the opposite:

Screenshot 2023-03-29 at 3 12 24 PM

luigi_pipeline/lib/model/sv_mt_schema.py
bpblanken: Alba confirmed that `EMPTY` means there's no overlap with the previous callset.

luigi_pipeline/lib/model/sv_mt_schema.py
```diff
@@ -196,17 +211,37 @@ class SeqrSVGenotypesSchema(SeqrGenotypesSchema):
...
         samples = super()._genotype_filter_samples(filter_)
-        return hl.or_missing(hl.len(samples) > 0, samples)            
+        return hl.or_missing(hl.len(samples) > 0, samples)
+
+    @row_annotation(fn_require=SeqrGenotypesSchema.genotypes)
+    def samples_new_call(self):
+        return self._genotype_filter_samples(lambda g: g.new_call)
+
+    @row_annotation(fn_require=SeqrGenotypesSchema.genotypes)
+    def samples_new_genotype(self):
```
hanars, line 243: So looking into the desired behavior from the analysts, they want if they use the existing filter in seqr for "new calls" it to return variants that are either `new_call` or `new_genotype`. So we actually just want the `samples_new_call` field to include anything with either of those, instead of adding a new field

luigi_pipeline/lib/model/sv_mt_schema.py
hanars: This looks right to me, but what does "Absent Genotype" refer to here? The concordance codes all seem to map to Concordant, Novel, and Discordant

hanars (changes requested):

luigi_pipeline/lib/model/sv_mt_schema.py
```diff
@@ -19,9 +19,24 @@
 CONSEQ_PREDICTED_PREFIX = 'PREDICTED_'
 NON_GENE_PREDICTIONS = {'PREDICTED_INTERGENIC', 'PREDICTED_NONCODING_BREAKPOINT', 'PREDICTED_NONCODING_SPAN'}
 
+PREVIOUS_GENOTYPE_N_ALT_ALLELES = hl.dict({
+    # Map of concordance string -> previous n_alt_alleles()
 
-INTERVAL_TYPE = 'array<struct{type: str, chrom: str, start: int32, end: int32}>'
```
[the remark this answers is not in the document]
bpblanken: Unused... I think it was hanging out from when we were building missing values explicitly for use in `hl.if_else` rather than using `hl.or_missing`.

luigi_pipeline/lib/model/sv_mt_schema.py
bpblanken: Absent Genotype refers to these two:

```
Absent genotype:
0/1	0/0	TN,FN
1/1	0/0	FN
```

in code I didn't think it made sense to split absent and discordant into two categories since the consequent behavior would be the same.

hanars: Makes sense. Do you mind updating the `PREVIOUS_GENOTYPE_N_ALT_ALLELES`, I like that you have it all grouped by category there but this category is missing

luigi_pipeline/lib/model/sv_mt_schema.py
```diff
@@ -196,17 +213,28 @@ class SeqrSVGenotypesSchema(SeqrGenotypesSchema):
 
     def _genotype_fields(self):
         is_called = hl.is_defined(self.mt.GT)
+        was_previously_called = hl.is_defined(self.mt.CONC_ST) & ~self.mt.CONC_ST.contains("EMPTY")
+        num_alt = self._num_alt(is_called)
+        prev_num_alt = PREVIOUS_GENOTYPE_N_ALT_ALLELES.get(hl.set(self.mt.CONC_ST), -1)
```
bpblanken, line 218: I fixed the issues I was having by lifting this out of an `hl.if_else`.  I think it's cleaner anyways... though maybe a bit implicit that we allow ourselves to look up the `hl.set` missing value.

luigi_pipeline/seqr_loading_optimized.py
```diff
@@ -22,6 +22,12 @@ class SeqrVCFToVariantMTTask(SeqrVCFToMTTask):
     """
     SCHEMA_CLASS = SeqrVariantSchema
 
+    def run(self):
```
bpblanken, line 25: hm.. what is this doing here?

luigi_pipeline/lib/model/sv_mt_schema.py
```diff
@@ -196,17 +213,28 @@ class SeqrSVGenotypesSchema(SeqrGenotypesSchema):
...
+        discordant_genotype = (num_alt != prev_num_alt) & (prev_num_alt > 0)
+        novel_genotype = (num_alt != prev_num_alt) & (prev_num_alt == 0)
         return {
             'sample_id': self.mt.s,
             'gq': self.mt.GQ,
             'cn': self.mt.RD_CN,
-            'num_alt': self._num_alt(is_called)
+            'num_alt': num_alt,
+            'prev_num_alt': hl.or_missing(discordant_genotype, prev_num_alt),
+            'new_call': hl.or_missing(is_called, ~was_previously_called | novel_genotype | discordant_genotype),
```
bpblanken, line 227: | new_call | new_genotype
-- | -- | --
Absent in previous callset. | T | F
Concordant Genotype | F | F
Novel Genotype | T | F
Absent Genotype | F | T
Discordant Genotype | F | T


Should I have instead left this `new_call` field as is and just modified the filter function in `samples_new_call` to filter for samples that had either `new_call` OR `prev_num_alt != missing`?  That made less sense to me than just expanding the definition here.

hanars: yeah, you should have left `new_call` as it was and modified `samples_new_call`. `samples_new_call` is what seqr will use to filter data at search time, the genotype field is what it will use to display data. For filtering, the ticket specifies that the existing filtering for "new call" should return new genotypes as well, so we want to change the `samples_new_call` field to include all the samples that search should match on. But then for display we will display new call and new genotype differently from each other

bpblanken: 👍 Yep that makes sense.

luigi_pipeline/seqr_loading.py
```diff
@@ -69,7 +69,7 @@ def run(self):
         # first validate paths
         for source_path in self.source_paths:
             check_if_path_exists(source_path, "source_path")
-        check_if_path_exists(self.reference_ht_path, "reference_ht_path")
+        if self.reference_ht_path: check_if_path_exists(self.reference_ht_path, "reference_ht_path")
```
bpblanken, line 72: out of scope but tests now fail because we're running through this pathway now.  I think this is probably fine for now... (clinvar has the same logic and is also required and we're going to fail when we try to load the reference data anyways)... but should realistically we should perform this check if `datasetType === SNV or datasetType === MITO`.

luigi_pipeline/lib/model/sv_mt_schema.py
hanars: Is there any expected case where `CONC_ST` does not match one of the defined keys in `PREVIOUS_GENOTYPE_N_ALT_ALLELES`? If not, I would rather not use a safe get with a default and instead have the pipeline run fail on an unexpected input

hanars (changes requested):
the logic all looks good to me now! just one outstanding question

luigi_pipeline/lib/model/sv_mt_schema.py
bpblanken: 👍 failing on unexpected is good by me!

bpblanken: @hanars 

Got this working but it was painful.  Ideally we'd have:
```
prev_num_alt = hl.or_missing(PREVIOUS_GENOTYPE_N_ALT_ALLELES[hl.set(self.mt.CONC_ST)])
```
but that's gonna fail until hail releases the bug fix.  I put the missing value and `["EMPTY"]` in the lookup for now with a comment.  Are we ok with that for now (I'll make a ticket to fix pending hail release)?

hanars: I see what you mean about this being ugly, would leaving the missing/empty values out of the dict and then doing something like so you only use the lookup expression when there is a valid value work?
```
prev_num_alt = PREVIOUS_GENOTYPE_N_ALT_ALLELES[hl.set(self.mt.CONC_ST)]
discordant_genotype = was_previously_called & (num_alt != prev_num_alt) & (prev_num_alt > 0)
novel_genotype = hl.or_else(~was_previously_called, (num_alt != prev_num_alt) & (prev_num_alt == 0))
```

bpblanken: Hrm.... still having trouble.

I've tried:

```
was_previously_called = hl.is_defined(self.mt.CONC_ST) & ~self.mt.CONC_ST.contains("EMPTY")
prev_num_alt = PREVIOUS_GENOTYPE_N_ALT_ALLELES[hl.set(self.mt.CONC_ST)]
discordant_genotype = was_previously_called & (num_alt != prev_num_alt) & (prev_num_alt > 0)
novel_genotype = was_previously_called & (num_alt != prev_num_alt) & (prev_num_alt == 0)

// and
'prev_num_alt': hl.or_missing(discordant_genotype, prev_num_alt),
```

and the lookup is called.

bpblanken: Still playing around though.

bpblanken: Standard short circuiting _does_ seem to work in the way we'd expect:

```
In [17]: a = hl.dict({"1": 2})

In [18]: b = hl.dict({"2": 1})

In [19]: hl.eval(a.contains("2") & (b["3"] == 1))
Out[19]: False
```

bpblanken: Aha

Here's more:

```
In [16]: a = hl.dict({"1": 2})

In [17]: b = hl.dict({"2": 1})

In [18]: c = a.contains("2")

In [19]: d = a.contains("1")

In [20]: e = b["3"]

In [21]: hl.eval(c & d & (e > 0)) // c is False so we never evaluate the KeyError in `e`
Out[21]: False

In [22]: hl.eval(c & d & (e > 0) & (b["3"] == a["1"])) // This is also fine.
Out[22]: False

In [23]: hl.eval(c & d & (e > 0) & (e == a["1"])) // This crashes since `e` seems to be evaluated.
```

larrybabb:
@bpblanken verify that Hail is definitely working on this.

luigi_pipeline/lib/model/sv_mt_schema.py
```diff
@@ -88,11 +111,11 @@ def gnomad_svs_af(self):
 
     @row_annotation(name='gnomad_svs_AC')
     def gnomad_svs_ac(self):
-        return unsafe_cast_int32(self.mt.info.gnomad_svs_AC)
+        return unsafe_cast_int32(self.mt.info.gnomAD_V2_AC)
 
     @row_annotation(name='gnomad_svs_AN')
     def gnomad_svs_an(self):
-        return unsafe_cast_int32(self.mt.info.gnomad_svs_AN)
+        return unsafe_cast_int32(self.mt.info.gnomAD_V2_AN)
```
ShifaSZ, line 118: The `gnomAD_V2_AC` and `gnomAD_V2_AN` use to be `gnomAD_V2_AC_AF` and `gnomAD_V2_AN_AF`. Did they change the field names?

bpblanken: Yes.  I fixed the filenames on a separate pr... then accidentally typoed them to `gnomad_svs_AN`.... `gnomAD_V2_AN` is correct here.

ShifaSZ (approved):

bpblanken:
Hail release is coming: https://github.com/hail-is/hail/pull/12853

bpblanken:
Hail has released, I've updated the requirements here, and the tests now pass!  

```
bblanken@wmbbe-67c luigi_pipeline % PYTHONPATH=.. nosetests tests/test_seqr_sv_loading.py
Initializing Hail with default parameters...
23/04/13 15:08:53 WARN Utils: Your hostname, wmbbe-67c resolves to a loopback address: 127.0.0.1; using 10.242.191.18 instead (on interface en0)
23/04/13 15:08:53 WARN Utils: Set SPARK_LOCAL_IP if you need to bind to another address
23/04/13 15:08:54 WARN NativeCodeLoader: Unable to load native-hadoop library for your platform... using builtin-java classes where applicable
Setting default log level to "WARN".
To adjust logging level use sc.setLogLevel(newLevel). For SparkR, use setLogLevel(newLevel).
Running on Apache Spark version 3.3.2
SparkUI available at http://10.242.191.18:4040
Welcome to
     __  __     <>__
    / /_/ /__  __/ /
   / __  / _ `/ / /
  /_/ /_/\_,_/_/_/   version 0.2.113-cf32652c5077
LOGGING: writing to /Users/bblanken/Desktop/code/seqr-loading-pipelines/luigi_pipeline/hail-20230413-1508-0.2.113-cf32652c5077.log
2023-04-13 15:09:03.632 Hail: INFO: scanning VCF for sortedness...
2023-04-13 15:09:31.280 Hail: INFO: Coerced sorted VCF - no additional import work to do
2023-04-13 15:09:42.969 Hail: INFO: Coerced sorted dataset     (436 + 12) / 500]
2023-04-13 15:09:51.712 Hail: INFO: wrote matrix table with 11 rows and 5 columns in 11 partitions to /var/folders/r2/68lhg74s3m130ynw3bn9wcyw0000gq/T/tmpueyf8q19/tmp8x4fl6gf.mt
    Total size: 2.93 KiB
    * Rows/entries: 2.79 KiB
    * Columns: 42.00 B
    * Globals: 107.00 B
    * Smallest partition: 1 rows (199.00 B)
    * Largest partition:  1 rows (318.00 B)
2023-04-13 15:09:55.787 Hail: INFO: wrote matrix table with 11 rows and 5 columns in 11 partitions to /var/folders/r2/68lhg74s3m130ynw3bn9wcyw0000gq/T/tmpueyf8q19/tmp_ostpi8k.mt
    Total size: 1.93 KiB
    * Rows/entries: 1.78 KiB
    * Columns: 42.00 B
    * Globals: 107.00 B
    * Smallest partition: 1 rows (152.00 B)
    * Largest partition:  1 rows (215.00 B)
.Stage 11:>                                                       (0 + 11) / 11]
Ran 1 test in 71.787s

OK
```

hanars (approved):

PR #429, opened 2023-03-31
multiple dataset types flowing through this method now... be specific when path checking.

bpblanken:
Fixes a bug introduced by directing all of the variants tasks through the same `run` method.

### DIFF
--- a/luigi_pipeline/seqr_loading.py
+++ b/luigi_pipeline/seqr_loading.py
@@ -69,9 +69,10 @@ class SeqrVCFToMTTask(HailMatrixTableTask):
         # first validate paths
         for source_path in self.source_paths:
             check_if_path_exists(source_path, "source_path")
-        check_if_path_exists(self.reference_ht_path, "reference_ht_path")
+        if self.dataset_type in set(['VARIANTS', 'MITO']):
+            check_if_path_exists(self.reference_ht_path, "reference_ht_path")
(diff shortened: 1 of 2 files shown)


luigi_pipeline/seqr_loading.py
```diff
@@ -69,9 +69,10 @@ def run(self):
         # first validate paths
         for source_path in self.source_paths:
             check_if_path_exists(source_path, "source_path")
-        check_if_path_exists(self.reference_ht_path, "reference_ht_path")
+        if self.dataset_type in set(['VARIANTS', 'MITO']):
+            check_if_path_exists(self.reference_ht_path, "reference_ht_path")
+            check_if_path_exists(self.clinvar_ht_path, "clinvar_ht_path")
```
hanars, line 74: this previously was conditional on `if self.clinvar_ht_path:`

bpblanken: Yep!  per [this line](https://github.com/broadinstitute/seqr-loading-pipelines/pull/429/files#diff-8ec7d2f4cfdee1a89ae994d56d5de0648a9f3928a08124029e77bee2dcc245cdR88) clinvar is not optional.  Both `GCNV` and `SV` override `get_schema_class_kwargs` so they won't call that line.

hanars (approved):

PR #430, opened 2023-03-31
Add `v03` clinvar script.

bpblanken:
- Writes to a `v03` prefix so that the existing stuff doesn't change.
- Only keeps the fields needed upstream.  I think we want to do this (we get a 2/3 size reduction on gcs)?
- Some fixes to the ftp download to make it work locally.  

Resolves https://github.com/broadinstitute/seqr-pipeline-airflow/issues/129


hail_scripts/utils/clinvar.py
```diff
@@ -34,18 +35,19 @@ def download_and_import_latest_clinvar_vcf(genome_version: str) -> hl.MatrixTabl
         raise ValueError("Invalid genome_version: " + str(genome_version))
 
     clinvar_url = CLINVAR_FTP_PATH.format(genome_version=genome_version)
-    local_tmp_file_path = "/tmp/clinvar.vcf.gz"
```
bpblanken, line 37: Ran into reliability issues with this code.  `local_tmp_file_path` was getting corrupted... also I didn't have `wget` and we don't need to shell exec.

hail_scripts/reference_data/constants.py
```diff
@@ -0,0 +1 @@
+GCS_PREFIX = "gs://seqr-reference-data/GRCh{genome_version}/v03/{environment}"
```
bpblanken, line 1: With how this is written now, we'll get both a `dev` and a `prod` prefix here.  I think `dev` is nice but `prod` might be unnecessary.  The implementation will be slightly trickier though if we want to have `prod` mean `''` though.  Anyone else have thoughts?

download_and_create_reference_datasets/v03/write_clinvar_ht.py
```diff
@@ -0,0 +1,44 @@
+#!/usr/bin/env python3
```
bpblanken, line 1: I made a completely separate script for `v03` so the old one (with the existing path structure and fields) is preserved.

download_and_create_reference_datasets/v03/write_clinvar_ht.py
```diff
@@ -0,0 +1,44 @@
...
+import hail as hl
+
+from hail_scripts.reference_data.constants import GCS_PREFIX
+from hail_scripts.utils.clinvar import (
+    download_and_import_latest_clinvar_vcf,
+    CLINVAR_GOLD_STARS_LOOKUP,
+)
+from hail_scripts.utils.hail_utils import write_ht
+
+CLINVAR_HT_PATH = os.path.join(GCS_PREFIX, 'clinvar/clinvar.GRCh{genome_version}.{timestamp}.ht')
```
ShifaSZ, line 14: The `CLINVAR_HT_PATH` in `hail_scripts.utils.clinvar.py` is unused now.

download_and_create_reference_datasets/v03/write_clinvar_ht.py
```diff
@@ -0,0 +1,44 @@
...
+        ht = ht.transmute(info=ht.info.select('ALLELEID', 'CLNSIG')).select('info', 'gold_stars')
+        ht = ht.repartition(100)
+        destination_path = CLINVAR_HT_PATH.format(
+            environment=environment,
+            genome_version=genome_version,
+            timestamp=timestamp,
+        )
+        print(f'Uploading ht from {hl.eval(ht.sourceFilePath)} to {destination_path}')
+        write_ht(ht, destination_path)
+        os.remove(ht.sourceFilePath)
```
ShifaSZ, line 34: Should we also delete the downloaded VCF file?

bpblanken: `sourceFilePath` is the VCF file here.  the `ht` is only in memory.

bpblanken: well... until we write to `gcs`.

download_and_create_reference_datasets/v03/write_clinvar_ht.py
bpblanken: This a good question.  I'm not sure if we want to preserve the "run on dataproc" wrapper scripts going forwards.... though I can see a case!  I think nobody else (the open source users of this repo) would run these reference data generators right?

download_and_create_reference_datasets/v03/write_clinvar_ht.py
ShifaSZ: I see.

download_and_create_reference_datasets/v03/write_clinvar_ht.py
hanars: I would argue we in fact *don't* want to preserve the wrapper scripts as I think they really reduce readability. Nobody but us needs to run these, so they should really work in one way, which should be the way we intend to call them

hail_scripts/reference_data/constants.py
hanars: I don't want us to pay to maintain dev versions of files, and any downstream dev versions of the pipeline should for the most part be using the main production files. If we want to support writing test versions of these files I would say we should be writing them to `gs://seqr-scratch-temp`

download_and_create_reference_datasets/v03/write_clinvar_ht.py
```diff
@@ -0,0 +1,44 @@
...
+    for genome_version in ['37', '38']:
+        mt = download_and_import_latest_clinvar_vcf(genome_version)
+        timestamp = hl.eval(mt.version)
+        ht = mt.rows()
+        ht = ht.annotate(
+            gold_stars=CLINVAR_GOLD_STARS_LOOKUP.get(hl.delimit(ht.info.CLNREVSTAT))
+        )
+        ht.describe()
+        ht = ht.transmute(info=ht.info.select('ALLELEID', 'CLNSIG')).select('info', 'gold_stars')
+        ht = ht.repartition(100)
```
hanars, line 26: how is this number chosen? Do we want to compute it more explicitly, or if not should we make it a named constant?

download_and_create_reference_datasets/v03/write_clinvar_ht.py
```diff
@@ -0,0 +1,44 @@
...
+def run(environment: str):
+    for genome_version in ['37', '38']:
+        mt = download_and_import_latest_clinvar_vcf(genome_version)
+        timestamp = hl.eval(mt.version)
+        ht = mt.rows()
+        ht = ht.annotate(
+            gold_stars=CLINVAR_GOLD_STARS_LOOKUP.get(hl.delimit(ht.info.CLNREVSTAT))
+        )
+        ht.describe()
+        ht = ht.transmute(info=ht.info.select('ALLELEID', 'CLNSIG')).select('info', 'gold_stars')
```
hanars, line 25: In the current implementation of the new backend, we expect the fields `clinical_significance_id`,`alleleId`, and `goldStars`, so this is close but not quite the same. `clinical_significance_id` is essentially the enum representation of CLNSIG, mapping all the strings to the much more compact integers. If we do not do this here we will have to do it during the data loading step so it seems better to me to do that mapping here and save extra storage space/ make the clinvar annotation in the pipeline as minimal as possible.

download_and_create_reference_datasets/v03/write_clinvar_ht.py
hanars: if we are running this on dataproc do we need to clean up the file system?

hail_scripts/utils/clinvar.py
```diff
@@ -34,18 +34,19 @@ def download_and_import_latest_clinvar_vcf(genome_version: str) -> hl.MatrixTabl
...
-    subprocess.run(["hdfs", "dfs", "-copyFromLocal", "-f", f"file://{local_tmp_file_path}", local_tmp_file_path])
-    clinvar_release_date = _parse_clinvar_release_date(local_tmp_file_path)
-    mt_contig_recoding = {'MT': 'chrM'} if genome_version == '38' else None
-    mt = import_vcf(local_tmp_file_path, genome_version, drop_samples=True, min_partitions=2000, skip_invalid_loci=True,
-                    more_contig_recoding=mt_contig_recoding)
-    mt = mt.annotate_globals(version=clinvar_release_date)
-
-    return mt
-
+    with tempfile.NamedTemporaryFile(delete=False, suffix='.vcf.gz') as local_tmp_file_path:
```
hanars, line 37: would be better to name the variable `local_tmp_file` as it is the file object and not the path

hanars (changes requested):

download_and_create_reference_datasets/v03/write_clinvar_ht.py
bpblanken: This is just copied from the existing script.  I'm not sure!  

I'm down to make this a constant though.

download_and_create_reference_datasets/v03/write_clinvar_ht.py
bpblanken: 👍 I hadn't ported over the mapping cause we hadn't figured out enum sharing yet but a dictionary is fine for now!

download_and_create_reference_datasets/v03/write_clinvar_ht.py
bpblanken: Aha: https://github.com/broadinstitute/seqr-loading-pipelines/pull/383

hail_scripts/reference_data/constants.py
bpblanken: 👍 I can just make the paths the same for `dev`/`prod` and have `dev` -> `gs://seqr-scratch-temp` and `prod` -> `gs://seqr-reference-data`.

fwiw you can setup a retention policy on path prefixes: https://cloud.google.com/blog/products/storage-data-transfer/understanding-cloud-storages-new-prefix-and-suffix-lifecycle-rules but that can get kinda messy.

download_and_create_reference_datasets/v03/write_clinvar_ht.py
bpblanken: 👍 I agree!

download_and_create_reference_datasets/v03/write_clinvar_ht.py
bpblanken: Ha.  Shower thought me remembered about the delimiter.  Ignore 2!

bpblanken: I think I understand what you're doing with the snake-case `_id`.  It's reserved for string mapping?  That makes sense to me.

hanars: I do think standardizing would be nice though - there are going to be a lot of `_id` fields in our final data. Maybe we do `finalVarianbleName_id` as our naming convention, so in this case `clinicalSignificance_id`? Or if that seems way too ugly I'm open to other suggestions

download_and_create_reference_datasets/v03/write_clinvar_ht.py
bpblanken: Honestly the cleanest answer here is to do all this work within the `with tempfile.NamedTemporaryFile(...)` context manager which will manage cleanup itself.  We've got some work within the context manager and some outside which is why this cleanup line kinda seems out of place.

hail_scripts/utils/clinvar.py
```diff
@@ -9,6 +9,43 @@
...
+            'Benign/Likely_benign',
+            'Benign',
+        ],
+        index_first=False
+    )
+)
+CLINVAR_CLINICAL_SIGNIFICANCE_MODIFIER_LOOKUP = hl.dict(
+    hl.enumerate(
+        # NB: alphabetical
+        [
```
bpblanken, line 33: missing risk_factor

hail_scripts/utils/clinvar.py
```diff
@@ -9,6 +9,43 @@
...
+    hl.enumerate(
+        # NB: sorted by pathogenicity
+        [
+            'Pathogenic',
+            'Pathogenic/Likely_pathogenic',
+            'Likely_pathogenic',
+            'Likely_risk_allele',
+            'Conflicting_interpretations_of_pathogenicity',
+            'Uncertain_risk_allele',
+            'Uncertain_significance',
```
bpblanken, line 22: New one is "No_pathogenic_assertion".

hail_scripts/utils/clinvar.py
```diff
@@ -9,6 +9,43 @@
 CLINVAR_FTP_PATH = "ftp://ftp.ncbi.nlm.nih.gov/pub/clinvar/vcf_GRCh{genome_version}/clinvar.vcf.gz"
 CLINVAR_HT_PATH = "gs://seqr-reference-data/GRCh{genome_version}/clinvar/clinvar.GRCh{genome_version}.ht"
 
+CLINVAR_CLINICAL_SIGNIFICANCE_LOOKUP = hl.dict(
```
bpblanken, line 12: Call this pathogenicity.

hail_scripts/utils/clinvar.py
```diff
@@ -9,6 +9,43 @@
...
+            'Uncertain_risk_allele',
+            'Uncertain_significance',
+            'Likely_benign',
+            'Benign/Likely_benign',
+            'Benign',
+        ],
+        index_first=False
+    )
+)
+CLINVAR_CLINICAL_SIGNIFICANCE_MODIFIER_LOOKUP = hl.dict(
```
bpblanken, line 30: Call this assertions.

download_and_create_reference_datasets/v03/write_clinvar_ht.py
```diff
@@ -0,0 +1,73 @@
...
+
+def run(environment: str):
+    for genome_version in ['37', '38']:
+        with tempfile.NamedTemporaryFile(delete=False, suffix='.vcf.gz') as tmp_file:
+            mt = download_and_import_latest_clinvar_vcf(genome_version, tmp_file)
+            timestamp = hl.eval(mt.version)
+            ht = mt.rows()
+            ht = ht.select(
+                alleleId=ht.info.ALLELEID,
+                pathogenicities_id=CLINVAR_PATHOGENICITIES_LOOKUP.get(
```
bpblanken, line 31: Another potential option here is to filter out the assertions, then do non-safe lookup of the pathogenicity lookup so we fail if we find an unexpected one.

hail_scripts/utils/clinvar.py
```diff
@@ -21,9 +40,68 @@
...
+        [
+            'Pathogenic',
+            'Pathogenic/Likely_pathogenic',
+            'Pathogenic/Likely_pathogenic/Likely_risk_allele',
+            'Pathogenic/Likely_risk_allele',
+            'Likely_pathogenic',
+            'Likely_pathogenic/Likely_risk_allele',
+            'Likely_risk_allele',
+            'Conflicting_interpretations_of_pathogenicity',
+            'Established_risk_allele',
```
bpblanken, line 55: Wanted to call out that I added `Established_risk_allele` here, which isn't present in `CLNSIG` but is present in `CLNSIGCONF`.

download_and_create_reference_datasets/v03/write_clinvar_ht.py
hanars: I think we also do want to make sure that the pathogenicity is only at the first position in the list, I think the way this is currently written it would not throw an error if the second list item was a pathogenicity and just skip it, and I think the other potential solution you propose would also have that issue. I think maybe we can leave this as is, but then add a conditional for where we start checking the assertion list, so something like 
```
assertions_ids=hl.if_else(CLINVAR_PATHOGENICITIES_LOOKUP.contains(clinsigs[0]), clinsigs[1:], clinsigs).map(
  lambda x: CLINVAR_ASSERTIONS_LOOKUP[x]
)
```

hail_scripts/utils/clinvar.py
```diff
@@ -21,9 +40,68 @@
...
+            'Uncertain_risk_allele',
+            'Uncertain_significance/Uncertain_risk_allele',
+            'Uncertain_significance',
+            'Likely_benign',
+            'Benign/Likely_benign',
+            'Benign',
+            CLINVAR_DEFAULT_PATHOGENICITY,
+        ],
+        index_first=False
```
hanars, line 64: what does this do?

download_and_create_reference_datasets/v03/write_clinvar_ht.py
```diff
@@ -0,0 +1,73 @@
...
+                    .starmap(lambda pathogenicity, _: CLINVAR_PATHOGENICITIES_LOOKUP[pathogenicity])
+                ),
+                conflictingPathogenicities_counts=(
+                    parsed_clnsigconf(ht)
+                    .starmap(lambda _, count: count)
+                ),
+                goldStars=CLINVAR_GOLD_STARS_LOOKUP.get(hl.delimit(ht.info.CLNREVSTAT)),
+            ).annotate_globals(
+                enum_definitions=hl.dict({
+                    'pathogenicities_id': CLINVAR_PATHOGENICITIES_LOOKUP,
```
hanars, line 51: since by definition the script that is reading this table needs to do a reverse lookup, I think it might be more straightforward to put the array of options into the global instead of the dictionary lookup

hail_scripts/utils/clinvar.py
hanars: This should be sorted above `Likely_risk_allele`

hail_scripts/utils/clinvar.py
```diff
@@ -21,9 +40,68 @@
...
+            'Likely_risk_allele',
+            'Conflicting_interpretations_of_pathogenicity',
+            'Established_risk_allele',
+            'Uncertain_risk_allele',
+            'Uncertain_significance/Uncertain_risk_allele',
+            'Uncertain_significance',
+            'Likely_benign',
+            'Benign/Likely_benign',
+            'Benign',
+            CLINVAR_DEFAULT_PATHOGENICITY,
```
hanars, line 62: this should be sorted above `Likely_benign`

download_and_create_reference_datasets/v03/write_clinvar_ht.py
```diff
@@ -0,0 +1,73 @@
...
+from hail_scripts.utils.hail_utils import write_ht
+
+CLINVAR_HT_PATH = 'clinvar/clinvar.GRCh{genome_version}.{timestamp}.ht'
+PARTITIONS = 100 # per https://github.com/broadinstitute/seqr-loading-pipelines/pull/383
+
+def run(environment: str):
+    for genome_version in ['37', '38']:
+        with tempfile.NamedTemporaryFile(suffix='.vcf.gz') as tmp_file:
+            mt = download_and_import_latest_clinvar_vcf(genome_version, tmp_file)
+            timestamp = hl.eval(mt.version)
```
hanars, line 27: just want to double check that this global will also be present on the final ht

hail_scripts/utils/clinvar.py
```diff
@@ -21,9 +40,68 @@
...
-
-def download_and_import_latest_clinvar_vcf(genome_version: str) -> hl.MatrixTable:
+def parsed_clnsig(ht: hl.Table):
+    return (
+        hl.delimit(ht.info.CLNSIG)
+        .replace(
+            'Likely_pathogenic,_low_penetrance', 'Likely_pathogenic|low_penetrance',
+        )
+        .replace(
+            '/Pathogenic,_low_penetrance', '|low_penetrance',
```
hanars, line 75: Do we ever just have `Pathogenic,_low_penetrance` as the CLINSIG?

hail_scripts/utils/clinvar.py
```diff
@@ -21,9 +40,68 @@
...
+        )
+        .split(r'\|')
+    )
+
+def parsed_clnsigconf(ht: hl.Table):
+
+    def parse_to_count(entry: str):
```
hanars, line 82: is there a reason to have this be a nested function?

hail_scripts/utils/clinvar.py
```diff
@@ -21,9 +40,68 @@
...
+            hl.int32(splt[1][:-1])
+        ])
+
+    return (
+        hl.delimit(ht.info.CLNSIGCONF)
+        .replace(',_low_penetrance', '')
+        .split(r'\|')
+        .map(parse_to_count)
+        .group_by(lambda x: x[0])
+        .map_values(lambda values: (
```
hanars, line 95: Shouldn't each pathogenicity only appear once with a count? Why do we need to do a group and summation?

hanars (changes requested):

download_and_create_reference_datasets/v03/write_clinvar_ht.py
bpblanken: Yep!  I had the same idea here: https://github.com/broadinstitute/seqr-loading-pipelines/pull/440#discussion_r1171862792

I like that!

hail_scripts/utils/clinvar.py
bpblanken: See the test here for an example: https://github.com/broadinstitute/seqr-loading-pipelines/pull/430/files/515135c47a496173421f57e277caf2ac4a8d7bc2#diff-9a66d5d4eb152375e3619fb8871357ab5a5196c837b9a0e9d03d4f57bdf6a66eR38

`Pathogenic,_low_penetrance` shows up as a unique pathogenicity.

hail_scripts/utils/clinvar.py
bpblanken: flips the index of the enumerate to be second rather than first:

```
In [4]: hl.eval(hl.enumerate(['a', 'b', 'c']))
Out[4]: [(0, 'a'), (1, 'b'), (2, 'c')]

In [5]: hl.eval(hl.enumerate(['a', 'b', 'c'], index_first=False))
Out[5]: [('a', 0), ('b', 1), ('c', 2)]
```

makes the `hl.dict` call work!

hail_scripts/utils/clinvar.py
hanars: 😢 unfortunately I can't think of a better way to do this, given that thats how the data is delivered

hail_scripts/utils/clinvar.py
bpblanken: Reasoning for it being nested is it's just `lambda` I moved up so that I could assign a variable.  It shouldn't have a meaning outside of this function.  Happy to move it up though!

download_and_create_reference_datasets/v03/write_clinvar_ht.py
bpblanken: should be there on this line: https://github.com/broadinstitute/seqr-loading-pipelines/pull/430/files/515135c47a496173421f57e277caf2ac4a8d7bc2#diff-a49ce7f83dbd25ead01293ef8d1eed2f0c0e79aac5607c74511d916be6848b8fR125

hanars: I just wanted to make sure that none of the selects we are doing overwrite the global, but I realize that you have to explicitly `select_globals` to do that

download_and_create_reference_datasets/v03/write_clinvar_ht.py
```diff
@@ -0,0 +1,73 @@
...
+            ht = mt.rows()
+            ht = ht.select(
+                alleleId=ht.info.ALLELEID,
+                pathogenicities_id=CLINVAR_PATHOGENICITIES_LOOKUP.get(
+                    parsed_clnsig(ht)[0], 
+                    CLINVAR_PATHOGENICITIES_LOOKUP[CLINVAR_DEFAULT_PATHOGENICITY],
+                ),
+                assertions_ids=(
+                    parsed_clnsig(ht)
```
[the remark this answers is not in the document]
hanars: If its been lifted up to a variable hail will only use it once. From my understanding theres no performance advantage to `hl.bind` if you have a variable defined for something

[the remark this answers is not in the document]
bpblanken: 👍

download_and_create_reference_datasets/v03/write_clinvar_ht.py
```diff
@@ -0,0 +1,75 @@
...
+                    CLINVAR_PATHOGENICITIES_LOOKUP[CLINVAR_DEFAULT_PATHOGENICITY],
+                ),
+                assertion_ids=hl.if_else(
+                    CLINVAR_PATHOGENICITIES_LOOKUP.contains(clnsigs[0]),
+                    clnsigs[1:],
+                    clnsigs,
+                ).map(lambda x: CLINVAR_ASSERTIONS_LOOKUP[x]),
+                conflictingPathogenicities=(
+                    parsed_clnsigconf(ht)
+                    .starmap(lambda pathogenicity, count: hl.Struct(
```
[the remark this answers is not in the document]
bpblanken: 👍 I moved the pathogenicity lookup into `parsed_clnsigconf` so it's also tested now.

hail_scripts/utils/clinvar.py
hanars: Yeah I think moving it up would be clearer

hail_scripts/utils/clinvar.py
```diff
@@ -21,9 +36,67 @@
...
+            values.fold(
+                lambda x, y: x + y[1],
+                0,
+            )
+        ))
+        .items()
+        .map(
+            lambda x: hl.Struct(pathogenicity_id=CLINVAR_PATHOGENICITIES_LOOKUP[x[0]], count=x[1])
```
hanars, line 95: This works, but I think the whole thing might be more readable if `parse_to_count` returned a struct in the first place.For example, I think that `.group_by(lambda x: x. pathogenicity_id)` is more readable than `.group_by(lambda x: x[0])`, etc.

hanars (approved):
looks good to me! I would also want @mike-w-wilson to take a look though

hail_scripts/utils/clinvar.py
mike-w-wilson: I tried thinking of a different approach as well but couldnt

mike-w-wilson (approved):
This looks good to me. I had no idea that these modifiers existed...do we know how common they are? Also just a general question for v03, we flip between camel case and snake case(did this in v02 too), is the idea to maintain that for now because seqr is expecting certain values until we can flip to one or the other in one big formatting PR, like running black on the entire repo? Obviously this would need to be coordinated with seqr. Anyways, this LGTM!

hail_scripts/utils/clinvar.py
```diff
@@ -21,9 +36,66 @@
...
+CLINVAR_PATHOGENICITIES_LOOKUP = hl.dict(hl.enumerate(CLINVAR_PATHOGENICITIES, index_first=False))
+
+def parsed_clnsig(ht: hl.Table):
+    return (
+        hl.delimit(ht.info.CLNSIG)
+        .replace(
+            'Likely_pathogenic,_low_penetrance', 'Likely_pathogenic|low_penetrance',
+        )
+        .replace(
+            '/Pathogenic,_low_penetrance', '|low_penetrance',
```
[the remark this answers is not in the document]
bpblanken: We occasionally see things like:

```
'Pathogenic/Likely_pathogenic/Pathogenic,_low_penetrance'
```
as pathogenicities.  We've decided that we want to parse this as a pathogenicity of `Pathogenic/Likely_pathogenic` and an assertion of `low_penetrance`.

hanars:
> Also just a general question for v03, we flip between camel case and snake case(did this in v02 too)

I think the idea is in v3 to be consistent and converge on camel case as it reduces the number of transformation seqr will have to do. The one exception is we seem to have adopted the convention of `camelCase_id` for fields where we have mapped a string to an integer id, but if folks don't like that we can have a different naming convention. Otherwise for v3 though, we should be consitant and change variable name as needed

PR #431, opened 2023-04-03
Fieldnames from SV callset.

bpblanken:
Updates filenames so that they correspond to the latest callset.

Also addresses a smattering of other issues.


luigi_pipeline/tests/test_seqr_sv_loading.py
```diff
@@ -254,7 +254,7 @@ def setUp(self):
     def tearDown(self):
         shutil.rmtree(self._temp_dir.name)
 
-    @mock.patch('seqr_sv_loading.load_gencode', return_value=GENE_ID_MAPPING)
+    @mock.patch('luigi_pipeline.seqr_sv_loading.load_gencode', return_value=GENE_ID_MAPPING)
```
bpblanken, line 257: Fixing a bug from the packaging pr... I've fixed it on another pr but it's not merged yet.

bpblanken:
On hold for a few more changes.

luigi_pipeline/lib/model/sv_mt_schema.py
```diff
@@ -74,15 +81,15 @@ def gnomad_svs_af(self):
 
     @row_annotation(name='gnomad_svs_AC')
     def gnomad_svs_ac(self):
-        return self.mt.info.gnomAD_V2_AC_AF
+        return hl.int32(self.mt.info.gnomAD_V2_AC)
```
bpblanken, line 84: I ran `mt.aggregate_rows(hl.agg.counter(mt.info.gnomAD_V2_AC))` and didn't see any non-integers (though I didn't look very thoroughly).  This'll throw if we encounter a float, so I think we're good as is here.

luigi_pipeline/lib/model/sv_mt_schema.py
```diff
@@ -74,15 +81,15 @@ def gnomad_svs_af(self):
...
 
     @row_annotation(name='gnomad_svs_AN')
     def gnomad_svs_an(self):
-        return self.mt.info.gnomAD_V2_AN_AF
+        return hl.int32(self.mt.info.gnomAD_V2_AN)
 
     @row_annotation(name='StrVCTVRE_score')
     def strvctvre(self):
-        return self.mt.info.StrVCTVRE
+        return hl.parse_float(self.mt.info.StrVCTVRE)
```
bpblanken, line 92: Confirmed that this'll just return `missing` if we can't parse a float.

luigi_pipeline/lib/model/sv_mt_schema.py
```diff
@@ -203,16 +205,11 @@ def _genotype_fields(self):
...
-    def _genotype_filter_samples(self, filter_):
-        samples = super()._genotype_filter_samples(filter_)
-        return hl.or_missing(hl.len(samples) > 0, samples)            
-
     @row_annotation(name="samples_gq_sv", fn_require=SeqrGenotypesSchema.genotypes)
     def samples_gq(self):
         # NB: super().samples_gq is a RowAnnotation... so we call the method under the hood.
         # ew it is gross.
-        return super().samples_gq.fn(self, start=0, end=1000, step=10)
+        return super().samples_gq.fn(self, start=0, end=100, step=10)
```
bpblanken, line 212: The inherited function `ends` at `95`... we want a 90->100 bucket right?  

Should `step` be `5`?

luigi_pipeline/lib/model/sv_mt_schema.py
```diff
@@ -203,16 +205,11 @@ def _genotype_fields(self):
             'num_alt': self._num_alt(is_called)
         }
 
-    # NB: override this function here to mimic the existing null handling behavior.
-    def _genotype_filter_samples(self, filter_):
-        samples = super()._genotype_filter_samples(filter_)
-        return hl.or_missing(hl.len(samples) > 0, samples)
```
bpblanken, line 209: I'd added this so that we return missing values.... now we return the empty set.

luigi_pipeline/lib/model/sv_mt_schema.py
```diff
@@ -158,7 +160,7 @@ def sv_type(self):
         sorted_transcript_consequences, sv_type,
     ])
     def transcript_consequence_terms(self):
-        return self.mt.sortedTranscriptConsequences.map(lambda x: x[MAJOR_CONSEQUENCE]).extend([self.mt.svType])
+        return hl.set(self.mt.sortedTranscriptConsequences.map(lambda x: x[MAJOR_CONSEQUENCE]).extend([self.mt.svType]))
```
bpblanken, line 163: This actually deviates from the previous behavior: https://github.com/broadinstitute/seqr-loading-pipelines/blob/4d0eb3486e14da9664a1c00810e395e0dbf6bfd3/sv_pipeline/genome/load_data_tests.py#L162

I think it's reasonable to not have duplicates though!

bpblanken: Also, I checked the old index and it also has duplicates!

luigi_pipeline/lib/model/sv_mt_schema.py
```diff
@@ -37,6 +37,13 @@ def __init__(self, *args, gene_id_mapping=None, **kwargs):
         super().__init__(*args, **kwargs)
         self._gene_id_mapping = gene_id_mapping
 
+    def end_locus(self):
```
bpblanken, line 40: This was a `row_annotation`, now a regular instance method.

luigi_pipeline/lib/model/sv_mt_schema.py
hanars: okay well I guess we didn't have to change this then, but yay making things better!

luigi_pipeline/lib/model/sv_mt_schema.py
hanars: step should still be 10, and no actually we do not need 90_100, seqr does searches by querying for "quality is not in these low bins" so you never need the top bin, which svaes a lot of space

luigi_pipeline/lib/model/sv_mt_schema.py
bpblanken: ah...

I did 
```
hl.eval(hl.int32('1.1'))
```

luigi_pipeline/lib/model/sv_mt_schema.py
bpblanken: 👍 ... though just calling out this deviates from previous behavior: https://github.com/broadinstitute/seqr-loading-pipelines/blob/4d0eb3486e14da9664a1c00810e395e0dbf6bfd3/sv_pipeline/genome/load_data_tests.py#L150

hanars (approved):

hanars (approved):

PR #433, opened 2023-04-04
lift hail init to top of `run`

bpblanken:
`check_if_path_exists` calls `hl.hadoop_exists` which will call `hl.init`.  This leads to calling `hl.init` multiple times... which is a crash.

Tagging @mike-w-wilson since I mentioned this to you in our meeting.


ShifaSZ (approved):
LGTM

PR #437, opened 2023-04-07
Sex check

mike-w-wilson:
This is the initial attempt at getting the sex check into luigi. There's room for improvement here, e.g. inheritance, but I wanted to make sure I was on the right path before messing with any of the other classes if thats even an approach we want to take. I've tested this on a single shard of our most recent callset and the results are here: `gs://seqr-loading-temp/luigi-sex-check`. The hail stuff is identical to a script that passed review last year so while I think there is some room for improvement, I'd love an SE take of this from the luigi framework point of view -- but obviously comments on the hail bits are welcome.

### DIFF
--- a/luigi_pipeline/sex_related_checks.py
+++ b/luigi_pipeline/sex_related_checks.py
@@ -1,0 +1,568 @@
+import logging
+import pkg_resources
+import sys
+from typing import List
+import matplotlib.pyplot as plt
+
+import luigi
+import hail as hl
+
+from luigi_pipeline.lib.hail_tasks import (
+    HailMatrixTableTask,
+    GCSorLocalTarget,
+)
+from luigi_pipeline.seqr_loading import (
+    check_if_path_exists,
+    contig_check,
+    SeqrValidationError,
+)
+
+logger = logging.getLogger(__name__)
+
+PYSPARK_SUBMIT_ARGS = "--driver-memory 400G pyspark-shell"
+
+FEMALE_PLOIDY = "XX"
+MALE_PLOIDY = "XY"
+SEX_PLOIDY_ANNOTATIONS = [
+    "is_female",
+    "f_stat",
+    "n_called",
+    "expected_homs",
+    "observed_homs",
+    "sex",
+]
+GRCh37_STANDARD_CONTIGS = {
+    "1",
+    "10",
+    "11",
+    "12",
+    "13",
+    "14",
+    "15",
+    "16",
+    "17",
+    "18",
+    "19",
+    "2",
+    "20",
+    "21",
+    "22",
+    "3",
+    "4",
+    "5",
+    "6",
+    "7",
+    "8",
+    "9",
+    "X",
+    "Y",
+    "MT",
+}
+GRCh38_STANDARD_CONTIGS = {
+    "chr1",
+    "chr10",
+    "chr11",
+    "chr12",
+    "chr13",
+    "chr14",
+    "chr15",
+    "chr16",
+    "chr17",
+    "chr18",
+    "chr19",
+    "chr2",
+    "chr20",
+    "chr21",
+    "chr22",
+    "chr3",
+    "chr4",
+    "chr5",
+    "chr6",
+    "chr7",
+    "chr8",
+    "chr9",
+    "chrX",
+    "chrY",
+    "chrM",
+}
+OPTIONAL_CHROMOSOMES = ["MT", "chrM", "Y", "chrY"]
+VARIANT_THRESHOLD = 100
+
+
+class SexPloidyCheckTask(luigi.Task):
+    """
+    Inherits from a Hail MT Class to get helper function logic. Main logic to do annotations here.
+    """
+
+    source_paths = luigi.Parameter(
+        default="[]", description="Path or list of paths of VCFs to be loaded."
+    )
+    wes_filter_source_paths = luigi.OptionalParameter(
+        default=None, description="Path to delivered VCFs with filter annotation"
+    )
+    output_dir = luigi.Parameter(
+        description="Path to write the sex ploidy output table.",
+        default="gs://seqr-loading-temp/luigi-sex-check/",
+    )
+    temp_dir = luigi.Parameter(
+        description="Path to write the temporary output. End with '/'",
+        default="gs://seqr-scratch-temp/luigi-sex-check/",
+    )
+    genome_version = luigi.ChoiceParameter(
+        description="Reference Genome Version (37 or 38)",
+        choices=["GRCh37", "GRCh38"],
+        default="GRCh38",
+    )
+    sample_type = luigi.ChoiceParameter(
+        choices=["WGS", "WES"], description="Sample type, WGS or WES", var_type=str
+    )
+    dont_validate = luigi.BoolParameter(
+        description="Disable checking whether the dataset matches the specified "
+        "genome version and WGS vs. WES sample type."
+    )
+    remap_path = luigi.OptionalParameter(
+        default=None, description="Path to a tsv file with two columns: s and seqr_id."
+    )
+    subset_path = luigi.OptionalParameter(
+        default=None, description="Path to a tsv file with one column of sample IDs: s."
+    )
+    use_y_cov = luigi.BoolParameter(
+        description="Whether to use chromosome Y coverage when inferring sex. Note that Y coverage is required to infer sex aneuploidies."
+    )
+    add_x_cov = luigi.BoolParameter(
+        description="Whether to also calculate chromosome X mean coverage. Must be specified with use-y-cov."
+    )
+    y_cov_threshold = luigi.FloatParameter(
+        default=0.1,
+        description="Y coverage threshold used to infer sex aneuploidies (XY samples below and XX samples above this threshold will be inferred as having aneuploidies).",
+    )
+    xy_fstat_threshold = luigi.FloatParameter(
+        default=0.75,
+        description="F-stat threshold above which a sample will be called XY. Default is 0.75.",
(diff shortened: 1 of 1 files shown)


luigi_pipeline/sex_related_checks.py
```diff
@@ -0,0 +1,568 @@
...
+                check_if_path_exists(source_path, "source_path")
+        if self.hail_temp_dir:
+            check_if_path_exists(self.hail_temp_dir, "hail_temp_dir")
+
+        self.read_input_write_mt()
+        self.call_sex()
+
+    def call_sex(
+        self,
+        use_y_cov: bool = False,
```
bpblanken, line 208: This is a bit opinion-ey... but I tend to prefer there only being a single place where default argument values are injected... in this case it would be where we set up the luigi arguments.  I guess the reasoning there is to actually prevent what I think we're doing... which is calling `self.call_sex()` without any of the luigi args.  

Should we instead be passing those:

```
self.call_sex(self.use_y_cov, ..., self.call_rate_threshold)
```

luigi_pipeline/sex_related_checks.py
```diff
@@ -0,0 +1,568 @@
...
+        add_x_cov: bool = False,
+        y_cov_threshold: float = 0.1,
+        normalization_contig: str = "20",
+        xy_fstat_threshold: float = 0.75,
+        xx_fstat_threshold: float = 0.5,
+        aaf_threshold: float = 0.05,
+        call_rate_threshold: float = 0.25,
+        final_annotations: List[str] = [
```
bpblanken, line 216: Aha!  This is a [common python gotcha](https://docs.python-guide.org/writing/gotchas/#mutable-default-arguments).

I think maybe the thing to do here is to move this list to the top of the file as a constant:

```
FINAL_ANNOTATIONS = [...]
```

and then inside this function make use of that constant:

```
if use_y_cov:
    final_annotations = [*FINAL_ANNOTATIONS, *[
                     f"chr{normalization_contig}_mean_dp",
                     "chrY_mean_dp",
                     "normalized_y_coverage",
                 ]]
```

(In reality the mutability is unlikely to be a bug for us because we only invoke these functions once time per task run and then kill the dataproc cluster).

luigi_pipeline/sex_related_checks.py
```diff
@@ -0,0 +1,568 @@
...
+
+    def annotate_globals(self, mt):
+        return mt.annotate_globals(
+            sourceFilePath=",".join(self.source_paths),
+            genomeVersion=self.genome_version,
+            sampleType=self.sample_type,
+            hail_version=pkg_resources.get_distribution("hail").version,
+        )
+
+    def read_input_write_mt(self):
```
bpblanken, line 451: Should this function be calling `call_sex()`?

In our other tasks `read_input_write_mt` does the bulk of the work and then writes the final output file.

luigi_pipeline/sex_related_checks.py
```diff
@@ -0,0 +1,568 @@
...
+        default="[]", description="Path or list of paths of VCFs to be loaded."
+    )
+    wes_filter_source_paths = luigi.OptionalParameter(
+        default=None, description="Path to delivered VCFs with filter annotation"
+    )
+    output_dir = luigi.Parameter(
+        description="Path to write the sex ploidy output table.",
+        default="gs://seqr-loading-temp/luigi-sex-check/",
+    )
+    temp_dir = luigi.Parameter(
```
bpblanken, line 107: This is different than `hail_temp_dir`.  Do we need both?

luigi_pipeline/sex_related_checks.py
```diff
@@ -0,0 +1,568 @@
...
+    )
+    hail_temp_dir = luigi.OptionalParameter(
+        default="gs://seqr-scratch-temp/",
+        description="Networked temporary directory used by hail for temporary file storage. Must be a network-visible file path.",
+    )
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+
+    def output_directory(self):
```
bpblanken, line 167: I think both `output_directory` and `temp_directory` don't need to be `Targets` here... which should make the calling code a little more readable.  

I think the `Target` class is generally used for the output of the task itself, so luigi will know that the task has succeeded.

luigi_pipeline/sex_related_checks.py
```diff
@@ -0,0 +1,568 @@
...
+
+class SexPloidyCheckTask(luigi.Task):
+    """
+    Inherits from a Hail MT Class to get helper function logic. Main logic to do annotations here.
+    """
+
+    source_paths = luigi.Parameter(
+        default="[]", description="Path or list of paths of VCFs to be loaded."
+    )
+    wes_filter_source_paths = luigi.OptionalParameter(
```
bpblanken, line 100: Should this default to "[]" so we can iterate over it without a `None` check later?

bpblanken (commented):
Looks mostly good!  

I _think_ structurally we can try to hammer this into inheriting from `HailMatrixTableTask` by being clearer on output locations and by moving `contig_check` and `validate_mt` up a level.

luigi_pipeline/sex_related_checks.py
```diff
@@ -0,0 +1,568 @@
...
+
+    def import_dataset(self):
+        """
+        Imports VCF to a MatrixTable.
+
+        If optional WES filters VCF is passed, import and join with VCF.
+        :return: hl.MatirxTable
+        """
+        # Import the VCFs from inputs. Set min partitions so that local pipeline execution takes advantage of all CPUs.
+        recode = {}
```
hanars, line 539: This seems to have a bunch of copy-paste from the `HailMatrixTableTask` task, we really should have this task inherit from that. Thats would also give us cleaner access to some of the class helper functions

luigi_pipeline/sex_related_checks.py
```diff
@@ -0,0 +1,568 @@
...
+            genomeVersion=self.genome_version,
+            sampleType=self.sample_type,
+            hail_version=pkg_resources.get_distribution("hail").version,
+        )
+
+    def read_input_write_mt(self):
+        mt = self.import_dataset()
+        mt = hl.split_multi_hts(mt)
+        if not self.dont_validate:
+            self.validate_mt(mt, self.genome_version, self.sample_type)
```
hanars, line 455: probably out of scope for the initial implementation, but I think it would be better in the long run if we could have an initial "import the VCF and validate it" task, then run sex check, then run the rest of the loading but re-use the imported and validated MT

luigi_pipeline/sex_related_checks.py
```diff
@@ -0,0 +1,568 @@
...
+        Default is 0.1
+        :param normalization_contig: Chosen chromosome for calculating normalized coverage. Default is "20"
+        :param xy_fstat_threshold: F-stat threshold above which a sample will be called XY. Default is 0.75
+        :param xx_fstat_threshold: F-stat threshold below which a sample will be called XX. Default is 0.5
+        :param aaf_threshold: Alternate allele frequency threshold for `hl.impute_sex`. Default is 0.05
+        :param call_rate_threshold: Minimum required call rate. Default is 0.25
+        """
+        # Read in matrix table and define output file name prefix
+
+        logger.info("Using chromosome Y coverage? %s", use_y_cov)
```
hanars, line 242: I know we aren't doing this anywhere else, but it would be great if we could standardize our string formatting in this repo

luigi_pipeline/sex_related_checks.py
```diff
@@ -0,0 +1,568 @@
...
+
+        else:
+            sex_ht = sex_ht.annotate(ambiguous_sex=hl.is_missing(sex_ht.is_female))
+            sex_expr = hl.if_else(
+                sex_ht.ambiguous_sex,
+                "ambiguous_sex",
+                hl.if_else(sex_ht.is_female, "XX", "XY"),
+            )
+        sex_ht = sex_ht.annotate(sex=sex_expr)
+        sex_ht.write(self.output().path, stage_locally=True, overwrite=True)
```
hanars, line 326: just to clarify - at this point the output of this task is a table with the imputed sex, its is not doing any validation against the expected sex?

hanars (changes requested):

luigi_pipeline/lib/hail_tasks.py
```diff
@@ -18,22 +18,86 @@
...
         self.missing_samples = missing_samples
 
 
+class VCFValidationError(Exception):
+    pass
+
+
 def GCSorLocalTarget(filename):
-    target = gcs.GCSTarget if filename.startswith('gs://') else luigi.LocalTarget
+    target = gcs.GCSTarget if filename.startswith("gs://") else luigi.LocalTarget
```
bpblanken, line 92: Looks like your linter may be set to double quotes 🤣

mike-w-wilson: dangit

luigi_pipeline/lib/hail_tasks.py
```diff
@@ -18,22 +18,86 @@
 
 logger = logging.getLogger(__name__)
 
+GRCh37_STANDARD_CONTIGS = {
```
bpblanken, line 21: I think it'll be easier to review this if we split the `black` formatting and moving `contig_check` & `validate_mt` into its own PR?

mike-w-wilson: Yeah that is fair, I'll split this and label it as stacked -- my VSCode formats with black on save so that explains the ".

bpblanken:
Closing as we've ported this over into v03.  I shamelessly stole lots of code from here though!

PR #438, opened 2023-04-07
Moving `combine` functions to their own module inside `hail_scripts`.

bpblanken:


download_and_create_reference_datasets/v02/hail_scripts/write_combined_reference_data_ht.py
```diff
@@ -1,101 +1,15 @@
...
-from datetime import datetime
-from functools import reduce
 import os
 
 import hail as hl
 
+from hail_scripts.reference_data.combine import join_hts
 from hail_scripts.reference_data.config import CONFIG
 
 VERSION = '2.0.5'
```
bpblanken, line 9: I preserved this `VERSION` const... but honestly am not sure about it.  

We're using `gs://seqr-reference-data/GRCh38/all_reference_data/combined_reference_data_grch38.ht` which does not have `v2` in the path or a `-{version}.ht` suffix.  The embedded version inside the file is `2.0.4`... because I think the most recent update only touched interval data and we haven't run it since then?

download_and_create_reference_datasets/v02/mito/write_combined_mito_reference_data_hts.py
```diff
@@ -4,8 +4,9 @@
 
 import hail as hl
 
-from download_and_create_reference_datasets.v02.hail_scripts.write_combined_reference_data_ht import join_hts
+from hail_scripts.reference_data.combine import join_hts
 
+VERSION = '2.0.5'
```
mike-w-wilson, line 9: Should the mito version match the combined?

bpblanken: Yeah, we want it to be 2.0.4?

bpblanken: The current mito version is`2.0.4`, so I'm gonna leave it at that.

PR #440, opened 2023-04-12
Add interval reference script.

bpblanken:
~The idea here is to have two scripts:~
~1) A `write` script which preserves the existing `join_ht` for initial creation of the ht.~
~2) An `update` script which takes in the name of the source dataset that's been updated, downloads the existing combined ht and performs the "transmute".~

EDIT:

Fresh eyes me thinks we can do better and just have a single script... switching on the existence of the ht (if it's there update else create).


download_and_create_reference_datasets/v03/update_combined_interval_reference_ht.py
```diff
@@ -0,0 +1,40 @@
...
+        '--environment',
+        default='dev',
+        choices=['dev', 'prod']
+    )
+    parser.add_argument(
+        '--dataset',
+        choices=['gnomad_non_coding_constraint', 'screen'],
+        required=True,
+    )
+    args, _ = parser.parse_known_args()
```
bpblanken, line 39: I'm using `parse_known_args` here so that passing an additional arg (like `--genome-version`) won't throw an error.  It saves us from having task specific arg parsing logic within airflow.

hail_scripts/reference_data/config.py
```diff
@@ -43,6 +58,11 @@ def custom_gnomad_select_v3(ht):
                               0, ht.freq[ht.globals.freq_index_dict['XY-adj']].AC)
     return selects
 
+def custom_screen_select(ht):
```
bpblanken, line 61: I made an attempt to do this enum mapping with a config-based approach.  

Something like:

```
'enum_definitions' : [{
  'dst_field_name': 'regionType_ids',
  'src_field_name': 'target',
  'lookup': SCREEN_REGION_TYPE_LOOKUP,
}]
```

might work?  And it would save us from having both `custom_select` and an `enum_definitions` field in the config.

download_and_create_reference_datasets/v03/update_combined_interval_reference_ht.py
```diff
@@ -0,0 +1,42 @@
+#!/usr/bin/env python3
```
bpblanken, line 1: My first pass here tried to preserve keeping `combined_interval_reference` and `combined_reference` in separate scripts.  I think those scripts are gonna end up looking basically identical though.

hail_scripts/reference_data/config.py
bpblanken: I'm gonna go ahead and implement this.   With fresh eyes I think this is cleaner.

hail_scripts/reference_data/config.py
```diff
@@ -202,8 +203,25 @@ def custom_gnomad_select_v3(ht):
     'screen': {
         '38': {
             'path' : 'gs://seqr-reference-data/GRCh38/ccREs/GRCh38-ccREs.ht',
-            'select': {'region_type': 'target'}
-        }
+            'enum_selects': [{
+                'src': 'target',
+                'dst': 'regionType_ids',
+                'mapping': hl.dict(
```
bpblanken, line 209: A reasonable optimization here is to store just the list (without the `hl.dict` and `hl.enumerate`) so that we can pass just a list to `seqr` on the globals.  the `hl.dict` and `hl.enumerate` would be called at mapping time.

hail_scripts/reference_data/combine_test.py
```diff
@@ -0,0 +1,79 @@
...
+                version='1.2.3', 
+                enum_definitions={
+                    'screen': {
+                        'regionType_ids': {
+                            'CTCF-bound': 0,
+                            'CTCF-only': 1,
+                            'DNase-H3K4me3': 2,
+                            'PLS': 3,
+                            'dELS': 4,
+                            'pELS': 5
```
bpblanken, line 74: missing `'DNase-only', 'low-DNase'`?

hail_scripts/reference_data/combine.py
```diff
@@ -46,6 +47,18 @@ def get_select_fields(selects, base_ht):
                 select_fields[key] = ht
     return select_fields
 
+def get_enum_select_fields(enum_selects, base_ht):
```
bpblanken, line 50: I have (sorta) lost faith that trying to configure the enum selects with a function like this (vs. using a custom select) is worthwhile.  

The more enum selects that get added the more options this function needs to support which is no bueno.  

One potential solution is to do two passes at the select... one which would `custom_select` and handle any field specific parsing and then a second select to do a straightforward enum mapping.

bpblanken: Clinvar can _maybe_ fit into this... it might be _too_ custom though.

hanars: I think this has gotten complex enough that its not super readable, and I think the two pass approach you propose might be the better solution

hail_scripts/reference_data/combine.py
```diff
@@ -40,15 +44,30 @@ def get_select_fields(selects, base_ht):
...
+        if ht[field_name].dtype in ENUM_MAPPABLE_TYPES:
+            enum_select_fields[f'{field_name}_ids'] = ht[field_name].map(lambda x: lookup[x])
+        else:
+            enum_select_fields[f'{field_name}_id'] = lookup[ht[field_name]]
+    return enum_select_fields
+
+
 def get_ht(dataset, reference_genome):
-    ' Returns the appropriate deduped hail table with selects applied.'
+    'Returns the appropriate deduped hail table with selects applied.'
```
bpblanken, line 70: why is this comment just a string???  I didn't even notice... but that's crazy!

bpblanken: I guess there's nothing requiring docstrings to be triple quoted!

hail_scripts/reference_data/config.py
```diff
@@ -272,8 +273,20 @@ def custom_gnomad_select_v3(ht):
     },
     'screen': {
         '38': {
-            'path': 'gs://seqr-reference-data/GRCh38/ccREs/GRCh38-ccREs.ht',
+            'path' : 'gs://seqr-reference-data/GRCh38/ccREs/GRCh38-ccREs.ht',
             'select': {'region_type': 'target'},
+            'enum_select': {
```
bpblanken, line 278: To get camelCase-ing here, we'd need to modify this to be:

```
'select': {'regionType': 'target'},
'enum_select': {'regionType: ['CTCF-bound',...]},
```

which is straightforward... but leaving as is would allow us backwards compatibility w/ v02.

download_and_create_reference_datasets/v03/write_combined_interval_reference_ht.py
```diff
@@ -0,0 +1,60 @@
...
+INTERVAL_REFERENCE_HT_PATH = 'combined_interval_reference/combined_interval_reference.GRCh{genome_version}.ht'
+VERSION = '1.0.0'
+
+def update_existing(destination_path: str, dataset: str, genome_version: str):
+    destination_ht = hl.read_table(destination_path)
+    dataset_ht = get_ht(dataset, genome_version)
+    destination_ht = (destination_ht
+        .drop(dataset)
+        .join(dataset_ht, 'outer')
+        .filter(
```
bpblanken, line 21: I updated this beyond the logic we have in `join_hts` to also filter out rows where every dataset is null.  

I'm making the assumption that, in addition to potentially adding rows with an updated dataset we might also be removing rows.  If the rows we're removing have presence in other datasets we want to keep them, but if they are unique to the old version of the dataset we want to remove them.  Is this assumption correct?

hanars: I'm not sure theres much harm in keeping rows with no datasets, especially if we are running VEP on the whole dataset. @mike-w-wilson do we anticipate any meaningful number of rows ever dropping form a dataset? My gut feeling is that this would be a pretty rare event but maybe I'm wrong

hail_scripts/reference_data/combine.py
```diff
@@ -29,62 +32,93 @@ def get_select_fields(selects, base_ht):
...
+    for field_name, values in enum_selects.items():
+        lookup = hl.dict(hl.enumerate(values, index_first=False))
+        # NB: this conditioning on type is "outside" the hail expression context.
+        if ht[field_name].dtype in ENUM_MAPPABLE_TYPES:
+            enum_select_fields[f'{field_name}_ids'] = ht[field_name].map(lambda x: lookup[x])
+        else:
+            enum_select_fields[f'{field_name}_id'] = lookup[ht[field_name]]
+    return enum_select_fields
+
+def get_ht(dataset: str, reference_genome: str):
```
bpblanken, line 71: Ended up rewriting this function to make the steps more clear.

bpblanken:
This branch is stable again.

hail_scripts/reference_data/combine.py
```diff
@@ -0,0 +1,123 @@
...
+    return ht.select(**{field_name: ht.row.drop(*ht.key)})
+
+def update_joined_ht_globals(
+    joined_ht, datasets, version, coverage_datasets, reference_genome
+):
+    # Track the dataset we've added as well as the source path.
+    included_dataset = {
+        k: v[reference_genome]['path']
+        for k, v in CONFIG.items()
+        if k in datasets + coverage_datasets
```
mike-w-wilson, line 90: What would not meet this condition?

hail_scripts/reference_data/combine.py
```diff
@@ -0,0 +1,123 @@
...
+        k: {enum_field_name: enum_values}
+        for k, v in CONFIG.items()
+        if k in datasets + coverage_datasets
+        if 'enum_select' in v[reference_genome]
+        for enum_field_name, enum_values in v[reference_genome]['enum_select'].items()
+    }
+    # Add metadata, but also removes previous globals.
+    return joined_ht.select_globals(
+        date=datetime.now().isoformat(),
+        datasets=hl.dict(included_dataset),
```
mike-w-wilson, line 102: Do you think it makes sense to add in functionality to get a specific datasets version? I know v02 grabs clinvar's version during the join in the loading pipeline but moving forward with one massive table, we'll want to keep version per dataset. This is probably out of scope for this but just a thought when I saw the globals. It should probably be added to this dict.

hail_scripts/reference_data/config.py
mike-w-wilson: Yeah I think keeping as is makes sense and then we maybe do a single pass of the repo and move to camelCase, similar to what we should do with black/adding docstrings to all funtions? I don't know what the best approach is for supporting v02 vs slightly cleaner v03 repo?

mike-w-wilson (approved):
This looks good to me overall! I have a few questions and opinions(no real strong ones). Most are thoughts on what we else we could do as we update...

hail_scripts/reference_data/combine.py
```diff
@@ -0,0 +1,123 @@
...
+    """
+    Annotates the hail table with the coverage dataset.
+        '<coverage_dataset>': <over_10 field of the locus in the coverage dataset.>
+    :param ht: hail table
+    :param coverage_dataset: coverage dataset e.g. gnomad genomes or exomes coverage
+    :param reference_genome: '37' or '38'
+    :return: hail table with proper annotation
+    """
+    coverage_ht = hl.read_table(CONFIG[coverage_dataset][reference_genome]['path'])
+    return ht.annotate(**{coverage_dataset: coverage_ht[ht.locus].over_10})
```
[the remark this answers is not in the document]
mike-w-wilson: Sorry just saw you do this in another PR.

hail_scripts/reference_data/combine.py
```diff
@@ -0,0 +1,123 @@
...
+            enum_select_fields[f'{field_name}_ids'] = ht[field_name].map(lambda x: lookup[x])
+        else:
+            enum_select_fields[f'{field_name}_id'] = lookup[ht[field_name]]
+    return enum_select_fields
+
+def get_ht(dataset: str, reference_genome: str):
+    config = CONFIG[dataset][reference_genome]
+    field_name = config.get('field_name') or dataset
+    ht = hl.read_table(config['path']).distinct()
+    ht = ht.filter(config['filter'](ht)) if 'filter' in config else ht
```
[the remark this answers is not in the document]
bpblanken: 👍 The `distinct()` used to be on the very last line of this function.  I'll just put it there!

hail_scripts/reference_data/combine.py
bpblanken: Ah.  So `datasets` and `coverage_datasets` are arbitrary lists of reference datasets that are passed in here.  

A cleaner way to write this would be to just iterate over `datasets + coverage_datasets`, but this is preserving the existing logic.

hail_scripts/reference_data/combine.py
bpblanken: Yep!  I think that would be a useful feature.  `v03` has that for clinvar as well.  I think all the other datasets are being treated as statically versioned though.

hanars: I think it would be nice to have a per-dataset version! I don't want to blow up the scope of what you are working on, but that would be really helpful in the future

hanars (approved):

PR #444, opened 2023-04-18
Fix staticmethod

bpblanken:


luigi_pipeline/lib/hail_tasks.py
```diff
@@ -125,8 +125,7 @@ def run_vep(mt, genome_version, runner='VEP', vep_config_json_path=None):
 
         return runners[runner]().run(mt, genome_version, vep_config_json_path=vep_config_json_path)
 
-    @staticmethod
-    def subset_samples_and_variants(mt, subset_path):
+    def subset_samples_and_variants(self, mt, subset_path):
```
bpblanken, line 128: We call this (and also the other functions below) exclusively as an `instance` method.

mike-w-wilson (approved):
LGTM!

PR #447, opened 2023-04-24
Noncoding exon transcript variants

bpblanken:
This was my best attempt at this, let me know if I'm on the right track!


hanars (approved):
The logic looks right to me and will do what we need for seqr to be able to do the searches we want. I would want @mike-w-wilson to take a look to see if he has any suggestions on hail syntax

mike-w-wilson (approved):
This looks good to me --  I think this is the best approach in hail -- any stops looking as soon as it finds a match.

Issue #448, opened 2023-04-26
Add EXON and INTRON fields to sorted transcript consequences


nvnieuwk:
Hi, would it be a possibility to add the EXON and INTRON fields generated by VEP to the output of the loading pipeline? We'd like to add these fields in our institute's version of Seqr (It's a highly requested feature).

I think it should be as simple as adding `'exon'` and `'intron'` to this list: https://github.com/broadinstitute/seqr-loading-pipelines/blob/8e8e28519152256d4dc5c63e87a3463c2ad5e80a/hail_scripts/computed_fields/vep.py#L158. I'll gladly make a PR for this if you think this belongs here.

Is there an interest to add this to this repository or should we make our own fork of the loading pipeline and only add it there?

Thanks!

Greets
-Nicolas

nvnieuwk:
I made a PR containing the suggestions #449

hanars:
Hi Nicolas, thanks for proposing the change and thanks for submitting the PR! Changes to the seqr pipeline are not particularly useful without corresponding changes to the seqr UI to show the new data - how would you propose showing this information to users?

nvnieuwk:
Hi, that makes absolute sense! We would want to add this information to each variant on the variant detail screen. I've not had a thorough look of where would be the best location (but maybe somewhere under or next to the allele counts?), would you have any suggestions where this would belong the best? 

Pinging @tomiles for a second opinion.

hanars:
I think we are just confused what information this is adding, given that we already show a transcript's predicted consequence which will either be explicitly intronic or will be a consequence that is only possible when a variant is exonic.

hanars:
Hi @nvnieuwk any update on how the display for this information would be different/ supplemental to what is already shown to users in seqr?

tomiles:
@hanars This issue for adding VEP exon and intro info that @nvnieuwk raised came from our analysts. There currently is no convenient way to see in which intron or exon a variant is positioned. And this info is more intuitive to indicate where in a gene the variant is positioned without having to first follow the link out to the UCSC browser and check the gene track to find the exon/intron nr.

nvnieuwk:
I'll give it a go! I'm very new to React so any feedback is welcome. I'll ping you when the PR has been opened

nvnieuwk:
[...]

PR #449, opened 2023-04-26
Add exon and intron to transcript consequences

nvnieuwk:
Answer to issue #448


bpblanken:
This looks reasonable to me, but requesting @mike-w-wilson and @hanars  to verify this is ok!

nvnieuwk:
Thank you!

hanars (approved):
Discussed in the ticket and agreed this functionality is worth us upstreaming

bpblanken (approved):

PR #453, opened 2023-04-28
v03 pipeline scaffold

bpblanken:
- Started a new folder to avoid interfering with the existing luigi stuff.
- I'm also unsure if luigi is worth using (at least in the way we've been using it) for the new tasks.  We're going to need a clean way to determine if a task has been completed that _isn't_ just checking for a file existence (for example, the variant annotations table is going to exist both before and after the reference data is updated).  I think luigi can help there but I'm not sure yet!
- If we re-use code I can just import from `luigi_pipeline`.


pyproject.toml
```diff
@@ -11,6 +16,11 @@ profile = 'black'
 sections = ['FUTURE', 'STDLIB', 'THIRDPARTY', 'EXTERNAL_PACKAGE', 'FIRSTPARTY', 'LOCALFOLDER']
 skip_glob = ['download_and_create_reference_datasets/*', 'gcloud_dataproc/*', 'hail_scripts/*', 'kubernetes/*']
 
+[tool.mypy]
```
bpblanken, line 19: Other benefit to just starting a new folder is we can get some foundational pieces like `mypy` in from the start.

Here's a mypy example from typo-ing a str for an enum:

```
bblanken@wmbbe-67c seqr-loading-pipelines % git diff
diff --git a/v03_pipeline/paths.py b/v03_pipeline/paths.py
index 4c3e543..c4fca0c 100644
--- a/v03_pipeline/paths.py
+++ b/v03_pipeline/paths.py
@@ -30,7 +30,7 @@ def project_table_path(
 ) -> str:
     return os.path.join(
         SEQR_SCRATCH_TEMP if env == Env.DEV else SEQR_DATASETS,
-        reference_genome.value,
+        reference_genome,
         V03,
         dataset_type.value,
         'projects',
bblanken@wmbbe-67c seqr-loading-pipelines % mypy
v03_pipeline/paths.py:31: error: Returning Any from function declared to return "str"  [no-any-return]
v03_pipeline/paths.py:31: error: No overload variant of "join" matches argument types "str", "ReferenceGenome", "str", "Union[str, str, str, str]", "str", "str", "str"  [call-overload]
v03_pipeline/paths.py:31: note: Possible overload variants:
v03_pipeline/paths.py:31: note:     def join(str, /, *paths: str) -> str
v03_pipeline/paths.py:31: note:     def join(Union[str, PathLike[str]], /, *paths: Union[str, PathLike[str]]) -> str
v03_pipeline/paths.py:31: note:     def join(Union[bytes, PathLike[bytes]], /, *paths: Union[bytes, PathLike[bytes]]) -> bytes
Found 2 errors in 1 file (checked 4 source files)
```

v03_pipeline/constants.py
```diff
@@ -0,0 +1,23 @@
...
+    SNV = 'SNV'
+    SV = 'SV'
+
+
+class Env(Enum):
+    DEV = 'dev'
+    PROD = 'prod'
+
+
+class ReferenceGenome(Enum):
```
bpblanken, line 21: I might also like `SampleType` here... though we didn't have that in the path in the design doc.

PR #455, opened 2023-04-28
Resolve some reference data eccentricities

bpblanken:
I know quite a bit of pain would be involved to update the existing reference data to modify the structure, but I think this makes things more logical and structured both now and for v03?

EDIT: After discussing with Mike we think the pain is probably worth it.  I think the plan would be to transmute the existing reference tables into the new structure with a one-off script rather than regenerate the whole thing from scratch.


hail_scripts/reference_data/config.py
```diff
@@ -68,7 +68,7 @@ def custom_gnomad_select_v3(ht):
     },
 """
 CONFIG = {
-    '1kg': {  # tgp
+    'tgp': {  # tgp
```
hanars, line 71: I'm pretty sure we don't use this anymore at all in seqr. @lynnpais just confirming that we do not want to show Thousand Genomes frequencies in seqr anymore?

mike-w-wilson: TGP is within gnomAD v3.1 frequencies if that helps make a decision.

mike-w-wilson (changes requested):
This is clean but the coverage HTs are keyed differently so require that join on locus.

hail_scripts/reference_data/combine.py
```diff
@@ -104,19 +94,22 @@ def update_joined_ht_globals(
     )
 
 
-def join_hts(datasets, version, coverage_datasets=[], reference_genome='37'):
+def join_hts(datasets, version, reference_genome='37'):
+
     # Get a list of hail tables and combine into an outer join.
-    hts = [get_ht(dataset, reference_genome) for dataset in datasets]
+    non_coverage_hts = [get_ht(dataset, reference_genome) for dataset in datasets if 'coverage' not in dataset]
```
bpblanken, line 98: Unclear if this is clearer or not. 

Another option here is to `join vs annotate` dependent on whether the key is `locus, alleles` or just `locus`.  That might be more explicit?

bpblanken: this doesn't work on second glance... refactoring!

hail_scripts/reference_data/config.py
lynnpais: Confirming we do no longer want to show or search on thousand genomes frequencies. Thanks.

hanars: okay so we can remove this entirely!

hail_scripts/reference_data/combine.py
```diff
@@ -95,13 +103,32 @@ def update_joined_ht_globals(
...
+    # and alleles, so we cannot join.  Instead we annotate w/ locus as
+    # as the key.
+    coverage_hts = [
+        (dataset, get_ht(dataset, reference_genome))
+        for dataset in datasets
+        if 'coverage' in dataset
+    ]
+    for dataset, coverage_ht in coverage_hts:
+        joined_ht.annotate(dataset=coverage_ht[coverage_ht.locus][dataset])
+
```
mike-w-wilson: This makes sense. I was wondering if we could compare the key values, when they match --  join, when one is a subset of the other annotate, but that seems like over-engineering for something that runs once.

mike-w-wilson (approved):
LGTM!

hail_scripts/reference_data/combine.py
bpblanken: yeah.. I actually thought that approach might be preferable as well.  Do you think that we'll have other datasets down the line that aren't coverage but are keyed only by `locus`?

mike-w-wilson: Nothing immediately comes to mind. The interval ref dataset is basically locus but just a range, `1-10` instead of `1`, `2`, `3`, `4`,... I imagine most datasets would be keyed by ranges and added to the interval dataset.

PR #456, opened 2023-05-02
Combined Reference Data.

bpblanken:


hail_scripts/reference_data/config.py
```diff
@@ -102,31 +112,38 @@ def custom_gnomad_select_v3(ht):
...
-                'SIFT_pred',
-                'Polyphen2_HVAR_pred',
-                'MutationTaster_pred',
-                'FATHMM_pred',
                 'MetaSVM_pred',
                 'REVEL_score',
                 'GERP_RS',
                 'phastCons100way_vertebrate',
             ],
+            'custom_select': dbnsfp_custom_select,
```
bpblanken, line 120: A few questions here:
- `REVEL_score` is a string that we parse to a float [here](https://github.com/broadinstitute/seqr/compare/hail-backend#diff-8fc83668e54552fcb504d1667e3256c36da1ceb8c893b64fad62c2dc0d508479R141).  Should it just be a `float`? 
- `MetaSVM_pred` appears to be an enum, should we also map it?
```
In [10]: ht.aggregate(hl.agg.counter(ht.MetaSVM_pred))
Out[10]: {'D': 13663364, 'T': 68434853, None: 7519568}
```

hail_scripts/reference_data/config.py
```diff
@@ -52,6 +63,10 @@ def custom_gnomad_select_v3(ht):
     )
     return selects
 
+def custom_mpc_select(ht):
+    selects = {}
+    selects['MPC'] = hl.parse_float(ht.info.MPC)
```
bpblanken, line 68: Thoughts on making this `hl.parse_float32`?

hail_scripts/reference_data/combine.py
```diff
@@ -132,3 +141,25 @@ def join_hts(datasets, version, reference_genome='37'):
...
+    if 'coverage' in dataset:
+        joined_ht = (
+            joined_ht.drop(dataset)
+            .join(dataset_ht, 'outer')
+            .filter(
+                hl.any([~hl.is_missing(joined_ht[dataset]) for dataset in datasets]),
+            )
+        )
+    else:
+        joined_ht.annotate(dataset=dataset_ht[dataset_ht.locus][dataset])
```
hanars, line 164: won't this add an annotation called `dataset`, instead of one with the name of the desired dataset?

hail_scripts/reference_data/config.py
hanars: yes lets map it

hail_scripts/reference_data/combine.py
bpblanken: Yep!  This is wrong.

bpblanken: Seems like adding a test for `update_existing_joined_hts` would be worthwhile?

bpblanken: I added a test.. it was an extremely useful exercise.

download_and_create_reference_datasets/v03/write_combined_reference_ht.py
```diff
@@ -0,0 +1,73 @@
...
+        ht = update_existing_joined_hts(
+            destination_path,
+            dataset,
+            DATASETS,
+            VERSION,
+            genome_version,
+        )
+    else:
+        ht = join_hts(DATASETS, VERSION, reference_genome=genome_version)
+    print(f'Uploading ht to {destination_path}')
```
mike-w-wilson, line 50: Same comment on describe as above

download_and_create_reference_datasets/v03/write_combined_interval_reference_ht.py
```diff
@@ -49,16 +24,26 @@ def run(environment: str, dataset: str):
...
+        ht = update_existing_joined_hts(
+            destination_path,
+            dataset,
+            DATASETS,
+            VERSION,
+            genome_version,
+        )
     else:
-        ht = create_new(genome_version)
+        ht = join_hts(DATASETS, VERSION, reference_genome=genome_version)
```
mike-w-wilson, line 35: I know we have tests but I think keeping a ht.describe() is helpful when running this just so the user can visually see the table is what they expect.

hail_scripts/reference_data/combine.py
```diff
@@ -46,7 +48,14 @@ def get_enum_select_fields(enum_selects, ht):
     if enum_selects is None:
         return enum_select_fields
     for field_name, values in enum_selects.items():
-        lookup = hl.dict(hl.enumerate(values, index_first=False))
+        lookup = hl.dict(
+            hl.enumerate(values, index_first=False).extend(
+                # NB: adding missing values here allows us to
+                # hard fail if a mapped key is present but an unexpected value
+                # but also propagate missing values.
```
mike-w-wilson, line 55:
```suggestion
                # hard fail if a mapped key is present and has an unexpected value
                # but propagate missing values.
```

mike-w-wilson: Nit: Slight rewording

download_and_create_reference_datasets/v03/write_combined_interval_reference_ht.py
```diff
@@ -49,16 +24,26 @@ def run(environment: str, dataset: str):
...
     print(f'Uploading ht to {destination_path}')
     write_ht(ht, destination_path)
```
mike-w-wilson, line 37: I don't think this works since you're reading and writing to the same path. Adding a checkpoint to some scratch storage location in the update existing function is probably the easiest way to get this working.

hail_scripts/reference_data/combine.py
```diff
@@ -114,21 +125,43 @@ def join_hts(datasets, version, reference_genome='37'):
...
+    if 'coverage' not in dataset:
+        joined_ht = joined_ht.drop(dataset)
+        joined_ht = joined_ht.join(dataset_ht, 'outer')
+        joined_ht = joined_ht.filter(
+            hl.any([~hl.is_missing(joined_ht[dataset]) for dataset in datasets]),
+        )
+    else:
+        joined_ht = joined_ht.annotate(
+            **{dataset: dataset_ht[joined_ht.locus][dataset]},
+        )
```
mike-w-wilson, line 166: Possibly add a checkpoint here?

mike-w-wilson (changes requested):
So as written, I don't think the update will work since you read and write to the same place which hail doesn't allow. Throwing a checkpoint line into the update function will allow this as written.

download_and_create_reference_datasets/v03/write_combined_interval_reference_ht.py
bpblanken: hrm... I think the combined interval reference doesn't have a version in the path name, but the others do.  Given that we now have individual dataset versions I think we don't need to keep the version in the pathname and we should just be updating a single file?  so we don't have to store a separate copy every time we update?

bpblanken: Which indicates both pathways should checkpoint.

hail_scripts/reference_data/combine.py
bpblanken: I added the checkpoint to the last line here...

I'm actually seeing some strange behavior in the tests here...

Doing

```
print(joined_ht.collect())
joined_ht = joined_ht.checkpoint(checkpoint_path, overwrite=True, stage_locally=True)
print(joined_ht.collect())
```

is giving me different results... with the checkpoint seemingly reverting the applied join.  What am I doing wrong here?

mike-w-wilson: So the first print statement shows the newly added ht annotations and the second one is not?

hail_scripts/reference_data/combine.py
```diff
@@ -164,4 +165,5 @@ def update_existing_joined_hts(
         joined_ht = joined_ht.annotate(
             **{dataset: dataset_ht[joined_ht.locus][dataset]},
         )
-    return update_joined_ht_globals(joined_ht, dataset, version, genome_version)
+    joined_ht = update_joined_ht_globals(joined_ht, dataset, version, genome_version)
+    return joined_ht.checkpoint(checkpoint_path, overwrite=True, stage_locally=True)
```
mike-w-wilson, line 169: I've never tried returning a checkpoint like this. Does this actually checkpoint the table? It makes sense but I wouldn't be surprised if it writes but does not read it back in.

hail_scripts/reference_data/combine.py
bpblanken: yep!  that's exactly it.  I'm gonna come back to this and see if I can re-produce in a shell.

hail_scripts/reference_data/combine.py
bpblanken: Yep!  It does.  The docs actually say say it's  just an alias for `write` and then `read_table`:

https://hail.is/docs/0.2/_modules/hail/table.html#Table.checkpoint

download_and_create_reference_datasets/v03/write_combined_reference_ht.py
```diff
@@ -0,0 +1,81 @@
...
+    'splice_ai',
+    'topmed',
+    'gnomad_genome_coverage',
+    'gnomad_exome_coverage',
+]
+VERSION = '1.0.0'
+
+
+def run(environment: str, genome_version: str, dataset: str):
+    checkpoint_path = os.path.join(
```
bpblanken, line 33: Do we feel like it's better to have the checkpoint paths be named or unnamed?

An alternative here is to just have this be `gs://seqr-scratch-temp/{uuid.uuid4()}.ht`, which may be clearer about what we're actually doing here.

bpblanken: I opted for this... it makes the `update_joined_ht` code cleaner to not have to dependency inject a path.

hanars:
>  we scrap the combined ref and only keep and update the all-variants-we-know-about-from-all-our-callsets-and-reference-data HT.

This is actually an open question we never landed on an answer to. Then last we discussed I took an action item to talk to Tim and determine feasibility, and the result was that it is feasible not to maintain 2 tables. Lets discuss further at Tuesdays meeting

download_and_create_reference_datasets/v03/write_combined_reference_ht.py
```diff
@@ -55,6 +48,9 @@ def run(environment: str, genome_version: str, dataset: str):
     else:
         ht = join_hts(DATASETS, VERSION, reference_genome=genome_version)
     ht.describe()
+    checkpoint_path = f"{GCS_PREFIXES[('dev', AccessControl.PUBLIC)]}/{uuid.uuid4()}.ht"
```
mike-w-wilson, line 51: Purely style choice but I typically like path defined up top so if we ever update them, we know where to look but this is totally opinion.

mike-w-wilson (approved):
:shipit:

download_and_create_reference_datasets/v03/write_combined_reference_ht.py
bpblanken: Yeah that makes sense.  

I think this is pretty jank as is (`write_ht` should probably just do the checkpointing internally), but can fix downstream.

PR #457, opened 2023-05-02
Implement prev_call for sv callset

bpblanken:
Waiting for a ticket to be sure... but I think this is on the right track!


hanars (approved):
this looks right to me

bpblanken:
@hanars I just left a comment on the ticket concerning overlapping vs identical.  Are we still ok?

PR #461, opened 2023-05-03
account for end locus when joining SV variants and genotype

bpblanken:


luigi_pipeline/lib/model/sv_mt_schema.py
```diff
@@ -61,11 +61,12 @@ def __init__(self, *args, gene_id_mapping=None, **kwargs):
         super().__init__(*args, **kwargs)
         self._gene_id_mapping = gene_id_mapping
 
+    @row_annotation()
```
bpblanken, line 64: The idea here is to make `end_locus` a real annotation and then add it to the matrix table's `key` when we export the `Variants` mt.  The `key` will persist when we create the `Genotypes` mt. 

Then, when we elastic search export (which joins the `Variants` and `Genotypes` tables), the `end_locus` field will be dropped alongside `locus` and `alleles`.

We actually do have test coverage to verify that the exported variant fields has not changed ([here's](https://github.com/broadinstitute/seqr-loading-pipelines/pull/461/files#diff-51c65deaa50a78ae1b01e0b6b370d1186ba53f59842232a1372f283eef64e689R331-R362) the list).

bpblanken: Also.... is `end_locus` even the right field to add to the key?

hanars: Exact coordinates for SVs are not precise enough to use as keys, and given that the start position isn't part of the key it seems pretty counterintuitive for the end position to be part of the key. I really think the unique string identifier that we currently use as the key should be the only field we key by

luigi_pipeline/seqr_sv_loading.py
```diff
@@ -37,6 +37,13 @@ class SeqrSVVariantMTTask(SeqrVCFToVariantMTTask):
     # NB: electing not to override import_vcf here eventhough the inherited args are slightly different
     # than from the old pipeline.
 
+    def import_dataset(self):
+        mt = self.import_vcf()
+        return mt.key_rows_by('rsid')
```
bpblanken, line 42: If my understanding of all of this is correct, `rsid` will persist when writing the Variants and Genotypes mt even though it's not included in the annotations.

luigi_pipeline/seqr_sv_loading.py
```diff
@@ -37,6 +37,13 @@ class SeqrSVVariantMTTask(SeqrVCFToVariantMTTask):
...
+
+    def annotate_old_and_split_multi_hts(self, mt):
```
bpblanken, line 44: `hl.split_multi_hts` isn't called in the old pipeline.

luigi_pipeline/seqr_sv_loading.py
hanars: So we want this keyed by `variantId`, which is derived from `rsid` but we don't want that duplicated. I think what we would want is `key_rows_by(variantId=mt.rsid)`

luigi_pipeline/seqr_sv_loading.py
hanars: might be worth either a) renaming this in the base class to something more generic and/pr b) adding a comment here to explain that annotating old and solitting multi are not needed for SVs

luigi_pipeline/seqr_sv_loading.py
bpblanken: Do we want to drop `variantId` when we export?  I think current behavior is to drop whatever we're using as keys when we export to elastic search.  

We also have [this annotation](https://github.com/broadinstitute/seqr-loading-pipelines/blob/master/luigi_pipeline/lib/model/sv_mt_schema.py#L211-L213) which we would need to remove if we're doing the annotation here.

hanars (approved):

PR #462, opened 2023-05-03
Benb/barebones ci

bpblanken:
Bringing this back up to date... and branched off of `master` for easier merge.


.github/workflows/unit-tests.yml
```diff
@@ -0,0 +1,41 @@
...
+      - name: Check Import Sort
+        run: isort . --check --diff
+      - name: Check Black
+        run: black . --check --diff
+      - name: Check Ruff
+        run: ruff . --format github
+      - name: Run Coverage Tests
+        run: |
+          nosetests --with-coverage --cover-package hail_scripts hail_scripts
+          coverage report --omit '*test*' --fail-under=40
```
bpblanken, line 41: These tests run in <30s, the `SV` & `GCNV` tests are still over 60s apiece.  

I think running some tests is better than no tests here, but there's a strong case that leaving tests out on purpose is going to confuse us down the line.

bpblanken: Ticket here: https://github.com/broadinstitute/seqr-loading-pipelines/issues/464

.github/workflows/unit-tests.yml
```diff
@@ -0,0 +1,41 @@
...
+        with:
+          python-version: ${{ matrix.python-version }}
+          cache: 'pip'
+          cache-dependency-path: 'requirements**.txt'
+      - name: Install dependencies
+        run: |
+          python -m pip install --upgrade pip
+          pip install -r requirements.txt
+          pip install -r requirements-dev.txt
+      - name: Check Import Sort
```
bpblanken, line 32: I think all of these tools have accompanying GitHub actions that manage their install.  I opted to use the command line checks because 1) we have to install the `dev` requirements anyways for them to work correctly. 2) The version of the tools themselves can be managed with `requirements-dev.in`.

hanars:
I would recommend having @sjahl review any CI work

docker/Dockerfile
```diff
@@ -109,7 +109,7 @@ COPY hail_builds/ ./hail_builds
 COPY hail_scripts/ ./hail_scripts
 COPY luigi_pipeline/ ./luigi_pipeline
 
-RUN python3 -m pip install -r ./luigi_pipeline/requirements.txt
+RUN python3 -m pip install -r ./requirements.txt
```
sjahl, line 112: I would see if you can try copying in the requirements.txt file alone and running this step as early on in the dockerfile as our installation process allows. Assuming the requirements.txt changes less often than the hail_* and luigi_* directories, we'll have to run this pip install any time something in those folders changes.

If we copy in the requirements.txt into the image individually earlier in the dockerfile, we can take advantage of more cache hits in the docker build, and only run that pip install step when the requirements.txt file changes.

bpblanken: 👍

bpblanken: I lifted up the `pip install` to be before copying files from the repo into the docker build.  The dockerfile can probably use quite a bit more tlc but it's probably out of scope for this pr.

.github/workflows/unit-tests.yml
sjahl: Yeah, we frequently end up shelling out here for reasons similar to the ones you mentioned; i'm fine with this approach

docker/Dockerfile
sjahl: Yeah, sounds good; only happened to notice this line since it appeared in the diff here 👍

sjahl (approved):
The CI and pip-tools bits look fine to me here, just left a few comments inline. I can't really speak for the tests themselves, so I'll leave that up to someone with more experience on the pipeline(s)

.github/workflows/unit-tests.yml
bpblanken:
```
Coverage failure: total of 38 is less than fail-under=40
```

.github/workflows/unit-tests.yml
```diff
@@ -0,0 +1,41 @@
...
+        run: isort . --check --diff
+      - name: Check Black
+        run: black . --check --diff
+      - name: Check Ruff
+        run: ruff . --format github
+      - name: Run Coverage Tests
```
bpblanken, line 38: Removing the tests after discussions with @mike-w-wilson

PR #465, opened 2023-05-08
Mw/move validate

mike-w-wilson:
This moves validate MT and contig check up a class to HailMatrixTableTask. This allows the sex inference code to inherit from somewhere that makes sense. This code also adds the ability for the pipeline to ingest VCFs that are split into two parts, genotypes and filters, which is the way DSP delivers our exome VCFs. I can move this out if youd like but seemed like a clean enough addition to have here?

### DIFF
--- a/luigi_pipeline/lib/hail_tasks.py
+++ b/luigi_pipeline/lib/hail_tasks.py
@@ -1,11 +1,11 @@
 """
 Tasks for Hail.
 """
-from collections import Counter
 import json
 import logging
 import math
 import os
+from collections import Counter
 
 import hail as hl
 import luigi
@@ -13,8 +13,9 @@ from luigi.contrib import gcs
 from luigi.parameter import ParameterVisibility
 
 from hail_scripts.elasticsearch.hail_elasticsearch_client import HailElasticsearchClient
-from lib.global_config import GlobalConfig
-import lib.hail_vep_runners as vep_runners
+
+import luigi_pipeline.lib.hail_vep_runners as vep_runners
+from luigi_pipeline.lib.global_config import GlobalConfig
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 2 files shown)


luigi_pipeline/lib/hail_tasks.py
```diff
@@ -84,12 +96,52 @@ def import_vcf(self):
...
+            reference_genome=self.genome_version,
+            skip_invalid_loci=True,
+            contig_recoding=recode,
+            force_bgz=True,
+            min_partitions=500,
+        )
+        if self.wes_filter_source_paths:
+            filters_ht = hl.import_vcf(
+                self.wes_filter_source_paths,
+                reference_genome=self.genome_version,
```
bpblanken, line 110: Looks like this line changed from the below: (`reference_genome='GRCh' + self.genome_version` -> `reference_genome=self.genome_version`).

Should we be using `hail_utils.py`'s version of `import_vcf` here anyways?

bpblanken: (I meant to comment this a few lines up, but it's the same in both places).

luigi_pipeline/lib/hail_tasks.py
```diff
@@ -45,6 +54,9 @@ class HailMatrixTableTask(luigi.Task):
     """
 
     source_paths = luigi.Parameter(description='Path or list of paths of VCFs to be loaded.')
+    wes_filter_source_paths = luigi.OptionalParameter(
```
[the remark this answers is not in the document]
bpblanken: Actually looks like there's a [ListParameter](https://luigi.readthedocs.io/en/stable/api/luigi.parameter.html#luigi.parameter.ListParameter) option we might want to use... though we'd need to validate whether it works with what we've done in airflow.

PR #467, opened 2023-05-09
`SeqrSchema` is unused

bpblanken:
`SeqrSchema` is only used a single time when it is inherited by `SeqrVariantSchema` (which is widely used).  

Just cleaning up stuff as I re-think the annotations for `v03`.


luigi_pipeline/lib/model/seqr_mt_schema.py
```diff
@@ -149,7 +149,7 @@ def clinvar(self):
     def dbnsfp(self):
         return self._selected_ref_data.dbnsfp
 
-class SeqrSchema(BaseSeqrSchema):
+class SeqrVariantSchema(BaseSeqrSchema):
```
bpblanken, line 152:
```
bblanken@wmbbe-67c seqr-loading-pipelines % grep 'SeqrSchema' * -r | grep -v Base
bblanken@wmbbe-67c seqr-loading-pipelines %
```

PR #469, opened 2023-05-12
Pedigree Utilities for `v03` pipeline.

bpblanken:
Some utilities to read a pedigree file and find the samples and families of interest against a the `mt.cols()` of an imported vcf.


pyproject.toml
```diff
@@ -32,6 +32,12 @@ skip_glob = [
     'kubernetes/*'
 ]
 
+[tool.mypy]
```
bpblanken, line 35: I've been running this locally... I think it'll have limited utility for the time being because neither luigi nor hail has the right support.

v03_pipeline/lib/misc/pedigree.py
```diff
@@ -0,0 +1,39 @@
+import hail as hl
+
+
+def import_pedigree(pedigree_path: str) -> hl.Table:
+    ht = hl.import_table(
+        pedigree_path,
+        types={
```
hanars, line 7: dpesn't hail default imputed types to str?

PR #476, opened 2023-05-16
Individual Reference Dataset Versions.

bpblanken:


hail_scripts/reference_data/combine.py
```diff
@@ -8,6 +8,15 @@
 from hail_scripts.reference_data.config import CONFIG
 
 
+def annotate_coverage(joined_ht: hl.Table, dataset_ht: hl.Table, dataset: str):
```
bpblanken, line 11: This function makes its triumphant return 🤣.

hail_scripts/reference_data/combine.py
```diff
@@ -81,34 +89,31 @@ def get_ht(dataset: str, reference_genome: str):
         },
     )
     ht = ht.transmute(**get_enum_select_fields(config.get('enum_select'), ht))
-    return ht.select(**{field_name: ht.row.drop(*ht.key)}).distinct()
+    ht = ht.select_globals(
+        **{
+            f'{dataset}_globals': hl.struct(
+                path=config['path'],
+                version=ht.globals.get(
```
bpblanken, line 96: idea here is to take the `version` off of the globals if it exists, then take it from the `config`.

hail_scripts/reference_data/combine.py
```diff
@@ -81,34 +89,31 @@ def get_ht(dataset: str, reference_genome: str):
         },
     )
     ht = ht.transmute(**get_enum_select_fields(config.get('enum_select'), ht))
-    return ht.select(**{field_name: ht.row.drop(*ht.key)}).distinct()
+    ht = ht.select_globals(
+        **{
+            f'{dataset}_globals': hl.struct(
```
bpblanken, line 94: I lifted the globals annotations up into `get_ht` because `version` needs to happen here and I wanted them all to be in the same place.  The `join` we do to combine the datasets also joins the globals, so we don't need to do a second step to combine the config.  Lastly, I think there is an existing bug where if we update two individual datasets together in code, and then run the code sequentially, the globals will reflect the in-code update before the second individual dataset is updated.

The downside of this approach is the naming of the variable.  We can't just name it `dataset` because we'll get a naming conflict with the row field.  And we also cannot nest inside a `datasets=hl.Struct(dataset=...)` struct as the join will mangle `datasets` into `datasets_1`, `datasets_2`, etc.

hail_scripts/reference_data/combine.py
```diff
@@ -81,34 +89,31 @@ def get_ht(dataset: str, reference_genome: str):
...
+                path=config['path'],
+                version=ht.globals.get(
+                    'version',
+                    config.get('version', hl.missing(hl.tstr)),
+                ),
+                enum_definitions=config.get(
```
bpblanken, line 100: Maybe name this `enum_select`?

hail_scripts/reference_data/combine.py
hanars: whats the situation where the existing global would be different from what is specified in the config?

bpblanken: I think it'll only happen if 1) human doesn't put a version tag in the config or puts the incorrect one OR 2) we have a dataset with a static path but a changing version.  I think neither might come into play?

hanars: I feel like since these are not expected cases its better to throw an error if we don't have the value we expect, but not have a graceful fallback mechanism for it

bpblanken: 👍 makes sense to me!

hail_scripts/reference_data/combine.py
```diff
@@ -156,12 +167,13 @@ def update_existing_joined_hts(
     dataset_ht = get_ht(dataset, genome_version)
     if 'coverage' not in dataset:
         joined_ht = joined_ht.drop(dataset)
+        joined_ht = joined_ht.select_globals(
+            **joined_ht.globals.drop(f'{dataset}_globals'),
+        )
```
mike-w-wilson, line 172:
```suggestion
        joined_ht = joined_ht.drop(f'{dataset}_globals')
```
I may be missing something but you could just do this to drop the global

bpblanken: Yes that is WAY better.  🤦

hail_scripts/reference_data/config.py
```diff
@@ -89,24 +121,48 @@ def custom_mpc_select(ht):
...
         '38': {
+            'version': 'v1.6',
             'path': 'gs://seqr-reference-data/GRCh38/CADD/CADD_snvs_and_indels.v1.6.ht',
             'select': ['PHRED'],
         },
     },
+    'clinvar': {
+        '37': {
+            'custom_import': download_and_import_latest_clinvar_vcf,
+            'path': 'ftp://ftp.ncbi.nlm.nih.gov/pub/clinvar/vcf_GRCh37/clinvar.vcf.gz',
```
mike-w-wilson, line 144: Do you think we need a `source_file` or something in the config for storing the ingested data  paths and we could keep ht as the ingested HT? Just think all of these datasets come to us as not-HTs, either VCFs or TSVs. This keeps things special for clinvar and hgmd but potentially well want to update all references at some point.

mike-w-wilson (commented):
Just a couple questions and suggestiongs!

download_and_create_reference_datasets/v03/write_clinvar_ht.py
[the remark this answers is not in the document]
mike-w-wilson: Oops -- meant to delete this -- saw the custom import later!

download_and_create_reference_datasets/v03/write_combined_interval_reference_ht.py
```diff
@@ -29,11 +28,10 @@ def run(environment: str, dataset: str):
             destination_path,
             dataset,
             DATASETS,
-            VERSION,
             genome_version,
         )
     else:
-        ht = join_hts(DATASETS, VERSION, reference_genome=genome_version)
+        ht = join_hts(DATASETS, reference_genome=genome_version)
```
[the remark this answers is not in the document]
bpblanken: Yes, I actually have already ported some of this into the `v03`'s `ReferenceGenome` enum on another branch somewhere, so I'll leave it as is here.

hail_scripts/reference_data/clinvar.py
```diff
@@ -117,19 +112,19 @@ def download_and_import_latest_clinvar_vcf(
...
-    mt = import_vcf(
-        tmp_file.name,
-        genome_version,
-        drop_samples=True,
-        min_partitions=2000,
-        skip_invalid_loci=True,
-        more_contig_recoding=mt_contig_recoding,
-    )
-    return mt.annotate_globals(version=clinvar_release_date)
+    with tempfile.NamedTemporaryFile(suffix='.vcf.gz', delete=False) as tmp_file:
```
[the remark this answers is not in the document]
bpblanken: Yeah... this is a tricky one.  It does conflict with hail's lazy eval, since by the time hail actually does its stuff, the file has been deleted.

hail_scripts/reference_data/combine.py
bpblanken: [:multi](https://media.giphy.com/media/WiCO2uZK05Klc1d28q/giphy.gif)

bpblanken: Alt Text

hail_scripts/reference_data/config.py
bpblanken: Yeah, that's a good point.  I've been thinking that `path` could be overloaded, but maybe making it `source_path` is more clear.  And then we can fix up the other reference datasets to align as we go forwards.

bpblanken: I updated to use `source_path`.

hail_scripts/reference_data/combine.py
```diff
@@ -110,8 +110,8 @@ def get_ht(dataset: str, reference_genome: str):
     ht = ht.select_globals(
         **{
             f'{dataset}_globals': hl.struct(
-                path=config['path'],
-                version=parse_version(ht, config),
+                path=config['source_path'] if 'custom_import' in config config['path'],
```
mike-w-wilson, line 113: Do this need an else?
```suggestion
                path=config['source_path'] if 'custom_import' in config else config['path'],
```

mike-w-wilson (approved):
Just one thing and then LGTM!

PR #487, opened 2023-05-23
Benb/unshim the annotations

bpblanken:


v03_pipeline/lib/annotations/annotate_all.py
```diff
@@ -1,20 +1,97 @@
...
+        [
+            snv.original_alt_alleles,
+            shared.pos,
+            shared.rg37_locus,
+            shared.sorted_transcript_consequences,
+            shared.variant_id,
+            shared.xpos,
+        ],
+    ],
+    DatasetType.MITO: [
```
bpblanken, line 22: Potential option for the common simple cases:

```
def identity(field_name: str) -> Callable:
	def f(mt: hl.MatrixTable) -> hl.Expression:
		return mt[field_name]
	f.__name__ == field_name
	return f

def rename(new_field_name: str, old_field_name: str) -> Callable:
	def f(mt: hl.MatrixTable) -> hl.Expression:
		return mt[old_field_name]
	f.__name__ == new_field_name
	return f

rename('AC', 'AC_hom')
```

PR #489, opened 2023-05-24
Add clinvar to combined reference

bpblanken:


hail_scripts/reference_data/clinvar.py
```diff
@@ -117,19 +112,19 @@ def download_and_import_latest_clinvar_vcf(
...
-    mt = import_vcf(
-        tmp_file.name,
-        genome_version,
-        drop_samples=True,
-        min_partitions=2000,
-        skip_invalid_loci=True,
-        more_contig_recoding=mt_contig_recoding,
-    )
-    return mt.annotate_globals(version=clinvar_release_date)
+    with tempfile.NamedTemporaryFile(suffix='.vcf.gz', delete=False) as tmp_file:
```
bpblanken, line 116: Thinking about `tempfile` cleanup was painful here so we're back to leaving a file behind.  Maybe the cleaner answer here is to checkpoint to `gs://seqr-scratch-temp`?  

I played around with `hl.import_vcf('https://ftp.ncbi.nlm....'` which appears to work before it doesn't.  That would be quite nice in the future!

download_and_create_reference_datasets/v03/write_hgmd_ht.py
```diff
@@ -2,43 +2,23 @@
...
 def run(environment: str, genome_version: str):
     dataset = 'hgmd'
-    source_path = CONFIG[dataset][genome_version]['path']
     destination_path = os.path.join(
         GCS_PREFIXES[(environment, AccessControl.PRIVATE)],
         HGMD_HT_PATH,
     ).format(
         genome_version=genome_version,
     )
-    mt = import_vcf(
```
bpblanken, line 28: I went ahead and ported `hgmd` into the `config.py` framework so we can just call `get_ht` here rather than a subset of the functions that `get_ht` calls.

hail_scripts/reference_data/config.py
```diff
@@ -13,6 +25,26 @@ def predictor_parse(field: hl.StringExpression):
     return field.split(';').find(lambda p: p != '.')
 
 
+def clinvar_custom_select(ht):
+    selects = {}
+    clnsigs = parsed_clnsig(ht)
+    selects['pathogenicity'] = hl.if_else(
```
bpblanken, line 31: The idea here was to split the string parsing and enum mapping apart so that the framework enum mapping could do its job here.  So we have the step here which first selects `pathogenicity` and `assertion` as a `hl.tstr` or `hl.tarray(hl.tstr)`, then the enum mapping step which will map to `pathogenicity_id` and `assertion_ids`.  

`conflictingPathogenicities` is a special snowflake that both parses and enum maps together and I left it as such with a comment.

hail_scripts/reference_data/config.py
```diff
@@ -106,6 +138,28 @@ def custom_mpc_select(ht):
...
+            'custom_import': download_and_import_latest_clinvar_vcf,
+            'path': 'ftp://ftp.ncbi.nlm.nih.gov/pub/clinvar/vcf_GRCh37/clinvar.vcf.gz',
+            'select': {'alleleId': 'info.ALLELEID'},
+            'custom_select': clinvar_custom_select,
+            'enum_select': {
+                'pathogenicity': CLINVAR_PATHOGENICITIES,
+                'assertion': CLINVAR_ASSERTIONS,
+            },
+        },
+        '38': {
```
hanars, line 152: Out of scope for MVP, but at some point its probably worth looking into the fact that for many of these the configs 37 and 38 are the same except for the file path.

hanars (commented):
Looks good to me, but I would wait for Mike's review

mike-w-wilson (approved):
This looks good to me. One question on your checkpoint remark and a couple path updates if you feel like it!

hail_scripts/reference_data/clinvar.py
hanars: Local installs never run the reference data creation, because we make the output publicly available and they just download that, so thats not something we need to design for

hail_scripts/reference_data/config.py
```diff
@@ -158,6 +212,7 @@ def custom_mpc_select(ht):
     },
     'hgmd': {
         '37': {
+            'custom_import': download_and_import_hgmd_vcf,
             'path': 'gs://seqr-reference-data-private/GRCh37/HGMD/HGMD_Pro_2022.4_hg19.vcf.gz',
```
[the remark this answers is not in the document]
bpblanken: Might as well!

Issue #415, opened 2023-03-22
Add single main file that replaces `seqr_loading_optimized`, `seqr_mito_loading`, `seqr_sv_loading`, `seqr_gcnv_loading` in airflow deployment.


bpblanken:

bpblanken:
Completed by v03_pipeline.

PR #491, opened 2023-06-01
V3 Pipeline MVP.

bpblanken:


v03_pipeline/lib/misc/sample_ids.py
```diff
@@ -0,0 +1,73 @@
...
+        super().__init__(message)
+        self.missing_samples = missing_samples
+
+
+def vcf_remap(mt: hl.MatrixTable) -> hl.MatrixTable:
+    # TODO: add logic from Mike to remap vcf samples delivered from Broad WGS
+    return mt
+
+
+def remap_sample_ids(mt: hl.MatrixTable, project_remap_ht: hl.Table) -> hl.MatrixTable:
```
bpblanken, line 17: These two functions are copied verbatim from the `luigi_pipeline`, _except_ I remove the line that removed variants with only reference data calls inside the `subset_samples` function.  Needed to keep the row (for sure) inside the `sample_lookup_table` and I felt we might as well keep the row inside the `variant_annotations_table` as well for simplicity.

v03_pipeline/lib/annotations/fields.py
```diff
@@ -0,0 +1,77 @@
...
+        sample_lookup_table.AF,
+        sample_lookup_table.AN,
+    ],
+    (DatasetType.SNV, AnnotationType.GENOTYPE_ENTRIES): [
+        snv.gq,
+        snv.ab,
+        snv.dp,
+        shared.sample_id,
+    ],
+    (DatasetType.MITO, AnnotationType.FORMATTING): [
```
bpblanken, line 41: The non-SNV annotations are partially implemented.

v03_pipeline/lib/annotations/fields.py
```diff
@@ -0,0 +1,77 @@
...
+
+
+def get_fields(
+    t: hl.Table | hl.MatrixTable,
+    annotation_type: AnnotationType,
+    **kwargs: Any,
+) -> dict[str, hl.Expression]:
+    dataset_type = kwargs['dataset_type']
+    fields = {
+        field_expression.__name__: field_expression(t, **kwargs)
```
bpblanken, line 71: Important (and maybe controversial) design choice here to have all of the annotation functions accept all of the task arguments (`DatasetType`, `Env`, `ReferenceGenome`, `liftover_ref_path` etc.).  Only a few of the functions (`rg37_locus`, `hgmd`, and the interval tables) need additional args...  so we're maybe over-designing for the special cases.

v03_pipeline/lib/tasks/base/base_pipeline_task.py
```diff
@@ -0,0 +1,43 @@
...
+        description='Networked temporary directory used by hail for temporary file storage. Must be a network-visible file path.',
+    )
+
+    def output(self) -> luigi.Target:
+        raise NotImplementedError
+
+    def complete(self) -> bool:
+        raise NotImplementedError
+
+    def run(self) -> None:
```
bpblanken, line 23: This is the `run` method which each task will inherit from.  The general flow is for every task to define an `initialize_table` function, which will run if the output does not exist and then an `update` to actually perform the work.

v03_pipeline/lib/tasks/base/base_variant_annotations_table.py
```diff
@@ -0,0 +1,94 @@
...
+)
+from v03_pipeline.lib.tasks.base.base_pipeline_task import BasePipelineTask
+from v03_pipeline.lib.tasks.files import (
+    GCSorLocalFolderTarget,
+    GCSorLocalTarget,
+    HailTableTask,
+)
+
+
+class BaseVariantAnnotationsTableTask(BasePipelineTask):
```
bpblanken, line 20: Both `UpdateVariantAnnotationsTableWithNewSamplesTask` and the (not yet written) `UpdateVariantAnnotationsTableWithNewReferenceDatasetTask` will share this code.

v03_pipeline/lib/tasks/base/base_variant_annotations_table.py
```diff
@@ -0,0 +1,94 @@
...
+            )
+        else:
+            ht = hl.read_table(
+                reference_dataset_collection_path(
+                    self.env,
+                    self.reference_genome,
+                    self.dataset_type.base_reference_dataset_collection,
+                ),
+            )
+            ht = ht.annotate(
```
bpblanken, line 78: I annotated the test reference data with `vep` under the assumption that we'd be doing that... but I think per discussion we can assume it will be unannotated for now and the `vep` annotation will happen here.

v03_pipeline/lib/tasks/files.py
```diff
@@ -0,0 +1,38 @@
...
+
+
+class RawFileTask(luigi.Task):
+    pathname = luigi.Parameter()
+
+    def output(self) -> luigi.Target:
+        return GCSorLocalTarget(self.pathname)
+
+
+class VCFFileTask(RawFileTask):
```
bpblanken, line 28: I would like :eyes: on this because I do not understand shared vcfs well.

v03_pipeline/lib/annotations/gcnv.py
```diff
@@ -0,0 +1,30 @@
...
+from typing import Any
+
+import hail as hl
+import pytz
+
+from hail_scripts.computed_fields import variant_id as expression_helpers
+
+
+def xpos(ht: hl.Table, **_: Any) -> hl.Expression:
+    # NB: `pos` is an aggregated field over samples.
```
bpblanken, line 13: I think this comment is probably wrong in hindsight... the genotype entries annotations require a `hl.MatrixTable` so we should just go with `hl.MatrixTable`.   The conflict arrises when we join to other tables and need to know whether to reference `key` or `row_key`.

v03_pipeline/bin/run_task.py
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/env python3
```
bpblanken, line 1: This resolves #415 (at least going forwards).

Example usage:

```
bblanken@wmbbe-67c seqr-loading-pipelines % ./v03_pipeline/bin/run_task.py WriteFamilyTableTask --env DEV --reference-genome GRCh37 --dataset-type SNV --sample-type WES --callset-path sample.vcf
Traceback (most recent call last):
  File "./v03_pipeline/bin/run_task.py", line 10, in <module>
    luigi.run() or sys.exit(1)
  File "/Users/bblanken/.pyenv/versions/3.8.10/lib/python3.8/site-packages/luigi/interface.py", line 194, in run
    luigi_run_result = _run(*args, **kwargs)
  File "/Users/bblanken/.pyenv/versions/3.8.10/lib/python3.8/site-packages/luigi/interface.py", line 211, in _run
    return _schedule_and_run([cp.get_task_obj()], worker_scheduler_factory)
  File "/Users/bblanken/.pyenv/versions/3.8.10/lib/python3.8/site-packages/luigi/cmdline_parser.py", line 114, in get_task_obj
    return self._get_task_cls()(**self._get_task_kwargs())
  File "/Users/bblanken/.pyenv/versions/3.8.10/lib/python3.8/site-packages/luigi/task_register.py", line 88, in __call__
    param_values = cls.get_param_values(params, args, kwargs)
  File "/Users/bblanken/.pyenv/versions/3.8.10/lib/python3.8/site-packages/luigi/task.py", line 420, in get_param_values
    raise parameter.MissingParameterException("%s: requires the '%s' parameter to be set" % (exc_desc, param_name))
luigi.parameter.MissingParameterException: WriteFamilyTableTask[args=(), kwargs={'env': <Env.DEV: 'dev'>, 'reference_genome': <ReferenceGenome.GRCh37: 'GRCh37'>, 'dataset_type': <DatasetType.SNV: 'SNV'>, 'sample_type': <SampleType.WES: 'WES'>, 'callset_path': 'sample.vcf'}]: requires the 'project_remap_path' parameter to be set
```

v03_pipeline/lib/annotations/fields.py
hanars: I think having them passed as keyword args is a reasonable compromise for making the calling logic straightforward, but the functions that do not actually need them can simply pass them through

v03_pipeline/lib/annotations/fields.py
```diff
@@ -0,0 +1,78 @@
...
+    (DatasetType.SNV, AnnotationType.REFERENCE_DATASET_COLLECTION): [
+        reference_dataset_collection.hgmd,
+        reference_dataset_collection.gnomad_non_coding_constraint,
+        reference_dataset_collection.screen,
+    ],
+    (DatasetType.SNV, AnnotationType.SAMPLE_LOOKUP_TABLE): [
+        sample_lookup_table.AC,
+        sample_lookup_table.AF,
+        sample_lookup_table.AN,
+        sample_lookup_table.homozygote_count,
```
hanars, line 34: it might be neater to name this field `Hom` in the lookup table/ annotation

v03_pipeline/lib/annotations/gcnv.py
```diff
@@ -0,0 +1,30 @@
...
+        hl.locus(
+            expression_helpers.replace_chr_prefix(ht.chr),
+            ht.pos,
+        ),
+    )
+
+
+def variant_id(ht: hl.Table, **_: Any) -> hl.Expression:
+    return hl.format(
+        f'%s_%s_{datetime.datetime.now(tz=pytz.timezone("US/Eastern")).date():%m%d%Y}',
```
hanars, line 27: I think now that the files will be delivered with variant IDs re-normalized to the old set of IDs, we do not want to add a timestamp anymore

v03_pipeline/lib/annotations/reference_dataset_collection.py
```diff
@@ -0,0 +1,96 @@
...
+
+
+def hgmd(
+    ht: hl.Table,
+    env: Env,
+    reference_genome: ReferenceGenome,
+    dataset_type: DatasetType,
+    **_: Any,
+) -> hl.Expression:
+    if (
```
hanars, line 23: This feels like you've copy-pasted the checking in all the functions here. Maybe make `reference_dataset_collection_path` into `valid_reference_dataset_collection_path` and have the helper function handle the checking?

v03_pipeline/lib/annotations/reference_dataset_collection.py
```diff
@@ -0,0 +1,96 @@
...
+        ReferenceDatasetCollection.INTERVAL_REFERENCE
+        not in dataset_type.annotatable_reference_dataset_collections(env)
+    ):
+        return None
+    interval_reference_ht = hl.read_table(
```
hanars, line 50: I feel like we only want to read the table in once and annotate everything that comes off of it, instead of reading it in for every annotation

v03_pipeline/lib/annotations/shared.py
```diff
@@ -0,0 +1,145 @@
...
+            ),
+        ).filter(lambda c: c.consequence_term_ids.size() > 0),
+        lambda c: (
+            hl.bind(
+                lambda is_coding, is_most_severe, is_canonical: (
+                    hl.cond(
+                        is_coding,
+                        hl.cond(
+                            is_most_severe,
+                            hl.cond(is_canonical, 1, 2),
```
hanars, line 123: I feel like using constants for these magic numbers would be more readable/maintainable

v03_pipeline/lib/annotations/sv.py
```diff
@@ -0,0 +1,11 @@
+from __future__ import annotations
+
+from typing import TYPE_CHECKING, Any
+
+# TODO: remove me once hail is used.
+if TYPE_CHECKING:
```
hanars, line 6: ?

v03_pipeline/lib/misc/sample_entries.py
```diff
@@ -0,0 +1,56 @@
...
+            },
+            sample_id=sample_id,
+        )
+        for sample_id in first_sample_ids_row[0]
+    ]
+
+
+def globalize_sample_ids(ht: hl.Table) -> hl.Table:
+    ht = ht.annotate_globals(
+        sample_ids=ht.aggregate(hl.agg.take(ht.entries.sample_id, 1)[0]),
```
hanars, line 23: wouldn't this only take the first sample_id?

v03_pipeline/lib/misc/sample_lookup.py
```diff
@@ -0,0 +1,65 @@
...
+
+
+def homozygote_count(sample_lookup_ht) -> hl.Int32Expression:
+    return sample_lookup_ht.hom_samples.length()
+
+
+def compute_sample_lookup_ht(mt: hl.MatrixTable) -> hl.Table:
+    sample_ids = hl.agg.collect_as_set(mt.s)
+    return mt.select_rows(
+        ref_samples=hl.agg.filter(mt.GT.n_alt_alleles() == N_ALT_REF, sample_ids),
```
hanars, line 35: probably better to use the hail `is_hom_ref`, `is_het`, `is_hom_var` instead of checking the number of alt alleles

v03_pipeline/lib/tasks/base/base_variant_annotations_table.py
hanars: I thought the agreement was the reference data WILL be annotated with vep, and that we would onyl be annotating the new rows with vep during loading

v03_pipeline/lib/tasks/update_variant_annotations_table_with_new_samples.py
```diff
@@ -0,0 +1,118 @@
...
+            self.ignore_missing_samples,
+        )
+
+        # Get new rows, annotate them with vep, format and annotate them,
+        # then stack onto the existing variant annotations table.
+        # We then re-annotate the entire table with the allele statistics.
+        # NB: the `unify=True` on the `union` here gives us the remainder
+        # of the fields defined on the existing table but not over the new rows
+        # (most importantly, the reference dataset fields).
+        new_variants_ht = callset_mt.anti_join_rows(ht).rows()
```
hanars, line 86: do we want to add explicit handling if this is empty?

luigi_pipeline/lib/hail_vep_runners.py
```diff
@@ -559,5 +559,7 @@ class HailVEPDummyRunner(HailVEPRunnerBase):
            'variant_class': 'SNV'},)
 
 
-    def run(self, mt, genome_version):
+    def run(self, mt, genome_version, vep_config_json_path=None):
+        if isinstance(mt, hl.Table):
+            return mt.annotate(vep=self.MOCK_VEP_DATA)
         return mt.annotate_rows(vep=self.MOCK_VEP_DATA)
```
hanars, line 565: Is there any case in the new pipeline that we need to annotate vep on a MatrixTable, instead of a regular table?

hanars (commented):
Theres a lot to review here so I did a pass but would definitely want another set of eyes on this as well

luigi_pipeline/lib/hail_vep_runners.py
bpblanken: I think at this point no!  So this code can be reverted.

v03_pipeline/lib/annotations/reference_dataset_collection.py
bpblanken: Yep!  This is bad copy-paste 🤔 .  

I think the intention here was to make the check very explicit, but it's probably unnecessary and over-engineered.  I believe `reference_dataset_collection_path` already does the check and will return `None` for hgmd in the wrong environment, so I can just use that.  The rename to `valid_reference_dataset_collection_path` is probably explicit enough.

v03_pipeline/lib/annotations/reference_dataset_collection.py
bpblanken: I've update the logic here so we read the table in outside of this function and pass the reference in as an argument.  

I think my original thinking here was that it was actually cleaner to read the table in multiple times because the table reading is actually lazy and there's no performance cost to having multiple references to the same persisted table.  Maybe that's not true?

v03_pipeline/lib/misc/sample_entries.py
bpblanken: `hl.agg.take` returns the list as a list....

e.g.

```
In [8]: ht.show()
+-------+-----------------------------------------+
|    id | entries                                 |
+-------+-----------------------------------------+
| int32 | array<struct{a: int32, sample_id: str}> |
+-------+-----------------------------------------+
|     0 | [(1,"a"),(2,"c"),(1,"e"),(2,"f")]       |
|     1 | [(2,"a"),(3,"c"),(4,"e"),(5,"f")]       |
+-------+-----------------------------------------+

In [9]: ht.aggregate(hl.agg.take(ht.entries.sample_id, 1))
2023-06-05 14:27:12.567 Hail: INFO: Coerced sorted dataset
Out[9]: [['a', 'c', 'e', 'f']]

In [10]: ht.aggregate(hl.agg.take(ht.entries.sample_id, 1))[0]
2023-06-05 14:27:51.531 Hail: INFO: Coerced sorted dataset
Out[10]: ['a', 'c', 'e', 'f']
```

v03_pipeline/lib/tasks/base/base_variant_annotations_table.py
bpblanken: Ah... I thought our conclusion was that the reference data would eventually be annotated with VEP but it was ok for us to leave it un-annotated for now and do the annotation here.  

I actually prefer just doing it now though since the reference data is still ungenerated... so I will proceed with adding the `run_vep` annotation to the `write_combined_ht.py` script.

v03_pipeline/lib/annotations/fields.py
```diff
@@ -0,0 +1,117 @@
...
+    (DatasetType.SNV, AnnotationType.REFERENCE_DATASET_COLLECTION): [
+        reference_dataset_collection.hgmd,
+        reference_dataset_collection.gnomad_non_coding_constraint,
+        reference_dataset_collection.screen,
+    ],
+    (DatasetType.SNV, AnnotationType.SAMPLE_LOOKUP_TABLE): [
+        sample_lookup_table.AC,
+        sample_lookup_table.AF,
+        sample_lookup_table.AN,
+        sample_lookup_table.hom,
```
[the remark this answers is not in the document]
bpblanken: These are nested under `gt_stats` now.

PR #493, opened 2023-06-01
Remove `geno2mp` and `coverage` datasets.  

bpblanken:


download_and_create_reference_datasets/v03/write_combined_reference_ht.py
```diff
@@ -9,35 +11,36 @@
...
-    'gnomad_genome_coverage',
-    'gnomad_exome_coverage',
 ]
 
 
-def run(environment: str, genome_version: str, dataset: str):
+def run(environment: str, genome_version: str, dataset: str | None):
+    hl._set_flags(  # noqa: SLF001
+        no_whole_stage_codegen='1',
+    )  # hail 0.2.78 hits an error on the join, this flag gets around it
```
mike-w-wilson, line 33: It would probably be good to test if this is still needed at some point

download_and_create_reference_datasets/v03/write_combined_reference_ht.py
```diff
@@ -9,35 +11,36 @@
 from hail_scripts.reference_data.config import GCS_PREFIXES, AccessControl
 from hail_scripts.utils.hail_utils import write_ht
 
-COMBINED_REFERENCE_HT_PATH = (
-    'combined_reference/combined_reference.GRCh{genome_version}.ht'
-)
+COMBINED_REFERENCE_HT_PATH = 'reference_datasets/combined.ht'
```
mike-w-wilson, line 14: I like these names

download_and_create_reference_datasets/v03/write_combined_reference_ht.py
```diff
@@ -67,10 +70,6 @@ def run(environment: str, genome_version: str, dataset: str):
         choices=['37', '38'],
         default='38',
     )
-    parser.add_argument(
-        '--dataset',
-        choices=DATASETS,
-        required=True,
-    )
+    parser.add_argument('--dataset', choices=DATASETS, default=None)
```
mike-w-wilson, line 73: Same thing with help.
```suggestion
    parser.add_argument('--dataset', choices=DATASETS, default=None, help="When passed, update the single dataset, otherwise update all datasets.")
```
Feature showed up here but feel free to rewrite that.

hail_scripts/reference_data/clinvar.py
```diff
@@ -55,6 +56,20 @@
 )
 
 
+def safely_move_to_gcs(tmp_file_name, gcs_tmp_file_name):
+    try:
+        subprocess.run(
+            [  # noqa: S603, S607
+                'gsutil',
+                'cp',
```
mike-w-wilson, line 64: Curious if we should use the multithread feature of `gsutil -m cp...`. Also not sure if we plan to use this outside of vcfs, would it hurt to through in the recursive option too? Or is that adding confusion? `gsutil -m cp -r`?

hail_scripts/reference_data/combine.py
```diff
@@ -22,15 +22,6 @@ def parse_version(ht: hl.Table, dataset: str, config: dict) -> hl.StringExpressi
     )
 
 
-def annotate_coverage(joined_ht: hl.Table, dataset_ht: hl.Table, dataset: str):
```
mike-w-wilson, line 25: 👋

hail_scripts/reference_data/combine.py
```diff
@@ -49,14 +40,17 @@ def get_select_fields(selects, base_ht):
...
             for attr in val.split('.'):
                 # Select from multi-allelic list.
                 if attr.endswith('#'):
-                    ht = ht[attr[:-1]][base_ht.a_index - 1]
+                    expression = expression[attr[:-1]][base_ht.a_index - 1]
                 else:
-                    ht = ht[attr]
-            select_fields[key] = ht
+                    expression = expression[attr]
+            # Parse float64s into float32s to save space!
```
mike-w-wilson, line 50: 💰

hail_scripts/reference_data/combine.py
```diff
@@ -49,14 +40,17 @@ def get_select_fields(selects, base_ht):
...
+            if expression.dtype == hl.tfloat64:
```
mike-w-wilson, line 51: Should this be bumped under the for loop? Though I guess we will always have values so expression=base_ht is not going to run right before this line in any case?

hail_scripts/reference_data/combine.py
```diff
@@ -167,13 +146,9 @@ def update_existing_joined_hts(
...
-    if 'coverage' not in dataset:
-        joined_ht = joined_ht.drop(dataset)
-        joined_ht = joined_ht.drop(f'{dataset}_globals')
-        joined_ht = joined_ht.join(dataset_ht, 'outer')
-        joined_ht = joined_ht.filter(
-            hl.any([~hl.is_missing(joined_ht[dataset]) for dataset in datasets]),
-        )
-    else:
-        joined_ht = annotate_coverage(joined_ht, dataset_ht, dataset)
+    joined_ht = joined_ht.drop(dataset, f'{dataset}_globals')
```
mike-w-wilson, line 149: huh, I did not know you could drop row annotations and globals at the same time, neat

hail_scripts/reference_data/combine_test.py
```diff
@@ -125,17 +131,21 @@ def test_custom_select(self, mock_read_table):
...
                         fathmm_MKL_coding_pred_id=1,
+                        GERP_RS=hl.eval(hl.float32(2.73)),
+                        phastCons100way_vertebrate=hl.eval(hl.float32(0.008)),
                     ),
                 ),
                 hl.Struct(
                     id=1,
                     mock_dbnsfp=hl.Struct(
-                        REVEL_score=0.052,
+                        REVEL_score=hl.eval(hl.float32(0.052)),
```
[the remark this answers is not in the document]
mike-w-wilson: ah does hail default to float64s?

mike-w-wilson (commented):
Just a couple questions, the vast majority are FMK.

hail_scripts/reference_data/combine_test.py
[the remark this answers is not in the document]
bpblanken: ah yes, great question.  yeah, hail defaults to float64s.  The idea here is to make the float32 exactly the same way hail is making it, so that the long tail of the float is equal.  [`assertAlmostEqual`](https://www.geeksforgeeks.org/python-unittest-assertalmostequal-function/) would be helpful here, but we're doing out assertions at the struct level and I didn't want to refactor the code here.

download_and_create_reference_datasets/v03/write_combined_interval_reference_ht.py
```diff
@@ -47,10 +50,6 @@ def run(environment: str, dataset: str):
         default='dev',
         choices=['dev', 'prod'],
     )
-    parser.add_argument(
-        '--dataset',
-        choices=DATASETS,
-        required=True,
-    )
+    parser.add_argument('--dataset', choices=DATASETS, default=None)
```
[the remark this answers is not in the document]
bpblanken: Yep, makes perfect sense.  Great suggestion.

download_and_create_reference_datasets/v03/write_combined_reference_ht.py
bpblanken: it definitely is, though I don't know why.  I was missing this line while testing at some point, and we failed at least a couple of times for strange reasons.  I added this line back in and we got way faster and did not fail.

hail_scripts/reference_data/clinvar.py
bpblanken: Yeah... I'm unsure what to do here.  I ran into several issues here and this was kinda my best stab at getting something working.

Our old approach had a line that added the tmp file to `hdfs`, which is the distributed filesystem shared by all the dataproc nodes and I had removed it several PRs ago because it blocked local testing.  After running into dataproc failures, I re-added that line... which worked for a while... and then _stopped_ working once we scaled the cluster because `hdfs` stopped being able to find the file.  Just using `gcs` with this command worked fine.

hail_scripts/reference_data/combine.py
bpblanken: I think the idea is we only need to convert to `hl.tfloat32` a single time rather than for each of the nested expressions.

hail_scripts/reference_data/combine.py
bpblanken: hm... I'm actually not sure this has been tested.. Also not sure why I decided to combine the `drop` lines.

bpblanken: I'll test real quick.

bpblanken: Yep!

```
In [7]: sample_lookup_ht.describe()
----------------------------------------
Global fields:
    'updates': set<struct {
        callset: str,
        project_guid: str
    }>
----------------------------------------
Row fields:
    'id': int32
    'ref_samples': dict<str, set<str>>
    'het_samples': dict<str, set<str>>
    'hom_samples': dict<str, set<str>>
----------------------------------------
Key: ['id']
----------------------------------------

In [8]: sample_lookup_ht.drop('ref_samples', 'updates').describe()
----------------------------------------
Global fields:
    None
----------------------------------------
Row fields:
    'id': int32
    'het_samples': dict<str, set<str>>
    'hom_samples': dict<str, set<str>>
----------------------------------------
Key: ['id']
----------------------------------------
```

I think I probably have been through this exercise before and don't have a memory of it.

download_and_create_reference_datasets/v03/write_combined_reference_ht.py
mike-w-wilson: Well that is good to know!

hail_scripts/reference_data/combine.py
mike-w-wilson: Ah that makes sense!

hail_scripts/reference_data/combine.py
mike-w-wilson: Sweet!

mike-w-wilson (approved):
LGTM!

PR #495, opened 2023-06-05
Fewer Hail Operations in remap. 

bpblanken:
@mike-w-wilson we talked about this briefly last week.


mike-w-wilson (approved):
Yeah, this is better!

PR #510, opened 2023-06-15
allow missing samples when remapping

bpblanken:


luigi_pipeline/lib/hail_tasks.py
```diff
@@ -176,12 +177,14 @@ def remap_sample_ids(mt, remap_path):
...
-                f'Only {remap_ht.semi_join(mt.cols()).count()} out of {remap_count} '
-                'remap IDs matched IDs in the variant callset.\n'
-                f'IDs that aren\'t in the callset: {missing_samples}\n'
-                f'All callset sample IDs:{mt.s.collect()}', missing_samples
-            )
+            message = f'Only {remap_ht.semi_join(mt.cols()).count()} out of {remap_count} '
+                    'remap IDs matched IDs in the variant callset.\n'
+                    f'IDs that aren\'t in the callset: {missing_samples}\n'
+                    f'All callset sample IDs:{mt.s.collect()}'
+            if self.ignore_missing_samples_when_remapping:
```
bpblanken, line 184: This allows us to preserve the behavior for the other pipelines but enable for (at least this load of) SVs.

bpblanken: I just double checked: https://console.cloud.google.com/dataproc/jobs/pyspark_compute_project_R0293_inmr_neuromuscular_disea_20230614_156ccf5e/monitoring?region=us-central1&project=seqr-project

Looks like the the sample that is missing from the remap file is already remapped in the VCF itself.  So... for that project at least, the cleaner solution would be to remove the `86AE_KM 86AE_KM_1` line from the remap file.

hanars: Theres no harm in cleaning up old remap files when we find issue. I think the SNP pipeline just ignores it so we never have any incentive to, but net it would be better to have those cleaned up. There is risk if the remap file is new, so the sample got remapped in the latest SNP callset after the SV callset was generated, but in the case of this project it has not had SNP data loaded since 2021 so that won't be the case

hanars (approved):

PR #513, opened 2023-06-21
vep enum mapping updates

bpblanken:
This is probably partially complete, since I think we have more string cleanup to do here.

### DIFF
--- a/v03_pipeline/lib/annotations/shared.py
+++ b/v03_pipeline/lib/annotations/shared.py
@@ -8,6 +8,104 @@ from hail_scripts.computed_fields import variant_id as expression_helpers
 
 from v03_pipeline.lib.model import ReferenceGenome
 
+AMINO_ACIDS = [
+    'A',
+    'C',
+    'D',
+    'E',
+    'F',
+    'G',
+    'H',
+    'I',
+    'K',
+    'L',
+    'M',
+    'N',
+    'P',
+    'Q',
+    'R',
+    'S',
+    'T',
+    'V',
+    'W',
+    'Y',
+    'X',
+    '*',
+    'U',
+]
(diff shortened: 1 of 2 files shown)


v03_pipeline/lib/annotations/shared.py
```diff
@@ -8,6 +8,103 @@
...
+    'S',
+    'T',
+    'V',
+    'W',
+    'Y',
+    'X',
+    '*',
+    'U',
+]
+AMINO_ACIDS_LOOKUP = hl.dict(hl.enumerate(AMINO_ACIDS, index_first=False))
```
bpblanken, line 36: Not sure we even need this, but the mapping was straightforwards so I went ahead and did it.

v03_pipeline/lib/annotations/shared.py
```diff
@@ -8,6 +8,103 @@
...
+
+BIOTYPES = [
```
bpblanken, line 38: Sourced from https://www.gencodegenes.org/pages/biotypes.html

The formats line up at least.

hanars (approved):
This all looks good. I also did confirm that we do not need all of  'lof',  'lof_filter', 'lof_flags', 'lof_info',and could instead use a much more compact rperesentation for those, but that might be out of thescope for this PR an warrant a different one

But for the record, we would need annotations for the following (names are proposals, you can make them whatever you want, and logic is very rough pseudo code):
- `is_lof_nagnag (bool): lof_flags === 'NAGNAG_SITE'`
- `lof_filters (array<str/enum>): (is_lof_nagnag() or lof == 'LC') and lof_filter and lof_filter.split(/&|,/)`

bpblanken:
Cool, I'll just go ahead and make a separate ticket for the `is_lof_nagnag` and `lof_filters` annotations.

PR #518, opened 2023-06-23
Vep annotation cleanup

bpblanken:


v03_pipeline/lib/annotations/shared_test.py
```diff
@@ -42,10 +42,8 @@ def test_sorted_transcript_consequences(self) -> None:
                         gene_id='ENSG00000188976',
                         hgvsc='ENST00000327044.6:c.1667C>T',
                         hgvsp='ENSP00000317992.6:p.Ser556Leu',
-                        lof=None,
-                        lof_filter=None,
-                        lof_flags=None,
-                        lof_info='INTRON_END:881781,EXON_END:881925,EXON_START:881782,DE_NOVO_DONOR_MES:-7.36719797135343,DE_NOVO_DONOR_PROB:0.261170618766552,DE_NOVO_DONOR_POS:-138,INTRON_START:881667,DE_NOVO_DONOR_MES_POS:-138,MUTANT_DONOR_MES:4.93863747168278',
+                        is_lof_nagnag=None,
+                        lof_filters=None,
```
hanars, line 46: could we add a test case for where there are filters?

hanars (changes requested):

v03_pipeline/lib/annotations/shared_test.py
bpblanken: Yep!

luigi_pipeline/lib/hail_vep_runners.py
```diff
@@ -115,8 +115,8 @@ class HailVEPDummyRunner(HailVEPRunnerBase):
                                                              'hgvsp': 'ENSP00000317992.6:p.Ser556Leu',
                                                              'impact': 'MODERATE',
                                                              'intron': hl.null('str'),
-                                                             'lof': hl.null('str'),
-                                                             'lof_filter': hl.null('str'),
+                                                             'lof': 'LC',
+                                                             'lof_filter': 'END_TRUNC,INCOMPLETE_CDS',
```
bpblanken, line 119: hm... actually, I'm not 100% sure this is the right format, gonna go double check.

bpblanken: Ok, I went into `gnomad_v3` and poked around, and I'm fairly confident that this format is correct.

hanars (approved):

PR #525, opened 2023-06-28
gCNV sample_fix fix

bpblanken:


luigi_pipeline/seqr_gcnv_loading.py
```diff
@@ -64,7 +64,12 @@ def import_dataset(self):
         )
 
         # rename the sample id column before the sample subset happens
-        mt = mt.key_cols_by(s = mt.sample_fix.first_match_in(SAMPLE_ID_REGEX)[0])
+        mt.key_cols_by(
+            s = hl.or_else(
```
bpblanken, line 68: There are now sample ids present that are actually the sample ids.

bpblanken: For more context, I think it was actually a bug that this did not hard fail without a regex match.  The old pipeline had that behavior, and I think I expected the `[0]` to `IndexError` rather than propagate a missing value.

hanars: I think we need to push back on the data delivery and only accept samples in a single format. I'm okay with either format, but I'm not okay with a mix

bpblanken: 👍 I'll ask for vanilla sample_ids.  It seems to be possible now!

PR #530, opened 2023-07-07
VEP annotation tweaks.

bpblanken:


v03_pipeline/lib/annotations/shared.py
```diff
@@ -98,10 +72,12 @@
     'unprocessed_pseudogene',
     'artifact',
     'lincRNA',
+    'lincrna',
```
bpblanken, line 75: I... am not sure fi we expect other `RNA` case sensitivity issues.  I think we'll find out soon enough.

v03_pipeline/lib/annotations/shared.py
```diff
@@ -8,33 +8,6 @@
...
-    'S',
-    'T',
-    'V',
-    'W',
-    'Y',
-    'X',
-    '*',
-    'U',
-]
-AMINO_ACIDS_LOOKUP = hl.dict(hl.enumerate(AMINO_ACIDS, index_first=False))
```
bpblanken, line 36: There's weird stuff buried in the amino acid string.

For example:

```
G/GPAPPPSLHSLPSX
```

or 

```
L/LGX
```

I think this is supposed to mean `L` -> `L` OR `G` or `X`?

bpblanken: I just ripped out the enumification so we could get a run in and see what's there.

hanars: can we run off this branch or do we need to merge this to master first?

bpblanken: Yeah, I (already did) can run off this branch.  It's a PR into a branch anyways... no rush here.

PR #539, opened 2023-07-20
hardcode `dont_validate` in MITO loader.

bpblanken:


luigi_pipeline/seqr_mito_loading.py
```diff
@@ -24,6 +24,7 @@ class SeqrMitoVariantMTTask(SeqrVCFToVariantMTTask):
     Loads all annotations for the variants of a Matrix Table into a Hail Table.
     """
     dataset_type = 'MITO'
+    dont_validate = True
```
bpblanken, line 27: We either need this... or re-work the validation so that it works with other dataset types.

hanars: this seems fine for now

hanars (approved):

PR #551, opened 2023-08-02
Drop mito fields.

bpblanken:


hail_scripts/reference_data/config.py
```diff
@@ -401,6 +396,27 @@ def custom_mpc_select(ht):
             },
         },
     },
+    'dbnsfp_mito': {
```
bpblanken, line 399: Just a heads up that I opted to rename the field from `dbnsfp` to `dbnsfp_mito` and separate the configs.  Keeping a single config got too hairy with different selects.

hanars: I like this. It also makes it easier for seqr to handle, as we can be explicit about which reference source we are pulling from

hail_scripts/reference_data/config.py
```diff
@@ -401,6 +396,27 @@ def custom_mpc_select(ht):
...
+        },
+        '38': {
+            'version': '4.2',
+            'path': 'gs://seqr-reference-data/GRCh38/dbNSFP/v4.2/dbNSFP4.2a_variant.ht',
+            'custom_select': dbnsfp_mito_custom_select,
+            'enum_select': {
+                'SIFT_pred': ['D', 'T'],
+                'MutationTaster_pred': ['D', 'A', 'N', 'P'],
+            },
+            'filter': lambda ht: ht.locus.contig == 'chrM',
```
hanars, line 417: is there a reason we only filter for 38 and not 37?

bpblanken: Nope, it should be there!

hanars (approved):

PR #552, opened 2023-08-03
tweak sv locii

bpblanken:


v03_pipeline/lib/annotations/sv.py
```diff
@@ -113,6 +103,14 @@ def cpx_intervals(ht: hl.Table, **_: Any) -> hl.Expression:
     )
 
 
+def end_locus(ht: hl.Table, **_: Any) -> hl.StructExpression:
+    return hl.if_else(
+        hl.is_defined(ht.info.END2),
+        hl.struct(contig=ht.info.CHR2, position=ht.info.END2),
```
hanars, line 109: wouuldn't it be netter to use the data type locus?

bpblanken: sure... I think this has just been carried forwards from https://github.com/broadinstitute/seqr-loading-pipelines/blob/78f84ae48e76cfadc8b3ebb53d09ea473ed6ad98/sv_pipeline/genome/load_data.py#L58-L60, so should be fine to change.

hanars (approved):

PR #555, opened 2023-08-10
Benb/data types

bpblanken:


hail_scripts/reference_data/config.py
```diff
@@ -67,6 +67,17 @@ def dbnsfp_mito_custom_select(ht):
     return selects
 
 
+def custom_gnomad_mito(ht):
+    selects = {}
+    selects['AN'] = hl.int32(ht.AN)
```
bpblanken, line 72: I've updated the mito `gt_stats` to use `AC_hom` and `AF_hom`.  Should we propagate the change to the reference data as well?

hanars: I think we do want the change in the reference data because it should make the file a bit smaller without losing anything important. If we propagate the change to the reference data, we shouldn't need this change at all correct?

bpblanken: Ah, this was a poorly phrased question.  To clarify, this is the reference_data config file, so we'll need something approximating this change to get the `hl.int32`s in there.  

I was asking about the rename we did in the `MITO` formatting annotations, where instead of denoting `AC_hom` and `AF_hom` as `AC` and `AF`, we decided to keep them as `AC_hom` and `AF_hom`.  The mito reference data (gnomad, but also helix) are still using `AC` and `AF`.  We want to use `AC_hom` and `AF_hom` everywhere right?

hanars: I would say we should keep them as `AC_hom` etc here too

PR #558, opened 2023-08-17
Bump the number of shards

bpblanken:


luigi_pipeline/lib/hail_tasks.py
```diff
@@ -272,6 +272,6 @@ def cleanup(self, es_shards):
 
     def _mt_num_shards(self, mt):
         # The greater of the user specified min shards and calculated based on the variants and samples
-        denominator = 1.4*10**9
+        denominator = 1.2*10**9
```
bpblanken, line 275: this is ~15-20% increase is shard count, we could go for more?

luigi_pipeline/lib/hail_tasks.py
```diff
@@ -272,6 +272,6 @@ def cleanup(self, es_shards):
 
     def _mt_num_shards(self, mt):
         # The greater of the user specified min shards and calculated based on the variants and samples
-        denominator = 1.4*10**9
+        denominator = 1.1*10**9
```
bpblanken, line 275: never mind there's an airflow op for this.

PR #560, opened 2023-08-18
Benb/fix gcnv missing entries

bpblanken:


v03_pipeline/lib/misc/io.py
```diff
@@ -49,13 +49,12 @@ def import_gcnv_bed_file(callset_path: str) -> hl.MatrixTable:
     )
     mt = ht.to_matrix_table(
         row_key=['variant_name', 'svtype'],
-        col_key=['sample_cram_basename'],
+        col_key=['sample_fix'],
```
bpblanken, line 52: 🍭

v03_pipeline/lib/misc/io.py
```diff
@@ -71,6 +70,8 @@ def import_gcnv_bed_file(callset_path: str) -> hl.MatrixTable:
             hl.agg.collect_as_set(parse_gcnv_genes(mt.genes_LOF_Ensemble_ID)),
         ),
     )
+    mt = mt.unfilter_entries()
```
bpblanken, line 73: This is the actual one line fix.

v03_pipeline/lib/tasks/write_family_table_test.py
```diff
@@ -522,5 +522,55 @@ def test_gcnv_write_family_table_task(self, mock_dataroot: Mock) -> None:
                         QS=30,
                     ),
                 ],
+                [
+                    hl.Struct(
```
bpblanken, line 526: Do we prefer to have actual "hl.missing" here, or is a struct with every entry itself `hl.missing` sufficient?  The latter is substantially easier to work with, but may take up more space.

v03_pipeline/lib/misc/io.py
hanars: yeah I hit this one too once, its a doozy

v03_pipeline/lib/tasks/write_family_table_test.py
hanars: either should work for search code, so whatever you prefer

hanars (approved):

v03_pipeline/lib/tasks/write_family_table_test.py
bpblanken: Just gonna leave as is.  Easier to reason about.

PR #574, opened 2023-09-06
Automate the 'filters merging' process.

bpblanken:
Intention here is to remove the need for a human to run [this script](https://github.com/broadinstitute/tgg_methods/blob/master/tgg/seqr_scripts/import_vcf.py)


v03_pipeline/lib/misc/io.py
```diff
@@ -108,6 +111,9 @@ def import_callset(
         mt = split_multi_hts(mt)
     if dataset_type == DatasetType.SV:
         mt = mt.annotate_rows(variant_id=mt.rsid)
+    if sample_type == SampleType.WES and filters_path:
```
hanars, line 114: is there a reason to restrict this to exome data? If currently we only do this for WES data but not WGS, we could just only provide the `filters_path` for WES but leave open the option in the future to use this more broadly

hanars (changes requested):

v03_pipeline/lib/tasks/base/base_update_task.py
```diff
@@ -2,12 +2,13 @@
 import luigi
 
 from v03_pipeline.lib.misc.io import write
-from v03_pipeline.lib.model import DatasetType, Env, ReferenceGenome
+from v03_pipeline.lib.model import DatasetType, Env, ReferenceGenome, SampleType
```
hanars, line 5: do we need to pass this parameter around anymore?

hanars (changes requested):

v03_pipeline/lib/tasks/base/base_update_task.py
bpblanken: I'd honestly prefer to just keep it around.  I think this would be the second (or third) time I've added and removed it... and we 100% need it passed in when we do the sample type validation.

hanars (approved):

PR #575, opened 2023-09-07
Add core "sex_check" functionality.  

bpblanken:
Pulls in the core methods of the [sex_check](https://github.com/broadinstitute/tgg_methods/blob/master/tgg/seqr_scripts/sex_check.py) script, but does not yet "task-ify" into the pipeline itself.


v03_pipeline/lib/methods/sex_check.py
```diff
@@ -0,0 +1,207 @@
...
+    mt = mt.filter_rows(hl.is_snp(mt.alleles[0], mt.alleles[1]))
+
+    # Filter to PASS variants only (variants with empty or missing filter set)
+    # TODO: Make this an optional argument before moving to gnomad_methods
+    mt = mt.filter_rows(
+        hl.is_missing(mt.filters) | (mt.filters.length() == 0),
+        keep=True,
+    )
+
+    impute_sex_ht = hl.impute_sex(
```
bpblanken, line 131: One note here, I removed the `run_hails_impute_sex` function and the line that was filtering to the x config itself.  Per the [hail docs](https://hail.is/docs/0.2/methods/genetics.html#hail.methods.impute_sex) for this method, that filtering is happening within the method itself.  We were also missing the step to "Remove loci in the pseudoautosomal region" anyways.

v03_pipeline/lib/methods/sex_check.py
```diff
@@ -0,0 +1,207 @@
...
+    FEMALE = 'XX'
+    MALE = 'XY'
+
+
+def get_contig_cov(
+    mt: hl.MatrixTable,
+    reference_genome: ReferenceGenome,
+    contig: str,
+    call_rate_threshold: float,
+    af_threshold: float = 0.01,
```
bpblanken, line 32: I removed the `af_field` argument because I think we're going to expect `AF` for our snv_indel callsets.  If that's not the case we can add it back.

v03_pipeline/lib/methods/sex_check.py
```diff
@@ -0,0 +1,207 @@
...
+    df = ht.to_pandas()
+    plt.clf()
+    plt.hist(df['f_stat'])
+    plt.xlabel('Fstat')
+    plt.ylabel('Frequency')
+    plt.axvline(xy_fstat_threshold, color='blue', linestyle='dashed', linewidth=1)
+    plt.axvline(xx_fstat_threshold, color='red', linestyle='dashed', linewidth=1)
+    plt.savefig(buf, format='png')
+    buf.seek(0)
+    return buf
```
bpblanken, line 82: This method returns a binary buffer now rather than writing to a static location.  the tasks themselves will manage paths/writing.

v03_pipeline/lib/methods/sex_check.py
bpblanken: A second pass here led me to pulling `AF` off the `variant_qc` struct rather than from the vcf itself.  I think that's right?

bpblanken: Following up on myself again, `af_field` is necessary even within the minimal test setup we have, so we need it for sure!

v03_pipeline/lib/methods/sex_check.py
hanars: what is the utility of a plot? It seems unlikely that this will be able to feed into any sort of automated decision making

v03_pipeline/lib/methods/sex_check.py
```diff
@@ -0,0 +1,215 @@
...
+        normalized_x_coverage=hl.or_missing(
+            ht[f'{normalization_contig}_mean_dp'] > 0,
+            ht.chrX_mean_dp / ht[f'{normalization_contig}_mean_dp'],
+        ),
+    )
+    ht = ht.annotate(
+        sex=(
+            hl.case()
+            .when(hl.is_missing(ht.is_female), Ploidy.AMBIGUOUS.value)
+            .when(
```
hanars, line 196: this is mostly shared with the `not use_chrY_cov` case above, it might be better to have a helper function to run all the shared code, with an optional argument to determine whether or not to add the one `when` clause thats different, or to reorganize the function so instead of having a branching return with duplicated logic and just have a single path to the return that handles that conditions

v03_pipeline/lib/model/definitions.py
```diff
@@ -22,60 +22,36 @@ def v02_value(self) -> str:
...
+            },
+        }[self]
+
+    @property
+    def standard_contigs(self) -> set[str]:
+        return {
+            ReferenceGenome.GRCh37: {
+                *self.autosomes,
+                *self.sex_chromosomes,
+                'MT',
```
hanars, line 50: doe this end up different than using hail's `reference.contigs`?

hanars (changes requested):

v03_pipeline/lib/methods/sex_check.py
bpblanken: I think we've been using it as a validation step to ensure the `impute_sex` algorithm hasn't behaved incorrectly and requires human intervention.  

My hope is that, if the image is actually used, we can post it to slack to get a human "OK" to proceed with the DAG.  

@matren395 and @mike-w-wilson would you mind providing some context about the plot and how often it's needed?

v03_pipeline/lib/methods/sex_check.py
```diff
@@ -0,0 +1,215 @@
...
+    if normalization_contig not in valid_contigs:
+        msg = f'Contig: {normalization_contig} is invalid for this reference genome'
+        raise ValueError(msg)
+
+    # Filter to SNVs and biallelics
+    # NB: We should already have filtered biallelics, but just
+    mt = mt.filter_rows(hl.is_snp(mt.alleles[0], mt.alleles[1]))
+
+    # Filter to PASS variants only (variants with empty or missing filter set)
+    # TODO: Make this an optional argument before moving to gnomad_methods
```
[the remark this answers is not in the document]
bpblanken: I'm just gonna remove the TODO.

v03_pipeline/lib/methods/sex_check.py
hanars: > My hope is that, if the image is actually used, we can post it to slack to get a human "OK" to proceed with the DAG.

When @mike-w-wilson and I initially discussed automating this workflow, the rough flow we had agreed on is that this check would remove failed families from the subset to load and then proceed with loading for all passed families in the project. The failed families would be reported back to the PMs somehow, and then they can always re-trigger loading a project whenever they felt like relatedness had been resolved. If this plot would be valuable to the PMs, I could see the utility in generating it for failed samples, but would want to make sure its something they can intepret

v03_pipeline/lib/model/dataset_type.py
```diff
@@ -111,9 +111,6 @@ def row_fields(
                 'filters',
                 'common_low_heteroplasmy',
                 'hap_defining_variant',
-                'AF_het',
```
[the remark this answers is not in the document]
bpblanken: unrelated bug (well... no symptoms, they're just unnecessary fields) fix that slipped in.  I'll remove from this pr.

v03_pipeline/lib/model/definitions.py
bpblanken: Yes, hail's reference contigs have many non-standard contigs after the first 56.  I swapped in using the first `[:25]` though.

v03_pipeline/lib/methods/sex_check.py
bpblanken: I cleaned this up a bit.

v03_pipeline/lib/methods/sex_check.py
mike-w-wilson: The manual review of the plot is to confirm  the cutoff used for the imputed chromosomal sex label is a good cutoff. It is not a check of provided information vs imputed. The plot is bimodal so its a very quick check. However, since we started running this, I have never adjusted the cutoff threshold and I dont expect it to change for our internal callsets given the samples are the same and the gvcfs havent changed. But it would be appropriate to check the plot whenever we there is a major change in the DSPs pipelines, so if our data is being recalled with the GVS pipeline or gVCFs start getting generated with DRAGEN. The cutoff will likely stay the same but this would ensure they arent calling more homozygotes than we are now.

hanars: In that case I think this function belong in a methods repo, not the seqr loading pipeline repo

bpblanken: +1 

Per convo with @mike-w-wilson we're going to remove the plot and be more aggressive about moving parameters to constant.  

I'm also going to spend a little bit of time trying to brainstorm a heuristic we could use to trigger a failure if the plot doesn't show a clear bimodal.

bpblanken: I've cleaned this up (and simplified) quite a bit.

1) Removed the "contig_coverage" logic.  Per @mike-w-wilson that was experimental and isn't used in practice; `hl.impute_sex` is going to be sufficient for what we need.
2) Added a simple heuristic to validate the `f_stat` distribution.  "If more than 1/100th of samples have `f_stats` between the `0.5` and `0.75` bounds and are classified as ambiguous, throw an error".  If 1/100 is too lenient we can lower the threshold to something safer!

hanars (approved):

PR #576, opened 2023-09-11
Add "relatedness" methods.

bpblanken:
Ports logic from [this script](https://github.com/broadinstitute/tgg_methods/blob/master/tgg/relatedness/get_ibd.py) over.


v03_pipeline/lib/methods/relatedness.py
```diff
@@ -0,0 +1,75 @@
+from __future__ import annotations
+
+import hail as hl
+from gnomad.sample_qc.pipeline import filter_rows_for_qc
+
+from v03_pipeline.lib.model import ReferenceGenome
+
+
+# Stub port from https://github.com/broadinstitute/gnomad_qc/blob/9763d8a581665ed0be8bf2ea09d74e3cabc4806e/gnomad_qc/v2/resources/sample_qc.py#L40
+def qc_mt_path(reference_genome: ReferenceGenome) -> str:
```
bpblanken, line 10: I hardcoded these to avoid trying to import `gnomad_qc`.

matren395: why would we avoid importing gnomad_qc ?

bpblanken: why would we want to?

I think just dependency weight (how the complexity of the dependency weighs against re-implementing the functionality locally).  We need these to run on `dataproc`, which means we need clean `pip` installs or a bundled `.whl` somewhere.  I found [gnomad](https://pypi.org/project/gnomad/) but not `gnomad_qc` on pip.

hanars: I agree that we do not want to be importing a heavyweight module just to get us a couple lightweight helper functions.

v03_pipeline/lib/methods/relatedness.py
```diff
@@ -0,0 +1,75 @@
...
+    ref_str = '.grch38' if reference_genome == ReferenceGenome.GRCh38 else ''
+    return f'gs://gnomad/sample_qc/mt/gnomad.joint.high_callrate_common_biallelic_snps.pruned{ref_str}.mt'
```
hanars, line 12: we probably want to use a constant for the shared google bucket path

hanars, line 12: are we sure the pipeline will have access to these resources?

v03_pipeline/lib/methods/relatedness.py
```diff
@@ -0,0 +1,75 @@
...
+        min_callrate=0.99,
+        apply_hard_filters=False,
+    )
+    if not use_gnomad_in_ld_prune:
+        mm_pruned = hl.ld_prune(mt.GT, r2=0.1)
+        return mt.filter_rows(hl.is_defined(mm_pruned[mt.row_key]))
+    if reference_genome == ReferenceGenome.GRCh37:
+        pruned_mt = hl.read_matrix_table(qc_mt_path(reference_genome))
+    elif reference_genome == ReferenceGenome.GRCh38:
+        pruned_mt = hl.read_matrix_table(qc_path())
```
hanars, line 37: wouldn't is be neater to just to modify `qc_mt_path(reference_genome)` to work properly for both 37 and 38? This is the only place it is used and you would expect the helper function that takes a reference_genome argument to work with multiple reference_genomes and not just one

v03_pipeline/lib/methods/relatedness.py
```diff
@@ -0,0 +1,75 @@
...
+    reference_genome: ReferenceGenome,
+    af_field: str = 'info.AF',
+    use_gnomad_in_ld_prune: bool = True,
+) -> hl.Table:
+    mt = filter_and_ld_prune(mt, reference_genome, use_gnomad_in_ld_prune)
+    # NB: ibd did not work by default with my pip install of `hail` on an M1 MacOSX.
+    # I had to build hail by source with the following:
+    # - brew install lz4
+    # - CXXFLAGS='-I/opt/homebrew/include/' HAIL_COMPILE_NATIVES=1 make -C hail install
+    # Hail issue here: https://discuss.hail.is/t/noclassdeffounderror-could-not-initialize-class-is-hail-methods-ibsffi/2453
```
hanars, line 67: I would hope if we run this on a supported hail docker image in dataproc we would not need to do anything special

v03_pipeline/lib/methods/relatedness.py
```diff
@@ -0,0 +1,75 @@
...
+    kin_ht = hl.identity_by_descent(mt, maf=mt[af_field], min=0.10, max=1.0)
+    kin_ht = kin_ht.annotate(
+        ibd0=kin_ht.ibd.Z0,
+        ibd1=kin_ht.ibd.Z1,
+        ibd2=kin_ht.ibd.Z2,
+        pi_hat=kin_ht.ibd.PI_HAT,
+    ).drop('ibs0', 'ibs1', 'ibs2', 'ibd')
+    return kin_ht.key_by('i', 'j')
```
hanars, line 75: I think instead of dropping all the fields it would be easier to select the one we do need by defining the annotations we are adding: 
```
annotations = {'ibd0': ...}
return kin_ht.annotate(**annotations).key_by('i', 'j').select(*annotations)
```

hanars (changes requested):

v03_pipeline/lib/methods/relatedness.py
bpblanken: 👍 It works fine in dataproc, and probably docker as well.  I left the comment since running the tests locally will now fail without building hail from source.

v03_pipeline/lib/methods/relatedness.py
bpblanken: Yeah, I left the code structure as is was in `ibd.py`, but might as well change it to make it make more sense.

v03_pipeline/lib/methods/relatedness.py
bpblanken: I believe it works as long as we set the "requester.pays" configuration for hail, which we do [here](https://github.com/broadinstitute/seqr-pipeline-airflow/blob/main/dags/dependencies/operators.py#L159-L161)

hanars: it may be cheaper for us to store a copy than pay every time we load data though? Depending on what region those are stored in etc

bpblanken: I wrote the code to import them and place cached copies in `seqr-reference-data-private` with the little "cached reference dataset query" pattern, so the import is at least reproducible and treated similarly to what we're doing with other supplemental tables.

PR #578, opened 2023-09-19
Taskify Sex and Relatedness Checks

bpblanken:


v03_pipeline/lib/methods/relatedness.py
```diff
@@ -1,55 +1,38 @@
...
+    gnomad_qc_ht: hl.Table | None,
 ) -> hl.MatrixTable:
     mt = filter_rows_for_qc(
         mt,
         min_af=0.001,
         min_callrate=0.99,
         apply_hard_filters=False,
     )
-    if not use_gnomad_in_ld_prune:
+    if not gnomad_qc_ht:
```
bpblanken, line 17: I ported the `read_table` into the relatedness task (so the io looks like it does everywhere else) and simplified the logic to "if the table is available use it, else use `ld_prune`).

v03_pipeline/lib/methods/sex_check.py
```diff
@@ -23,30 +23,6 @@
 )
 
 
-class Ploidy(Enum):
-    AMBIGUOUS = 'ambiguous_sex'
-    FEMALE = 'XX'
-    MALE = 'XY'
-
-
-def annotate_discrepant_sex(
```
bpblanken, line 32: This logic (which was ported directly from the methods scripts) is happening within [this function now](https://github.com/broadinstitute/seqr-loading-pipelines/pull/578/files#diff-81bda30e9edc96dbaf6b9533d4db17f8afa17df35de4b37c3c05e1856e6779dbR157)

v03_pipeline/lib/misc/family_loading_failures.py
```diff
@@ -0,0 +1,169 @@
...
+    if not coefficients:
+        return False
+    return np.allclose(
+        coefficients,
+        relation.coefficients,
+        0.1,
+    )
+
+
+def passes_all_relatedness_checks(  # noqa: C901
```
bpblanken, line 37: To briefly summarize the approach here, the idea is to check every identified relationship in the pedigree for a sample.  A sample succeeds if every relationship is validated.  A family succeeds if every sample succeeds.  If a relationship exists but isn't identifiable in the pedigree (for example a brother/sister pair with no listed parents), we don't have anything to check so the family will succeed.

v03_pipeline/lib/misc/pedigree.py
```diff
@@ -1,28 +1,176 @@
...
+    def parse_collateral_lineage(
+        samples: dict[str, Sample],
+    ) -> dict[str, Sample]:
+        # NB: relationships are identified unidirectionally here (for better or for worse)
+        # A sample_i that is siblings with sample_j, will list sample_j as as sibling, but
+        # sample_j will not list sample_i as a sibling.  Relationships only appear in the
+        # ibd table a single time, so we only need to check the pairing once.
+        for sample_i, sample_j in itertools.combinations(samples.keys(), 2):
+            # If other sample is already related, continue
+            if sample_j in {
```
hanars, line 107: not super important, but you use a series of `if` with a `continue` at the end, and that could be rewritten as `elif`

v03_pipeline/lib/tasks/write_family_table.py
```diff
@@ -71,23 +71,21 @@ def requires(self) -> luigi.Task:
     def create_table(self) -> hl.Table:
         pedigree_ht = import_pedigree(self.project_pedigree_path)
         callset_mt = hl.read_matrix_table(self.input().path)
-        callset_family_guids = set(callset_mt.family_guids.collect()[0])
-        if self.family_guid not in callset_family_guids:
-            msg = f'Family: {self.family_guid} was not complete in this callset'
-            raise ValueError(msg)
-        sample_subset_ht = samples_to_include(
-            pedigree_ht,
+        families = parse_pedigree_ht_to_families(pedigree_ht)
```
hanars, line 74: it feels like we would want to do this once, not parse the full set of pedigrees once per family

PR #596, opened 2023-10-19
dynamic partitioning

bpblanken:


v03_pipeline/lib/misc/io.py
```diff
@@ -17,6 +20,30 @@ def does_file_exist(path: str) -> bool:
     return os.path.exists(path)
 
 
+def file_size_bytes(path: str) -> int:
```
bpblanken, line 23: I asked [zulip](https://hail.zulipchat.com/#narrow/stream/123010-Hail-Query-0.2E2-support/topic/Possible.20to.20access.20the.20size.20of.20Hail.20Table.2FMatrixTable.3F) if there was a better way, and I don't think there is 😢

v03_pipeline/lib/misc/io.py
```diff
@@ -144,18 +169,17 @@ def write(
...
-            ),
+        path = os.path.join(
+            Env.HAIL_TMPDIR,
+            f'{uuid.uuid4()}.{suffix}',
         )
-    # "naive_coalesce" will decrease parallelism of hail's pipelined operations
-    # , so we sneak this re-partitioning until after the checkpoint.
-    if n_partitions:
+        # not using checkpoint to read/write here because the checkpoint codec is different, leading to a different on disk size.
+        t.write(path)
```
hanars, line 181: I feel like writing twice is not ideal but I can't think of a better way to do it, unless we wanted to do some sort of really rough estimate involving the number of rows. Is there a reason that this is nested inside an `if checkpoint:` block though? I feel like we would always want to compute the correct partition number.

bpblanken: I think there isn't a good reason at this point, so I will change it!  Originally, I believe the family tables were not checkpointed because they are freshly written each time, but that isn't happening any more.  Pretty sure `checkpoint` is never false at this point, in looking at the function calls in the rest of the codebase, so I'll just remove it.

hanars (approved):

Issue #607, opened 2023-10-31
Changes filter behavior for unexpected chromosomes in the AnVIL data loading pipeline


lynnpais:
Several users have submitted VCFs with non-standards chromosomes. We had initially decided to hold off on making changes in V2, but this seems necessary to keep things moving. 

Once Ben updates the AnVIl data loading pipeline to remove this filter, @matren395 can reload the projects that failed for this reason.

PR #608, opened 2023-10-31
Filter invalid contigs rather than throw an error, drop `AF` and `PL`.

bpblanken:
Closes https://github.com/broadinstitute/seqr-loading-pipelines/issues/607

### DIFF
--- a/luigi_pipeline/seqr_loading.py
+++ b/luigi_pipeline/seqr_loading.py
@@ -115,7 +115,17 @@ class SeqrVCFToMTTask(HailMatrixTableTask):
         hl._set_flags(use_new_shuffle='1') # Interval ref data join causes shuffle death, this prevents it
 
         mt = self.import_dataset()
+        if hasattr(mt, 'PL'):
+            mt = mt.drop('PL')
+        if hasattr(mt, 'AF'):
+            mt = mt.drop('AF')
         mt = self.split_multi_hts(mt)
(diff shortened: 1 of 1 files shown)


luigi_pipeline/seqr_loading.py
```diff
@@ -115,7 +115,17 @@ def read_input_write_mt(self):
         hl._set_flags(use_new_shuffle='1') # Interval ref data join causes shuffle death, this prevents it
 
         mt = self.import_dataset()
+        if hasattr(mt, 'PL'):
```
bpblanken, line 118: This has been running in production... it doesn't hurt anything in theory.

hanars (approved):

karynne7:
Is it possible to put the AF portion back? Was there a reason for this change? We spoof the AF field to filter artifacts as internal controls, and loading really large VCFs hasn't been easy for us cloud-wise. This had been our solution until now.

hanars:
These changes provide stability and reliability when loading data with potentially inaccurate or malformed annotations. We will not be reverting this behavior, as reverting changes that make the pipeline broadly stable is not something we are willing to do.
 I am not entirely sure why spoofing the AF field feels like a reasonable way to handle artifacts - if you have
variants you would never want returned from search you should prefilter your VCF not to include them in loading. If you want variants to come bback in search but have the option to filter them out, I would recommend adding flags to the "filters" list

PR #612, opened 2023-11-06
add sample type annotation to project/family tables.

bpblanken:


bpblanken:
good thing we have tests :/

hanars (approved):

PR #622, opened 2023-11-17
Support multiple callsets.

bpblanken:
Ended up doing this because it was straightforwards (and less painful than running 20 times is turning out to be).

Screenshot 2023-11-17 at 10 13 18 AM


hanars (approved):
wow that is so much easier to do than I anticipated. Does it run into any resource issues?

bpblanken:
I think we will see with respect to resource issues.  The annotations task are still running serially (I've got it set up so each annotations task takes the equivalent of a global lock).  I'm actually seeing crazy issues with the lookup table getting poorly partitioned so am trying a repartition there to see what happens...

PR #625, opened 2023-11-21
Fix up dockerfile

bpblanken:
@barslmn reported this was broken.  We've updated our `requirements.txt` since the last push to `gcr.io/seqr-project/pipeline-runner:gcloud-prod` , and this is the minimum to get us back to something functional (and does not fix many other low hanging fruit like pip caching).  

FWIW we should probably be using a hail docker image going forwards, but I couldn't get one to load performantly locally.

PR #627, opened 2023-11-22
bump the luigi version to 3.4.0

barslmn:
Fix for https://github.com/broadinstitute/seqr-loading-pipelines/pull/625#issuecomment-1822477463.


barslmn:
> FYI we also need to bump in the [root of the repo](https://github.com/broadinstitute/seqr-loading-pipelines/pull/630), which I'm handling! Thank you for finding though!

Thanks for pointing it out.

PR #628, opened 2023-11-22
corrected the ref genome paths

barslmn:
Fasta files are [bgzipped here](https://github.com/broadinstitute/seqr-loading-pipelines/blob/aed106c8250b527197b39ddc6374d85b2942b673/docker/bin/download_reference_data.sh#L19C5-L19C5). However, vep-config argument path is plain text.


bpblanken (approved):
Yep, this looks correct to me!

Issue #624, opened 2023-11-18
Container is not building


barslmn:
I got the following changes:

1. stretch -> bookworm. because stretch repos are gone.
2. python3.7 -> 3.9. because 3.7 no longer compiles. There is one python module that doesn't work anything about 3.9.
3. gcloud sdk from 288 -> 488. Again it doesn't install.
4. python-pkg-resources is not installed with apt as python3-pkg-resources, and python-smbus is python3-smbus.

I got two questions:
1. Do you forsee updating gcloud sdk from 288 to 455 would be problematic?
5. bw2/hail-utils repo is pretty much empty now. Is that still a requirement? hail-utils is no longer a module and can't be installed.

It builds after this patch. I can create a pull request if changes are okay.

```
diff --git a/docker/Dockerfile b/docker/Dockerfile
index 97d9526f..c08e73af 100644
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,4 +1,4 @@
-FROM bitnami/minideb:stretch
+FROM bitnami/minideb:bookworm
 
 LABEL maintainer="Broad TGG"
 
@@ -27,14 +27,14 @@ RUN install_packages \
 # install java-8    (adds 340Mb)
 RUN install_packages default-jdk
 
-# install python3.7 (adds 800Mb)
+# install python3.9 (adds 800Mb)
 RUN apt-get update \
        && apt-get dist-upgrade -y \
-       && apt-get install -y --no-install-recommends python-smbus libncursesw5-dev libgdbm-dev libc6-dev zlib1g-dev libsqlite3-dev libssl-dev openssl libffi-dev \
+       && apt-get install -y --no-install-recommends python3-smbus python3-pkg-resources libncursesw5-dev libgdbm-dev libc6-dev zlib1g-dev libsqlite3-dev libssl-dev openssl libffi-dev \
        && apt-get clean \
     && rm -rf /var/lib/apt/lists/*
 
-ENV PYTHON_VERSION="3.7.1"
+ENV PYTHON_VERSION="3.9.17"
 RUN wget https://www.python.org/ftp/python/${PYTHON_VERSION}/Python-${PYTHON_VERSION}.tar.xz \
        && tar xvf Python-${PYTHON_VERSION}.tar.xz \
        && rm Python-${PYTHON_VERSION}.tar.xz
@@ -44,16 +44,14 @@ RUN ./configure --enable-optimizations \
     && make clean
 WORKDIR /
 
[...]

Issue #623, opened 2023-11-18
ValueError: grch38_to_grch37_ref_chain path not found: gs://hail-common/references/grch38_to_grch37.over.chain.gz


barslmn:
Hi,
I have a local seqr installation and trying to load the data locally without gcloud. I have followed the manual installation document linked in seqr repo and have docker containers. 

I'm trying to load data using the following command:

```
docker-compose exec pipeline-runner load_data.sh 37 WES demo demo.vcf.gz
```

However, it is giving the following error:

```
Traceback (most recent call last):
  File "/usr/local/lib/python3.7/site-packages/luigi/worker.py", line 199, in run
    new_deps = self._run_get_new_deps()
  File "/usr/local/lib/python3.7/site-packages/luigi/worker.py", line 141, in _run_get_new_deps
    task_gen = self.task.run()
  File "/seqr-loading-pipelines/luigi_pipeline/seqr_loading.py", line 89, in run
    if self.grch38_to_grch37_ref_chain: check_if_path_exists(self.grch38_to_grch37_ref_chain, "grch38_to_grch37_ref_chain")
  File "/seqr-loading-pipelines/luigi_pipeline/seqr_loading.py", line 23, in check_if_path_exists
    raise ValueError(f"{label} path not found: {path}")
ValueError: grch38_to_grch37_ref_chain path not found: gs://hail-common/references/grch38_to_grch37.over.chain.gz
```

This is [set to a bucket link as default here](https://github.com/barslmn/seqr-loading-pipelines/blob/7db47b2c0bbe01398f768888a5ef18aff8b9542e/luigi_pipeline/seqr_loading.py#L62).
But tries  [check_if_path_exists](https://github.com/barslmn/seqr-loading-pipelines/blob/7db47b2c0bbe01398f768888a5ef18aff8b9542e/luigi_pipeline/seqr_loading.py#L35)  [here](https://github.com/barslmn/seqr-loading-pipelines/blob/7db47b2c0bbe01398f768888a5ef18aff8b9542e/luigi_pipeline/seqr_loading.py#L85)  instead of [check_if_file_exists](https://github.com/barslmn/seqr-loading-pipelines/blob/7db47b2c0bbe01398f768888a5ef18aff8b9542e/luigi_pipeline/seqr_loading.py#L30).

[...]

PR #632, opened 2023-11-22
packages for hail and psproc and purge the pip cache

barslmn:
Related discussion:

https://github.com/broadinstitute/seqr-loading-pipelines/issues/623#issuecomment-1822719116

https://github.com/broadinstitute/seqr-loading-pipelines/issues/624#issuecomment-1820102168

https://github.com/broadinstitute/seqr-loading-pipelines/issues/624#issuecomment-1820671218

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -22,7 +22,10 @@ RUN install_packages \
     build-essential \
     libcurl4-openssl-dev \
     libbz2-dev \
-    liblzma-dev
+    liblzma-dev \
+    procps \
+    liblapack3 \
+    libopenblas0
 
 # install java-11    (adds 340Mb)
 RUN install_packages default-jdk
@@ -94,6 +97,7 @@ RUN ln -s /ensembl-vep-release-${VEP_VERSION}/vep /vep
 WORKDIR /seqr-loading-pipelines
 COPY requirements.txt .
 RUN python3 -m pip install -r ./requirements.txt
+RUN python3 -m pip cache purge
 
 # install GCS connector using the same steps as in https://raw.githubusercontent.com/broadinstitute/install-gcs-connector/master/install_gcs_connector.py
 # assumes ~/.config/gcloud/application_default_credentials.json will be mounted into the docker container


docker/Dockerfile
```diff
@@ -94,6 +97,7 @@ RUN ln -s /ensembl-vep-release-${VEP_VERSION}/vep /vep
 WORKDIR /seqr-loading-pipelines
 COPY requirements.txt .
 RUN python3 -m pip install -r ./requirements.txt
+RUN python3 -m pip cache purge
```
bpblanken, line 100: Can we just do `pip install  --no-cache-dir ` on the line above?

barslmn: That's just a matter of choice I guess (your choice as maintainer). Either `--no-cache-dir` needs to be added to every `pip install` command or one `pip cache purge` at the end.

bpblanken: 👍 looks like the only place we're missing it would be [here](https://github.com/broadinstitute/seqr-loading-pipelines/pull/632/files#diff-f34da55ca08f1a30591d8b0b3e885bcc678537b2a9a4aadea4f190806b374ddcR60), which may not even be necessary... 🤔 (I think `gustil` comes bundled with `gcloud` but I will double check).

bpblanken: Yep, the `pip install of gsutil` appears unnecessary.

bpblanken: I'm going to approve and make that change on a different branch.

bpblanken (approved):

docker/Dockerfile
barslmn: There are two more `pip install` commands wo `--no-cache-dir`. [here](https://github.com/broadinstitute/seqr-loading-pipelines/blob/ec6b322a104c6e086636c8619d0a8196915861ae/docker/Dockerfile#L99) and [here](https://github.com/broadinstitute/seqr-loading-pipelines/blob/ec6b322a104c6e086636c8619d0a8196915861ae/docker/Dockerfile#L51)

sjahl: I know I'm late on this one (sorry), but the `pip cache purge` here is ineffective unless it's chained onto the end of the pip installs with an `&&`: The cache contents introduced by the pip installs in the previous layers will continue to exist in those layers unless it's removed in the same `RUN` directive. Clearing the cache in its own step will make the final image appear cleaner when you run it, but all of the data in the previous layers still needs to be stored in the docker repository and shipped over the network when downloading the images.

`--no-cache-dir` on every pip call will actually accomplish reducing the final image size and saving bandwidth -- setting that option in a pip.conf file could be done if adding the flag feels too repetitive.

Alternatively, switching to a multi-stage build and not worrying about cache would be my preferred solution (see how we did this over in the broadinstitute/seqr Dockerfile).

sjahl: And, I see Ben has addressed this: https://github.com/broadinstitute/seqr-loading-pipelines/commit/5fb1be85f80cb244662fda056b4b8072b650611b 👍

barslmn: Thanks, I learned something.

PR #638, opened 2023-11-29
Missing 37 logic

bpblanken:
Noticed while generating the "37" reference data that "37" does not have an interval file.  This updates the pipeline to support that case.


v03_pipeline/lib/model/cached_reference_dataset_query.py
```diff
@@ -146,18 +147,30 @@ def query(self) -> Callable[[hl.Table, ReferenceGenome], hl.Table]:
...
                 CachedReferenceDatasetQuery.CLINVAR_PATH_VARIANTS,
                 CachedReferenceDatasetQuery.HIGH_AF_VARIANTS,
                 CachedReferenceDatasetQuery.GNOMAD_CODING_AND_NONCODING_VARIANTS,
                 CachedReferenceDatasetQuery.GNOMAD_QC,
             ],
-            DatasetType.MITO: [
+            (ReferenceGenome.GRCh38, DatasetType.MITO): [
+                CachedReferenceDatasetQuery.CLINVAR_PATH_VARIANTS,
+            ],
+            (ReferenceGenome.GRCh37, DatasetType.SNV_INDEL): [
```
bpblanken, line 165: We actually have the same "cached_reference_dataset_queries" for 37 as we do for 38, so this change is purely for cosmetic consistency reasons.  For more context, the four queries we have are:
- clinvar_path_variants (used by hail search)
- high_af_variants (used by hail search)
- common coding and non-coding variants (used internvally by the sample type validation logic)
- gnomad qc (used by sex/relatedness for... something...)

v03_pipeline/lib/model/dataset_type.py
```diff
@@ -153,10 +153,12 @@ def sample_lookup_table_fields_and_genotype_filter_fns(
...
 
-    @property
-    def formatting_annotation_fns(self) -> list[Callable[..., hl.Expression]]:
+    def formatting_annotation_fns(
+        self,
+        reference_genome: ReferenceGenome,
+    ) -> list[Callable[..., hl.Expression]]:
         return {
-            DatasetType.SNV_INDEL: [
+            (ReferenceGenome.GRCh38, DatasetType.SNV_INDEL): [
```
bpblanken, line 161: I opted to just make the formatting function list a function of both `reference_genome` and `dataset_type` instead of just `dataset_type` (with some internal logic conditioning on `reference_genome` [like this](https://github.com/broadinstitute/seqr-loading-pipelines/pull/638/files#diff-bfdea0351ad0697e98a4b8fd1a7a8ace4bf559d5e611b3c69aff92d32f0cfcf7L59)).

v03_pipeline/lib/model/cached_reference_dataset_query.py
hanars: I see the advantage of the cosmetic change, but I think we would want the default behavior for new reference data sources that they be made available for both builds, so I feel like we would want some sort of shared list as the default place things go, and then that way you would need to make an explicit choice to exclude something for a given build. This could be as simple as defining a top-level constant for `SNV_INDEL_CRDQS` and then here having
```
(ReferenceGenome.GRCh38, DatasetType.SNV_INDEL): SNV_INDEL_CRDQS,
(ReferenceGenome.GRCh37, DatasetType.SNV_INDEL): SNV_INDEL_CRDQS,
```

v03_pipeline/lib/model/dataset_type.py
```diff
@@ -165,7 +167,13 @@ def formatting_annotation_fns(self) -> list[Callable[..., hl.Expression]]:
                 shared.variant_id,
                 shared.xpos,
             ],
-            DatasetType.MITO: [
+            (ReferenceGenome.GRCh37, DatasetType.SNV_INDEL): [
```
hanars, line 170: similar to my above comment, I think it would be better to have a constant for the shared fields so it would be more explicit what differs between builds

hanars (approved):

PR #639, opened 2023-11-29
Reference data "score" updates

bpblanken:
If these fields look right, we can probably squeeze this in before all the data gets loaded.


hanars (commented):
for v3 we should get rid of the _pred fields, we won't be using them so theres no need to keep them around

download_and_create_reference_datasets/v02/hail_scripts/write_dbnsfp_ht.py
```diff
@@ -68,16 +68,16 @@
         'pos(1-based)': tint,
         'ref': tstr,
         'alt': tstr,
-        'SIFT_pred': tstr,
-        'Polyphen2_HVAR_pred': tstr,
+        'SIFT_score': tstr,
```
hanars, line 71: shouldn't scores by floats?

bpblanken: they should be yes, however it looks like these are formatted in the same way as `VEST4_score` and the rest of the predictors (`'.;.;.;0.32'`). 

I tried
```
ht = hl.import_table('gs://seqr-reference-data/GRCh37/dbNSFP/v2.9.3/dbNSFP2.9.3_variant.chr*.gz', force=True, types={'SIFT_score': hl.tfloat}, missing='.')
```

and 

```
ht = hl.import_table('gs://seqr-reference-data/GRCh37/dbNSFP/v2.9.3/dbNSFP2.9.3_variant.chr*.gz', force=True, types={'SIFT_score': hl.tarray(hl.tfloat)}, missing='.')
```

Letting these into the first hail table (prior to the combine) isn't great, but is consistent with what's there.

hanars (approved):

PR #640, opened 2023-11-30
use ReferenceGenome enum in custom download functions

jklugherz:
- also added 'contig_recoding' as a property on ReferenceGenome 
- includes tests for ReferenceGenome enum properties
- moved some clinvar constants to enums.py

### DIFF
--- a/download_and_create_reference_datasets/v02/hail_scripts/write_clinvar_ht.py
+++ b/download_and_create_reference_datasets/v02/hail_scripts/write_clinvar_ht.py
@@ -12,7 +12,7 @@ from hail_scripts.utils.hail_utils import write_ht
 CLINVAR_PATH = 'ftp://ftp.ncbi.nlm.nih.gov/pub/clinvar/vcf_{reference_genome}/clinvar.vcf.gz'
 CLINVAR_HT_PATH = 'gs://seqr-reference-data/{reference_genome}/clinvar/clinvar.{reference_genome}.ht'
 
-for reference_genome in [ReferenceGenome.GRCh37, ReferenceGenome.GRCh38]:
+for reference_genome in [x for x in ReferenceGenome]:
     clinvar_url = CLINVAR_PATH.format(reference_genome=reference_genome.value)
     ht = download_and_import_latest_clinvar_vcf(clinvar_url, reference_genome)
     timestamp = hl.eval(ht.version)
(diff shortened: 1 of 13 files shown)


download_and_create_reference_datasets/v02/hail_scripts/write_clinvar_ht.py
```diff
@@ -1,18 +1,20 @@
...
 
-CLINVAR_PATH = 'ftp://ftp.ncbi.nlm.nih.gov/pub/clinvar/vcf_GRCh{genome_version}/clinvar.vcf.gz'
-CLINVAR_HT_PATH = 'gs://seqr-reference-data/GRCh{genome_version}/clinvar/clinvar.GRCh{genome_version}.ht'
+CLINVAR_PATH = 'ftp://ftp.ncbi.nlm.nih.gov/pub/clinvar/vcf_{reference_genome}/clinvar.vcf.gz'
+CLINVAR_HT_PATH = 'gs://seqr-reference-data/{reference_genome}/clinvar/clinvar.{reference_genome}.ht'
 
-for genome_version in ["37", "38"]:
-    clinvar_url = CLINVAR_PATH.format(genome_version=genome_version)
-    ht = download_and_import_latest_clinvar_vcf(clinvar_url, genome_version)
+for reference_genome in [ReferenceGenome.GRCh37, ReferenceGenome.GRCh38]:
```
bpblanken, line 15: small suggestion, enums are sequences!:

```
>>> [x for x in ReferenceGenome]
[<ReferenceGenome.GRCh37: 'GRCh37'>, <ReferenceGenome.GRCh38: 'GRCh38'>]
```

v03_pipeline/lib/model/definitions_test.py
```diff
@@ -0,0 +1,143 @@
+import unittest
+
+from v03_pipeline.lib.model.definitions import ReferenceGenome
+
+
+class TestReferenceGenomeProperties(unittest.TestCase):
```
bpblanken, line 6: 🙏

bpblanken:
I think one missing thing here is to make sure to change the calling arguments every time a `custom_import` is called.  There are a couple of places where we're doing:

```
config['custom_import'](config['source_path'], reference_genome.v02_value)
```

that should now reference the `ReferenceGenome` itself.  Our test coverage is just not close to actually getting this handled correctly (I think the one `get_ht` test we have just mocks out the download 😢) .

bpblanken:
Looks like [one more missing place](https://github.com/broadinstitute/seqr-loading-pipelines/blob/main/v03_pipeline/bin/write_cached_reference_dataset_query_ht.py#L27-L34)!  but otherwise looks great.

bpblanken (approved):

PR #641, opened 2023-12-01
Account for missing parents in pedigrees

bpblanken:
I've run into this behavior quite a bit after getting the "subsetted_pedigree" logic live on airflow.

### DIFF
--- a/v03_pipeline/lib/misc/pedigree.py
+++ b/v03_pipeline/lib/misc/pedigree.py
@@ -77,17 +77,24 @@ class Family:
         for row in rows:
             # Maternal GrandParents
             maternal_s = samples[row.s].mother
-            if maternal_s and samples[maternal_s].mother:
-                samples[row.s].maternal_grandmother = samples[maternal_s].mother
-            if maternal_s and samples[maternal_s].father:
-                samples[row.s].maternal_grandfather = samples[maternal_s].father
+            if maternal_s and maternal_s not in samples:
+                # sample may be removed from the pedigree :/
+                samples[row.s].mother = None
+            elif maternal_s:
+                if samples[maternal_s].mother:
+                    samples[row.s].maternal_grandmother = samples[maternal_s].mother
+                if samples[maternal_s].father:
+                    samples[row.s].maternal_grandfather = samples[maternal_s].father
 
             # Paternal GrandParents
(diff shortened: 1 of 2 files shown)


v03_pipeline/lib/misc/pedigree.py
```diff
@@ -64,7 +64,7 @@ def __hash__(self):
         return hash(self.family_guid)
 
     @staticmethod
-    def parse_direct_lineage(rows: list[hl.Struct]) -> dict[str, Sample]:
+    def parse_direct_lineage(rows: list[hl.Struct]) -> dict[str, Sample]:  # noqa: C901
```
bpblanken, line 67: For context, I'm running off of this code to parse the pedigrees, but the actual sex/relatedness checking is [disabled via env var](https://github.com/broadinstitute/seqr-loading-pipelines/blob/main/v03_pipeline/lib/tasks/write_remapped_and_subsetted_callset.py#L111-L114)

v03_pipeline/lib/misc/pedigree.py
```diff
@@ -77,17 +77,24 @@ def parse_direct_lineage(rows: list[hl.Struct]) -> dict[str, Sample]:
         for row in rows:
             # Maternal GrandParents
             maternal_s = samples[row.s].mother
-            if maternal_s and samples[maternal_s].mother:
-                samples[row.s].maternal_grandmother = samples[maternal_s].mother
-            if maternal_s and samples[maternal_s].father:
-                samples[row.s].maternal_grandfather = samples[maternal_s].father
+            if maternal_s and maternal_s not in samples:
+                # sample may be removed from the pedigree :/
```
hanars, line 81: what does this mean?

bpblanken: I can be more descriptive with the comment.  

It's basically just the pedigree getting subsetted from:
```
Individual_ID Paternal_ID Maternal_ID
sample_1       sample_2     sample_3
sample_2
sample_3
```

to

```
Individual_ID Paternal_ID Maternal_ID
sample_1       sample_2     sample_3
sample_3
```

so the parent is no longer present as a loadable sample.  It's also pretty straightforward to do this in airflow, just less testable.

hanars (approved):

PR #642, opened 2023-12-06
Lower block size

bpblanken:


matren395 (approved):
lgtm! let's see how it works with this

PR #643, opened 2023-12-06
Add "write all project family tables" task.

bpblanken:
This adds a new task at the "project level", which farms out the work of writing each family table to the existing `WriteFamilyTable` tasks.  This is purely to avoid overloading airflow with "family" tasks.


v03_pipeline/lib/tasks/write_family_table.py
```diff
@@ -20,22 +20,18 @@ class WriteFamilyTableTask(BaseWriteTask):
     project_remap_path = luigi.Parameter()
     project_pedigree_path = luigi.Parameter()
     ignore_missing_samples_when_subsetting = luigi.BoolParameter(
-        default=False,
```
jklugherz, line 23: why did you remove these?

v03_pipeline/lib/tasks/write_project_family_tables.py
```diff
@@ -0,0 +1,55 @@
...
+    validate = luigi.BoolParameter(
+        default=True,
+        parsing=luigi.BoolParameter.EXPLICIT_PARSING,
+    )
+    is_new_gcnv_joint_call = luigi.BoolParameter(
+        default=False,
+        description='Is this a fully joint-called callset.',
+    )
+
+    @functools.cached_property
```
jklugherz, line 36: why use this decorator here?

v03_pipeline/lib/tasks/write_family_table.py
bpblanken: `WriteFamilyTable` isn't called as its own task anymore, so the defaults aren't actually going to be used.  I generally prefer to have defaults injected in as few places as possible so that args don't get accidentally dropped along the way.

jklugherz (approved):

PR #651, opened 2023-12-20
Distinct variants only

bpblanken:
We have a single set of AnVIL project that contain duplicate variants.  Literally:

```
In [15]: mt.filter_rows(mt.locus.position == 3834167).GT.show()
+---------------+------------+--------------+==================>  (24 + 1) / 25]
| locus         | alleles    | 'sample_id'.GT |
+---------------+------------+--------------+
| locus<GRCh38> | array<str> | call         |
+---------------+------------+--------------+
| chr1:3834167  | ["T","G"]  | 1|0          |
| chr1:3834167  | ["T","G"]  | 1|1          |
+---------------+------------+--------------+
```

I think also reasonable to just throw an exception here... but that might end up being more work for us in the long run.

### DIFF
--- a/v03_pipeline/lib/misc/io.py
+++ b/v03_pipeline/lib/misc/io.py
@@ -51,7 +51,8 @@ def split_multi_hts(mt: hl.MatrixTable) -> hl.MatrixTable:
     bi = bi.annotate_rows(a_index=1, was_split=False)
     multi = mt.filter_rows(hl.len(mt.alleles) > BIALLELIC)
     split = hl.split_multi_hts(multi)
-    return split.union_rows(bi)
+    mt = split.union_rows(bi)
+    return mt.distinct()
 
 
 def import_gcnv_bed_file(callset_path: str) -> hl.MatrixTable:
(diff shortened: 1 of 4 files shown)


hanars:
I think we should throw an exception here, we don't know what this means and I wouldn't want these counting toward our global AFs. I think if we really start to have too much work communicating errors in peoples VCFs back to them, the solution is to automate that communication, not just accept crappy data. 
Which project is this? I'd like to see how its currently working in seqr

hanars:
unless there was handling for this in the v2 pipeline, in which case I am okay maintaining that behavior

bpblanken:
A few things:

1) Re the v2 pipeline.... we have no 'distinct' handling there from what I can tell.
2) Only one of the rows will be counted in the global AFs, as we do an inner join, fwiw.

v03_pipeline/lib/misc/io.py
```diff
@@ -51,7 +51,8 @@ def split_multi_hts(mt: hl.MatrixTable) -> hl.MatrixTable:
     bi = bi.annotate_rows(a_index=1, was_split=False)
     multi = mt.filter_rows(hl.len(mt.alleles) > BIALLELIC)
     split = hl.split_multi_hts(multi)
-    return split.union_rows(bi)
+    mt = split.union_rows(bi)
+    return mt.distinct_by_row()
```
bpblanken, line 55: Leaving this here because these ran without validation enabled.

hanars:
This variant loaded successfully in the v2 pipeline, with GT of 1|1

hanars (approved):

jklugherz (approved):

PR #652, opened 2023-12-20
Optimize the sample_lookup_table annotations.  

bpblanken:
Hail started "stack overflowing" during the variant annotations update...my best guess after debugging/checkpointing is that we're coming close to a limit within the hail/spark execution engine's compute graph.  The hail log file is _massive_ and _full_ of project names from this annotation... so I optimized it.  I have a feeling we will run into more issues here soon though, but this resolved and kept us running for now.


v03_pipeline/lib/annotations/snv_indel.py
```diff
@@ -7,38 +8,6 @@
 N_ALT_HOM = 2
 
 
-def _AC(row: hl.StructExpression) -> hl.Int32Expression:  # noqa: N802
```
bpblanken, line 10: These functions were originally split out as separate annotations.  Good for modularity but bad for performance.

hanars (approved):

jklugherz (approved):

PR #654, opened 2024-01-01
add unit tests for remap/subset sample ids

jklugherz:


v03_pipeline/lib/misc/sample_ids.py
```diff
@@ -20,18 +18,23 @@ def remap_sample_ids(
     ignore_missing_samples_when_remapping: bool,
 ) -> hl.MatrixTable:
     mt = vcf_remap(mt)
-    collected_remap = project_remap_ht.collect()
-    s_dups = [k for k, v in Counter([r.s for r in collected_remap]).items() if v > 1]
+
+    remap_count_by_s = project_remap_ht.aggregate(hl.agg.counter(project_remap_ht.s))
```
jklugherz, line 22: we can use hail's `aggregate` function to count duplicates without collecting into python lists

v03_pipeline/lib/misc/sample_ids_test.py
```diff
@@ -0,0 +1,94 @@
+import unittest
+
+import hail as hl
+from hail import Struct
```
bpblanken, line 4: this stands out to me as being off style-wise... I've always seen `hl.Struct` used. Though I don't really know if there's actually codified style rules or just conventions 🤔.

v03_pipeline/lib/misc/sample_ids.py
bpblanken: 👍 love the idea!

I think we want to be careful, though, about the performance of calling the `.aggregate()` method twice rather than the single `.collect()`.

I think the idea behind the python Counter here was to take advantage of our knowledge that the number of samples we have is reasonably small (<5k or so) and fits into memory in python without much worry.  in memory python is _way_ faster than hail.

Some local timings:

```
In [19]: %timeit collected_remap = project_remap_ht.collect(); counted = Counter(collected_remap)
5.91 ms ± 320 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)

In [20]: %timeit project_remap_ht.aggregate(hl.agg.counter(project_remap_ht.s))
346 ms ± 8.78 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

In [21]: %timeit project_remap_ht.aggregate(hl.agg.counter(project_remap_ht.seqr_id))
358 ms ± 30.9 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

v03_pipeline/lib/misc/sample_ids.py
```diff
@@ -46,7 +49,7 @@ def remap_sample_ids(
             raise MatrixTableSampleSetError(message, missing_samples)
 
     mt = mt.annotate_cols(**project_remap_ht[mt.s])
-    remap_expr = hl.cond(hl.is_missing(mt.seqr_id), mt.s, mt.seqr_id)
+    remap_expr = hl.if_else(hl.is_missing(mt.seqr_id), mt.s, mt.seqr_id)
```
bpblanken, line 52: 🙇

v03_pipeline/lib/misc/sample_ids.py
jklugherz: this all makes sense and wow, it is actually so much faster to use `collect`. I also got those timings with a tiny remap table. I hadn't seen `timeit` before but I'll definitely be using it again.

what was the size of the `project_remap_ht` you used for testing here?

v03_pipeline/lib/misc/sample_ids_test.py
jklugherz: good catch I'm using `hl.Struct` instead now

bpblanken (approved):

PR #655, opened 2024-01-02
use ruff only for linting and formatting (to match airflow repo)

jklugherz:


pyproject.toml
```diff
@@ -14,39 +14,39 @@ include = ["v03_pipeline*"]
...
-    'gcloud_dataproc/*',
-    'hail_scripts/computed_fields/*',
-    'hail_scripts/elasticsearch/*',
-    'hail_scripts/update_models/*',
-    'hail_scripts/utils/*',
-    'kubernetes/*'
-]
+#[tool.isort]
+#include_trailing_comma = 'True'
+#known_external_package = ['hail_scripts']
```
bpblanken, line 37: I love that I played this weird game with the `hail_scripts` import and now it's just 💀

pyproject.toml
```diff
@@ -14,39 +14,39 @@ include = ["v03_pipeline*"]
...
+#line_length = 88
+#profile = 'black'
+#sections = ['FUTURE', 'STDLIB', 'THIRDPARTY', 'EXTERNAL_PACKAGE', 'FIRSTPARTY', 'LOCALFOLDER']
```
jklugherz, line 40: ruff doesn't have as many configuration options for sorting as isort (no `sections` or `known_external_package` equivalent), so is it still okay to remove?

bpblanken: Yeah, I think so!  The whole point of the "sections" bit was to deal with `isort` doing the wrong thing with `hail_scripts`... but that dependency was removed from the `v03_pipeline` and we gave up on linting the `luigi_pipeline`.

bpblanken (approved):

PR #656, opened 2024-01-03
create updated reference dataset collection task

jklugherz:
- new luigi task `UpdatedReferenceDatasetCollectionTask`
- rename `combine.py` to `dataset_table_operations.py`, consolidate join/update logic to 1 function
- unit tests for both


v03_pipeline/lib/tasks/reference_data/updated_reference_dataset_collection_task.py
```diff
@@ -0,0 +1,206 @@
...
+from v03_pipeline.lib.tasks.base.base_update_task import BaseUpdateTask
+from v03_pipeline.lib.tasks.files import GCSorLocalTarget
+
+# NOTE: this is intended to replace the code in v03_pipeline/bin/write_combined_reference_ht.py, v03_pipeline/bin/write_combined_interval_reference_ht.py, and combine.py
+
+class UpdatedReferenceDatasetCollectionTask(BaseUpdateTask):
+    reference_dataset_collection = luigi.EnumParameter(enum=ReferenceDatasetCollection)
+    dataset = luigi.OptionalStrParameter()
+
+    @cached_property
```
bpblanken, line 22: I think this is probably not necessary here, though the idea is correct.  iirc the place I used this recently involved a hail update of some kind... this is just a string path join under the hood and I don't think needs to be cached.

v03_pipeline/lib/tasks/reference_data/updated_reference_dataset_collection_task.py
```diff
@@ -0,0 +1,206 @@
...
+                key_type,
+                key=key_type.fields,
+                globals=hl.Struct(
+                    paths=hl.Struct(),
+                    versions=hl.Struct(),
+                    enums=hl.Struct(),
+                ),
+            )
+
+        if not hl.hadoop_exists(os.path.join(self._destination_path, '_SUCCESS')):
```
bpblanken, line 47: I think you can probably do something like `GCSorLocalFolderTarget(self._destination_path).exists()` here.  It does literally the same thing as `hl.hadoop_exists` though... but while we're baked into `luigi` might as well standardize on the features.

bpblanken: also... I _think_ even the above might be unnecessary, there's existing [logic](https://github.com/broadinstitute/seqr-loading-pipelines/blob/main/v03_pipeline/lib/tasks/base/base_update_task.py#L29-L33) in `BaseUpdateTask` that does this bit for us!

v03_pipeline/lib/tasks/reference_data/updated_reference_dataset_collection_task.py
```diff
@@ -0,0 +1,206 @@
...
+                f'Cannot find existing valid ht at ({self._destination_path}). '
+                f'Consider omitting the --dataset flag to re-initialize the ht.'
+            )
+            raise Exception(message) # TODO: new exception type?
+
+        else:
+            return hl.read_table(self._destination_path)
+
+    def update_table(self, ht: hl.Table) -> hl.Table:
+        joined_ht = ht
```
bpblanken, line 58: I bet we can get away with combining this logic now, maybe something like:

```
datasets = [self.dataset] if self.dataset is not None else self.reference_dataset_collection.datasets(self.dataset_type)

for dataset in datasets:
   dataset_ht = self._get_dataset_ht(dataset)
   // drop the field if it exists?
   // outer join
   // filter the rows where the dataset is now null
  joined_ht = self._annotate_dataset_globals(joined_ht, dataset, dataset_ht)
return joined_ht
   
```

bpblanken: if that's cleaner

v03_pipeline/lib/tasks/reference_data/updated_reference_dataset_collection_task.py
```diff
@@ -0,0 +1,206 @@
...
+                        ~hl.is_missing(joined_ht[dataset])
+                        for dataset in self.reference_dataset_collection.datasets(self.dataset_type)
+                    ],
+                ),
+            )
+            joined_ht = self._annotate_dataset_globals(joined_ht, self.dataset, dataset_ht)
+
+        return joined_ht
+
+    def _get_dataset_ht(self, dataset: str) -> hl.Table:
```
bpblanken, line 85: This is getting more opinion-ey, but I'd opt for leaving the rest of these functions in the `lib/reference_data` folder, but probably in something not named `combine.py`.  They're more general than this task in particular, and I have a feeling one or two will be be used in other tasks.  We can also just leave the existing unit tests that we have working.  

I'm a fan of the rename from `get_ht` to `get_dataset_ht` though, that's for sure a win.

v03_pipeline/lib/tasks/reference_data/updated_reference_dataset_collection_task.py
jklugherz: oh yes I missed that logic, that simplifies my implementation of `initialize_table` for sure

v03_pipeline/lib/tasks/reference_data/updated_reference_dataset_collection_task.py
jklugherz: This is something considered and I agree here - to keep these operations separate from the luigi task as they are now. I'll add the single function to create/update the joined table in that file. I've settled on renaming the file `dataset_table_operations.py` for now but am open to a different name

v03_pipeline/lib/tasks/reference_data/updated_reference_dataset_collection_task.py
jklugherz: I combined the logic into 1 function in the former 'combine' file!

v03_pipeline/lib/reference_data/dataset_table_operations.py
```diff
@@ -11,21 +12,73 @@
...
+) -> hl.Table:
+    datasets = (
+        [dataset]
+        if dataset is not None
+        else reference_dataset_collection.datasets(dataset_type)
+    )
+
+    for dataset in datasets:
+        dataset_ht = get_dataset_ht(dataset, reference_genome)
+        joined_ht = joined_ht.drop(dataset)
```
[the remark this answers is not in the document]
jklugherz: encountered this in tests and modified this logic a bit to handle missing keys

v03_pipeline/lib/tasks/reference_data/updated_reference_dataset_collection_task.py
```diff
@@ -0,0 +1,56 @@
...
+
+    @property
+    def _destination_path(self) -> str:
+        return valid_reference_dataset_collection_path(
+            self.reference_genome,
+            self.dataset_type,
+            self.reference_dataset_collection,
+        )
+
+    def complete(self) -> bool:
```
bpblanken, line 25: nit but I think just leaving the comment is fine.  the method itself, as written, is a no-op because of the inheritance.

v03_pipeline/lib/tasks/reference_data/updated_reference_dataset_collection_test.py
```diff
@@ -0,0 +1,106 @@
...
+        self,
+        mock_reference_dataset_collection_datasets,
+        mock_get_dataset_ht,
+        mock_destination_path,
+    ) -> None:
+        mock_reference_dataset_collection_datasets.return_value = ['primate_ai']
+        # delete the dataset collection directory to fail initial task 'complete()' and force 'run()'
+        # since collection won't exist, we can't test the codepath for updating an existing collection yet
+        shutil.rmtree(COMBINED_2_PATH)
+        mock_destination_path.return_value = COMBINED_2_PATH
```
bpblanken, line 41: One tiny thing here... the existing pattern for this has been to move the in-repo hts to the tmp dir setup by the `MockedDatarootTestCase`.  [here](https://github.com/broadinstitute/seqr-loading-pipelines/blob/main/v03_pipeline/lib/tasks/update_variant_annotations_table_with_new_samples_test.py#L79-L86) is an example.

The goal there has been to use a unique tmpdir for every test so that (one day) they're fully parallelizable.  We also probably don't want to just mock `valid_reference_dataset_collection_path` since it's gonna get called multiple times with multiple different arguments under the hood!

You can use moving the file as a proxy for "has the task run yet" in the way you sorta are now.  There's an existing example [here](https://github.com/broadinstitute/seqr-loading-pipelines/blob/main/v03_pipeline/lib/tasks/update_variant_annotations_table_with_new_samples_test.py#L138-L159).

v03_pipeline/lib/tasks/reference_data/updated_reference_dataset_collection_test.py
```diff
@@ -0,0 +1,106 @@
...
+                    alleles=['A', 'C'],
+                    primate_ai=hl.Struct(score=0.5),
+                ),
+            ],
+        )
+        self.assertEqual(
+            ht.globals.collect(),
+            [
+                hl.Struct(
+                    paths=hl.Struct(
```
bpblanken, line 98: It'd be awesome to have a test that updates the existing reference dataset collection with a dataset that is present on the ht already!

v03_pipeline/lib/tasks/reference_data/updated_reference_dataset_collection_task.py
jklugherz: removed the comment in the code, for now it lives in my brain

v03_pipeline/lib/tasks/reference_data/updated_reference_dataset_collection_test.py
jklugherz: thanks for linking examples, helped me figure out the missing shutil  piece. I was close!

v03_pipeline/lib/tasks/reference_data/updated_reference_dataset_collection_test.py
jklugherz: I added 2 tests: 
1. initializing the task without a `dataset` param should create a new ht with all datasets
2. a test which updates existing reference data collection ht for a dataset that exists on the ht already (actually covers these cases - updating an existing reference dataset collection ht in general, and dropping a field that already exists on it)

v03_pipeline/lib/tasks/reference_data/updated_reference_dataset_collection_test.py
```diff
@@ -0,0 +1,328 @@
...
+    @mock.patch.object(ReferenceDatasetCollection, 'datasets')
+    def test_update_task_with_dataset_and_existing_reference_dataset_collection_table(
+        self,
+        mock_reference_dataset_collection_datasets,
+        mock_get_dataset_ht,
+        mock_initialize_table,
+    ) -> None:
+        delete_reference_data_ht()
+        # make initialize_table() return existing reference dataset collection table
+        mock_initialize_table.return_value = hl.Table.parallelize(
```
bpblanken, line 224: We can just... not call `delete_reference_data_ht` and not mock here right?

v03_pipeline/lib/tasks/reference_data/updated_reference_dataset_collection_test.py
bpblanken: Yep!  These look awesome.  🙇

v03_pipeline/lib/tasks/reference_data/updated_reference_dataset_collection_test.py
bpblanken: Ah.  I see.. we're ending up with weird stuff in here because our `def complete()` isn't implemented yet.  In that case, maybe not having a `setUp` and having `mock_initialize_table.return_value = hl.read_table(OMBINED_2_PATH)` here works?

jklugherz: even so, since `complete()` is still the base class logic, if the file exists at `COMBINED_2_PATH`, initialize_table() won't run right?

bpblanken: ah, I don't think so (at least, unless we're mocking the `valid_reference_dataset_collection` path).  `complete()` won't run if whatever `valid_reference_dataset_collection` points at exists... and that only exists if we `shutil.copy` it from the repo into the tmp dir.

jklugherz: oh you're right!

bpblanken: yeah, for some additional context on why it is this way...

we want to be able to test the pipeline under different reference dataset collection cases, but within the pipeline, there is only a single reference dataset collection (with a static path) per type.  That allows us to now have to pass a file path through the whole pipeline, like [v02 was doing](https://github.com/broadinstitute/seqr-loading-pipelines/blob/main/luigi_pipeline/seqr_loading.py#L46-L49)

bpblanken (approved):

hanars (approved):

PR #669, opened 2024-01-18
add complete() logic to UpdatedReferenceDatasetCollectionTask, including check globals against config

jklugherz:
- defined complete() function in UpdatedReferenceDatasetCollectionTask
- if reference dataset collection table exists, check validate dataset globals to determine which datasets to update


v03_pipeline/lib/reference_data/config.py
```diff
@@ -211,6 +224,12 @@ def custom_mpc_select(ht):
             'version': '2.9.3',
             'path': 'gs://seqr-reference-data/GRCh37/dbNSFP/v2.9.3/dbNSFP2.9.3_variant.with_new_scores.ht',
             'custom_select': dbnsfp_custom_select,
+            'custom_select_keys': [
```
jklugherz, line 227: I added `custom_select_keys` because I needed a way to get these keys for validation and couldn't get them from the `custom_select` functions.

v03_pipeline/lib/reference_data/dataset_table_operations.py
```diff
@@ -173,6 +177,113 @@ def annotate_dataset_globals(joined_ht: hl.Table, dataset: str, dataset_ht: hl.T
...
+        set(raw_config_selects)
+        if isinstance(raw_config_selects, list)
+        else set(raw_config_selects.keys())
+    )
+
+    if 'custom_select' in dataset_config:
+        config_selects.update(set(dataset_config.get('custom_select_keys')))
+
+    # check if all fields in 'config_selects' are in 'joined_ht_selects'
+    return config_selects.issubset(joined_ht_selects)
```
bpblanken, line 284: Is the idea here that we would want to leave the joined_ht alone if we've removed a select?

v03_pipeline/lib/reference_data/config.py
bpblanken: Yep!  Love the idea.  Would this capture a change to the logic in a custom select though? Say we change:

```
selects['MutPred_score'] = hl.parse_float32(ht.MutPred_score)
```

to 

```
selects['MutPred_score'] = hl.int32(hl.parse_float32(ht.MutPred_score))
```
or something.

v03_pipeline/lib/reference_data/dataset_table_operations.py
```diff
@@ -49,11 +46,7 @@ def get_dataset_ht(
     reference_genome: ReferenceGenome,
 ) -> hl.Table:
     config = CONFIG[dataset][reference_genome.v02_value]
-    ht = (
```
bpblanken, line 52: I think there's code in `bin/write_cached_reference_dataset_query_ht.py` that could use this function (it'll be dead soon but good to be complete!).

v03_pipeline/lib/reference_data/dataset_table_operations.py
```diff
@@ -173,6 +177,113 @@ def annotate_dataset_globals(joined_ht: hl.Table, dataset: str, dataset_ht: hl.T
...
+    return all(results)
+
+
+def _ht_version_matches_config(
+    joined_ht: hl.Table,
+    dataset: str,
+    dataset_config: dict,
+    reference_genome: ReferenceGenome,
+) -> bool:
+    joined_ht_version = hl.eval(
```
bpblanken, line 214: One quick hail performance tip.. you want t run `hl.eval` as new times as possible.  So my guess is that writing these functions to take in a `joined_ht`'s evaluated globals is going to be much nicer and faster!

v03_pipeline/lib/reference_data/config.py
jklugherz: Since we're checking the config selects the joined table selects, we only need to check if the fields from or the left hand of that equals expression from your comment are in the joined table. 

Do we also need to compare the fields on the right hand side of the expression with the fields on the dataset tables?

jklugherz: it does make sense to couple this list of keys with the `custom_select` functions though so if they ever change they only change in one spot

v03_pipeline/lib/reference_data/dataset_table_operations.py
jklugherz: now that you bring this up, we should probably remove it from the joined_ht too, right?

bpblanken: yep!  I think any change to the selects or custom selects should trigger a drop and re-join of the dataset from the `joined_ht`.

v03_pipeline/lib/reference_data/config.py
bpblanken: I think we probably should do something to ensure the logic of the function itself has changed or not changed.  My initial thought would be to do something like hash the definition of the function itself so, so if we change the source code in any way we'd be "out of sync".  It might be over-engineering (or even under-engineering) though.

Something like this works:
```
def f():
	pass

import inspect
print(inspect.getsource(f))
```

v03_pipeline/lib/tasks/reference_data/updated_reference_dataset_collection.py
```diff
@@ -1,19 +1,28 @@
...
 )
 from v03_pipeline.lib.tasks.base.base_update_task import BaseUpdateTask
 from v03_pipeline.lib.tasks.files import GCSorLocalTarget
 
+logger = logging.getLogger(__name__)
+
 
 class UpdatedReferenceDatasetCollectionTask(BaseUpdateTask):
     reference_dataset_collection = luigi.EnumParameter(enum=ReferenceDatasetCollection)
     dataset = luigi.OptionalStrParameter(default=None)
```
bpblanken, line 21: We don't even need `dataset` anymore here do we?

v03_pipeline/lib/reference_data/dataset_table_operations_test.py
```diff
@@ -15,10 +15,15 @@
     dbnsfp_mito_custom_select,
 )
 from v03_pipeline.lib.reference_data.dataset_table_operations import (
+    _ht_enums_match_config,
```
bpblanken, line 18: This is kinda a me not loving underscore prefixes outside of classes thing, but what if we just moved these compare functions to a new module?  `dataset_table_operations` is getting big and these functions have a shared purpose.

jklugherz: That's a good way to think about what should be a class. I was thinking of creating a `Validator` class that would have `joined_ht_globals` as a class attribute or something, but 
- it's not necessarily an improvement from the functional way I've written it now
- it doesn't follow the pattern you described of what actually should be an object
- it'd diverge from codebase standards
- it'd make moving my tests to a new file into a bigger refactor, which I actually don't want 
so I'll just move them and tweak the logic per your other comments

bpblanken: Yep!  I think everything you've mentioned makes lots of sense.  I'd actually love for the dictionaries we have in the config (with the schema that's loosely defined in the big comment) and on the globals to be more defined (maybe something like a [pydantic](https://docs.pydantic.dev/latest/) model or a data class or something), which I think does pave the way for having a more class based approached.

v03_pipeline/lib/reference_data/config.py
jklugherz: if we went this route, would we store the hashed function definition in the config?

v03_pipeline/lib/reference_data/dataset_table_operations.py
jklugherz: instead of `issubset()`, I'm using `symmetric_difference()` which should trigger a drop and rejoin if either set contains a field not in the other

v03_pipeline/lib/reference_data/dataset_table_operations.py
jklugherz: I decided to return a `hl.BooleanExpression` from the each of the check functions and call `hl.eval()` on the result of each of those functions, but I'll see if I can get away with calling it once on joined_ht's evaluated globals

v03_pipeline/lib/tasks/reference_data/updated_reference_dataset_collection.py
jklugherz: removed `dataset` task parameter

v03_pipeline/lib/tasks/reference_data/updated_reference_dataset_collection.py
```diff
@@ -22,6 +33,56 @@ def _destination_path(self) -> str:
...
+        logger.info(
+            f'Reading existing reference dataset collection from {self._destination_path}',
+        )
+        joined_ht = hl.read_table(self._destination_path)
+        for dataset in (
+            [self.dataset]
+            if self.dataset is not None
+            else self.reference_dataset_collection.datasets(self.dataset_type)
+        ):
+            if dataset not in joined_ht.row:
```
bpblanken, line 62: I think it'd be cleaner to push this logic, and the `for dataset in 'datasets'` loop into the `validate_joined_ht_globals_match_config`.  To me, the "dataset present in one but not in the other" case should be treated as a validation step similar to the others.  It also makes unit testing this case easier as it no longer is bundled into a luigi method.

v03_pipeline/lib/tasks/reference_data/updated_reference_dataset_collection.py
```diff
@@ -1,19 +1,30 @@
...
 from v03_pipeline.lib.tasks.files import GCSorLocalTarget
 
+logger = logging.getLogger(__name__)
+
 
 class UpdatedReferenceDatasetCollectionTask(BaseUpdateTask):
     reference_dataset_collection = luigi.EnumParameter(enum=ReferenceDatasetCollection)
     dataset = luigi.OptionalStrParameter(default=None)
 
+    _should_validate: bool = True
```
bpblanken, line 25: Is it possible to remove this variable now?  

1) The joined_ht doesn't exist so `complete()` returns True.
2) We want to run the validation and return True/False depending on it's computation, resulting in us either running the `run()` method or being actually complete.

v03_pipeline/lib/tasks/reference_data/updated_reference_dataset_collection.py
jklugherz: good call, I agree

v03_pipeline/lib/reference_data/compare_globals.py
```diff
@@ -0,0 +1,142 @@
...
+def ht_path_matches_config(
+    joined_ht_globals: ReferenceDataGlobals,
+    dataset: str,
+    dataset_config: dict,
+) -> bool:
+    joined_ht_path = joined_ht_globals.paths.get(dataset)
+    if joined_ht_path is None:
+        return False
+
+    config_path = (
```
bpblanken, line 106: Should `import_ht_from_config_path` be used here?

v03_pipeline/lib/tasks/reference_data/updated_reference_dataset_collection.py
jklugherz: yep, I removed the variable and simplified the logic in `complete()`

v03_pipeline/lib/reference_data/dataset_table_operations_test.py
jklugherz: Yeah, with the current structure of `CONFIG`, rules like if `custom_import` exists then `source_path` should too aren't really enforced. I think pydantic is the right choice because it has nice syntax for writing separate validation functions with `@validator` decorator that enforce these rules.

v03_pipeline/lib/reference_data/compare_globals.py
jklugherz: Good point, I moved this chunk of code into a fn `get_ht_path()` which `import_ht_from_config_path` also calls.

v03_pipeline/lib/reference_data/compare_globals.py
```diff
@@ -0,0 +1,156 @@
...
+    dataset_ht = import_ht_from_config_path(dataset_config, reference_genome)
+    try:
+        config_or_dataset_version = hl.eval(
+            parse_dataset_version(
+                dataset_ht,
+                dataset,
+                dataset_config,
+            ),
+        )
+    except HailUserError:
```
bpblanken, line 105: hrm... I'm 🤔 hard on this one.... but I think we might actually want to just let the error ring out loudly here.  This basically means we've misconfigured one of the datasets (the path is pointing to a file with a mismatching version from what's in the config).

v03_pipeline/lib/reference_data/config.py
jklugherz: this is the only pending piece for me on this issue

bpblanken: I think we'd need to store it in the globals actually.  It actually might be worth chatting about this offline

jklugherz: We chatted offline and decided only to check the resulting selected fields and _not_ the expressions that select the fields from the dataset_ht.

v03_pipeline/lib/reference_data/compare_globals.py
```diff
@@ -0,0 +1,143 @@
...
+            datasets_to_update.append(dataset)
+    return datasets_to_update
+
+
+def validate_joined_ht_globals_match_config(
+    joined_ht: hl.Table,
+    dataset: str,
+    reference_genome: ReferenceGenome,
+) -> bool:
+    joined_ht_globals = ReferenceDataGlobals(hl.eval(joined_ht.index_globals()))
```
bpblanken, line 64: Can we lift this eval up one more function?  I think we want to call the eval of the globals just one single time, not once per dataset!

v03_pipeline/lib/reference_data/compare_globals.py
```diff
@@ -0,0 +1,143 @@
...
+    dataset_config = CONFIG[dataset][reference_genome.v02_value]
+    dataset_ht = import_ht_from_config_path(dataset_config, reference_genome)
+    checks = {
+        'version': ht_version_matches_config(
```
bpblanken, line 68: These function signatures look so nice and clean! 😍

v03_pipeline/lib/tasks/reference_data/updated_reference_dataset_collection.py
```diff
@@ -22,6 +30,34 @@ def _destination_path(self) -> str:
...
+        joined_ht = hl.read_table(self._destination_path)
+        self._datasets_to_update.extend(
+            get_datasets_to_update(
+                joined_ht,
+                all_datasets_for_collection,
+                self.reference_genome,
+            ),
+        )
+
+        if len(self._datasets_to_update) != 0:
```
bpblanken, line 53: This lines might be cleaner as:

```
logger.info(f'Datasets to update: {self._datasets_to_update}', # just an empty list if nothing to do.)
return not self._datasets_to_update
```

v03_pipeline/lib/reference_data/config.py
```diff
@@ -102,10 +102,11 @@ def custom_gnomad_select_v2(ht):
     """
     selects = {}
     global_idx = hl.eval(ht.globals.freq_index_dict['gnomad'])
-    selects['AF'] = hl.float32(ht.freq[global_idx].AF)
-    selects['AN'] = ht.freq[global_idx].AN
-    selects['AC'] = ht.freq[global_idx].AC
-    selects['Hom'] = ht.freq[global_idx].homozygote_count
+    freq = ht.freq[global_idx]
```
bpblanken, line 105: can we actually run a test to make sure  that this is faster?  We can do it locally with a hl.Table.parallelize and just see what hail logs.

v03_pipeline/lib/reference_data/compare_globals.py
jklugherz: so true, will do

v03_pipeline/lib/reference_data/compare_globals.py
jklugherz: my code in this file satisfies me

v03_pipeline/lib/reference_data/config.py
jklugherz: mind showing me how you'd do this?

bpblanken: Yeah maybe something like:

```
table1 = hl.Table.parallelize(...)
table2 = hl.Table.parallelize(...)
selects = {}
select['a'] = table1[table2.key].a
select['a'] = table1[table2.key].a

Comment (jklugherz) [reply v03_pipeline/lib/reference_data/config.py:105]: thanks, how would you run it/see the logs?

File: v03_pipeline/lib/tasks/reference_data/updated_reference_dataset_collection.py [hunk 15]
```diff
@@ -22,6 +30,29 @@ def _destination_path(self) -> str:
...
+        all_datasets_for_collection = self.reference_dataset_collection.datasets(
+            self.dataset_type,
+        )
+
+        if not self.output().exists():
+            logger.info('Creating a new reference dataset collection')
+            self._datasets_to_update.extend(all_datasets_for_collection)
+            return False
+
+        joined_ht = hl.read_table(self._destination_path)
```
bpblanken, line 44: You can also you `self.output().path` instead of `_destination_path`... I think I'm not particularly opinionated on this one since I think you'll see both patterns in the tasks thus far.

jklugherz: ok yeah I removed the extra `destination_path` property since it's kind of redundant

v03_pipeline/lib/tasks/reference_data/updated_reference_dataset_collection.py
```diff
@@ -22,6 +30,29 @@ def _destination_path(self) -> str:
             self.reference_dataset_collection,
         )
 
+    def complete(self) -> bool:
+        self._datasets_to_update.clear()
+        all_datasets_for_collection = self.reference_dataset_collection.datasets(
+            self.dataset_type,
+        )
+
+        if not self.output().exists():
```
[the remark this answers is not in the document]
jklugherz: makes sense, I changed this l ine to use `super().complete()`

bpblanken (approved):
perfect!

PR #670, opened 2024-01-19
Filter training projects

bpblanken:


v03_pipeline/lib/annotations/mito.py
```diff
@@ -90,12 +66,26 @@ def rsid(ht: hl.Table, **_: Any) -> hl.Expression:
     return ht.rsid.find(lambda x: hl.is_defined(x))
 
 
-def gt_stats(ht: hl.Table, sample_lookup_ht: hl.Table, **_: Any) -> hl.Expression:
+def gt_stats(
```
bpblanken, line 69: this is just a refactor to make the mito gt_stats work/look the same as the snv_indel one (which was re-worked for performance issues that we don't have for mito).

v03_pipeline/lib/annotations/snv_indel_test.py
```diff
@@ -25,33 +25,37 @@ def test_allele_count_annotations(self) -> None:
             [
                 {
                     'id': 0,
-                    'ref_samples': hl.Struct(project_1={'a', 'c'}, project_2=set()),
-                    'het_samples': hl.Struct(project_1={'b', 'd'}, project_2=set()),
-                    'hom_samples': hl.Struct(project_1={'e', 'f'}, project_2=set()),
+                    'ref_samples': hl.Struct(project_1={'a', 'c'}, project_2=set(), R0607_gregor_training_project_=set()),
+                    'het_samples': hl.Struct(project_1={'b', 'd'}, project_2=set(), R0607_gregor_training_project_=set()),
+                    'hom_samples': hl.Struct(project_1={'e', 'f'}, project_2=set(), R0607_gregor_training_project_={'l', 'm'}),
```
hanars, line 30: I'm a bit confused as to how this test is working... shouldn;t we be testing that the samples are NOT added?

bpblanken: Ah.  So I wrote this so that the projects are added to the lookup table, and we just ignore them at gt_stats lookup time.  This means we're doing the unnecessary work of touching the lookup table.  

I went this route because it was conceptually simpler and easier to unit test, and I didn't think about trying to optimize away the lookup table work.  The code to skip adding to the lookup table itself isn't much harder to write though.

hanars: no this is fine. we are going to rearchitect this whole table anyways, and when we rebuild it we can figure out what the better long term approach is

hanars (approved):

v03_pipeline/lib/annotations/snv_indel_test.py
bpblanken: yeah, that was my thought as well!

PR #672, opened 2024-01-24
update task requirements from HailTableTask with UpdatedReferenceDatasetCollectionTask

jklugherz:


v03_pipeline/lib/tasks/base/base_variant_annotations_table.py
```diff
@@ -21,12 +23,11 @@ def output(self) -> luigi.Target:
 
     def requires(self) -> list[luigi.Task]:
         return [
-            HailTableTask(
-                valid_reference_dataset_collection_path(
-                    self.reference_genome,
-                    self.dataset_type,
-                    rdc,
-                ),
+            UpdatedReferenceDatasetCollectionTask(
```
bpblanken, line 26: Yep this look like the right start to me.  I'd grep for more though!

jklugherz: talked offline. I found other references to `HailTableTask` but they're for the cached query reference dataset, for which we have a refactor planned later as part of this epic.

v03_pipeline/lib/tasks/base/base_variant_annotations_table_test.py
```diff
@@ -49,7 +50,12 @@ def setUp(self) -> None:
             ),
         )
 
-    def test_should_create_initialized_table(self) -> None:
+    @mock.patch(
+        'v03_pipeline.lib.tasks.reference_data.updated_reference_dataset_collection.get_datasets_to_update',
+    )
+    def test_should_create_initialized_table(self, mock_get_datasets_to_update) -> None:
+        mock_get_datasets_to_update.return_value = []
```
bpblanken, line 57: Yep this looks right to me!  What does this method return without the mock though!

jklugherz: forgot to respond to this - if I don't mock the task it tries to read the dataset ht tables which are not mocked out and errors trying to read a file from google

PR #674, opened 2024-01-29
Add callset annotation to imported callset

bpblanken:


v03_pipeline/lib/tasks/write_imported_callset.py
```diff
@@ -104,5 +104,7 @@ def create_table(self) -> hl.MatrixTable:
                 self.sample_type,
             )
         return mt.annotate_globals(
+            callset_path=self.callset_path,
```
bpblanken, line 107: Makes it more clear what the source of the callset was.

matren395: this will be the path passed in the DAG, yeah ?

matren395 (approved):

v03_pipeline/lib/tasks/write_imported_callset.py
bpblanken: yep

PR #676, opened 2024-01-30
new task for updating annotations w/ rdc data 

jklugherz:
- new task `UpdateVariantAnnotationsTableWithUpdatedReferenceDataset` becomes a dependency of `UpdateVariantAnnotationsTableWithNewSamplesTask`
- `UpdatedReferenceDatasetCollectionTask` becomes a dependency of `BaseVariantAnnotationsTableTask`


v03_pipeline/lib/tasks/base/base_variant_annotations_table_test.py
```diff
@@ -49,7 +51,11 @@ def setUp(self) -> None:
             ),
         )
 
-    def test_should_create_initialized_table(self) -> None:
+    @patch(
+        'v03_pipeline.lib.tasks.base.base_variant_annotations_table.UpdatedReferenceDatasetCollectionTask',
+    )
```
jklugherz, line 56: I was iffy on mocking the entire output of this task but since it's already unit tested in its own test file it made sense to me to just mock the output as complete instead of mocking all of the dataset tables, the config dict, and other task dependencies when it's not really the purpose of the test.

v03_pipeline/lib/tasks/update_variant_annotations_table_with_new_samples.py
```diff
@@ -83,63 +73,82 @@ def read_annotation_dependencies(self):
...
+                        self.callset_paths,
+                        self.project_guids,
+                        self.project_remap_paths,
+                        self.project_pedigree_paths,
+                    )
+                ],
+            )
+        upstream_table_tasks.extend(
+            [
+                UpdateVariantAnnotationsTableWithUpdatedReferenceDataset(
```
jklugherz, line 139: this seems right to me, but on some local test runs I'm getting 
```
RuntimeError: Unfulfilled dependency at run time: UpdateVariantAnnotationsTableWithUpdatedReferenceDataset_SNV_INDEL_COMBINED_GRCh38_e9bde72efb
```
so I think the way I'm initializing these dependencies could be wrong

v03_pipeline/lib/tasks/base/base_variant_annotations_table.py
```diff
@@ -7,10 +9,30 @@
...
 from v03_pipeline.lib.tasks.base.base_update_task import BaseUpdateTask
-from v03_pipeline.lib.tasks.files import GCSorLocalTarget, HailTableTask
+from v03_pipeline.lib.tasks.files import GCSorLocalTarget
+from v03_pipeline.lib.tasks.reference_data.updated_reference_dataset_collection import (
+    UpdatedReferenceDatasetCollectionTask,
+)
 
 
 class BaseVariantAnnotationsTableTask(BaseUpdateTask):
+    @cached_property
```
jklugherz, line 19: does `cached_property` make sense here? in my new task, we're reading the annotations table at least twice if the annotations table needs to be updated (in complete() and update_table())

v03_pipeline/lib/tasks/update_variant_annotations_table_with_new_samples.py
```diff
@@ -231,7 +244,7 @@ def update_table(self, ht: hl.Table) -> hl.Table:
         ):
             if rdc.requires_annotation:
                 continue
-            rdc_ht = annotation_dependencies[f'{rdc.value}_ht']
+            rdc_ht = self.rdc_annotation_dependencies[f'{rdc.value}_ht']
             new_variants_ht = new_variants_ht.join(rdc_ht, 'left')
```
jklugherz, line 248: unsure if I need to keep this step, the left join seems significant

v03_pipeline/lib/annotations/mito.py
```diff
@@ -79,9 +79,13 @@ def gt_stats(
         ref_samples_length = row.ref_samples[project_guid].length()
         heteroplasmic_samples_length = row.heteroplasmic_samples[project_guid].length()
         homoplasmic_samples_length = row.homoplasmic_samples[project_guid].length()
-        AC_het += heteroplasmic_samples_length # noqa: N806
-        AC_hom += homoplasmic_samples_length # noqa: N806
-        AN += (ref_samples_length + heteroplasmic_samples_length + homoplasmic_samples_length) # noqa: N806
+        AC_het += heteroplasmic_samples_length  # noqa: N806
```
bpblanken, line 82: Have our `ruff` settings diverged somehow? If so.. can we put these lint changes on another pr?

v03_pipeline/lib/reference_data/compare_globals.py
```diff
@@ -15,131 +19,125 @@
 logger = logging.getLogger(__name__)
 
 
+def struct_to_dict(struct: hl.Struct) -> dict:
```
bpblanken, line 22: does something like `dict(hl.Struct(a=1, b=2))` work here?

v03_pipeline/lib/reference_data/compare_globals.py
```diff
@@ -15,131 +19,125 @@
...
+        return cls(paths, versions, enums, selects)
+
+    @classmethod
+    def from_ht(
+        cls,
+        ht: hl.Table,
+        rdc: ReferenceDatasetCollection,
+        dataset_type: DatasetType,
+    ):
+        rdc_globals_struct = hl.eval(ht.index_globals())
```
bpblanken, line 72: I think you maybe only need `hl.eval(ht.globals)` here, or `ht.globals.collect()[0]`.

v03_pipeline/lib/tasks/base/base_variant_annotations_table.py
bpblanken: I think the idea is correct!  We can definitely leave it, however hail is lazy, so I generally haven't bothered caching `hl.read_table`... it's mostly just initializing the table in memory and not doing any actual meaningful work.

v03_pipeline/lib/tasks/update_variant_annotations_table_with_new_samples.py
```diff
@@ -83,63 +73,82 @@ def read_annotation_dependencies(self):
             annotation_dependencies['gencode_mapping'] = hl.literal(
                 load_gencode(GENCODE_RELEASE, ''),
             )
-
         return annotation_dependencies
 
     def requires(self) -> list[luigi.Task]:
+        upstream_table_tasks = []
```
bpblanken, line 79: You could maybe slot `UpdateVariantAnnotationsTableWithUpdatedReferenceDataset` here when initializing the list!

v03_pipeline/lib/tasks/reference_data/update_variant_annotations_table_with_updated_reference_dataset.py
```diff
@@ -0,0 +1,59 @@
...
+    BaseVariantAnnotationsTableTask,
+)
+
+
+class UpdateVariantAnnotationsTableWithUpdatedReferenceDataset(
+    BaseVariantAnnotationsTableTask,
+):
+    rdc = luigi.EnumParameter(enum=ReferenceDatasetCollection)
+
+    def complete(self) -> bool:
```
bpblanken, line 19: Is this method gonna work correctly if the reference dataset collection has been correctly updated and persisted before this task is run?

v03_pipeline/lib/reference_data/compare_globals.py
jklugherz: it does but if it's nested like this, 
```
g = hl.Struct(
                versions=hl.Struct(),
                paths=hl.Struct(
                    a='gs://a_path_OLD.ht',
                ),
                enums=hl.Struct(),
            )
print(dict(g))
```
prints {'versions': Struct(), 'paths': Struct(a='gs://a_path_OLD.ht'), 'enums': Struct()}, 

I can simplify the function a bit by converting the struct into a dict initially:
```
def struct_to_dict(struct: hl.Struct) -> dict:
    struct = dict(struct)
    for field in struct:
        if isinstance(struct[field], hl.Struct):
            struct[field] = struct_to_dict(struct[field])
    return struct
```

jklugherz: do you think it's just unnecessary to turn structs into the dicts? I think my reasoning for this was that it was easier for me to think about equality checks with just simple python data structures as opposed to hail structs

bpblanken: I'm not sure!  It looks like hail's types work just fine with `__eq__`:

```
In [5]: hl.Struct(a=1,b={'d': 2}) == hl.Struct(a=1,b={'d': 1})
Out[5]: False

In [6]: hl.Struct(a=1,b={'d': 1}) == hl.Struct(a=1,b={'d': 1})
Out[6]: True
```

v03_pipeline/lib/tasks/reference_data/update_variant_annotations_table_with_updated_reference_dataset.py
jklugherz: not sure what you mean by this, got a sec before 5 to chat?

bpblanken: I went through and re-read this before our chat so I could re-fresh myself... I think this is actually correct as is and I misunderstood it before!  sorry for the confusing question!

v03_pipeline/lib/tasks/base/base_variant_annotations_table.py
jklugherz: makes sense, I just made it a property

v03_pipeline/var/test/tasks/mock_complete_task.py
```diff
@@ -0,0 +1,6 @@
+import luigi
+
+
+class MockCompleteTask(luigi.Task):
```
bpblanken, line 4: This probably belongs in `v03_pipeline/lib/test`.  `var` is meant for miscellaneous non-code files!

v03_pipeline/lib/tasks/update_variant_annotations_table_with_new_samples.py
```diff
@@ -246,38 +251,14 @@ def update_table(self, ht: hl.Table) -> hl.Table:
...
-                    **ht.globals.versions,
-                    **rdc_globals.versions,
-                ),
-                enums=hl.Struct(
-                    **ht.globals.enums,
-                    **rdc_globals.enums,
-                ),
-            )
-        ht = annotate_enums(ht, self.reference_genome, self.dataset_type)
+        ht = self.fix_globals(ht, clear_globals=True)
```
bpblanken, line 261: Could we get rid of the `clear_globals` arg if we managed the `rdc_globals` in the method but left `annotate_enums` here?

v03_pipeline/lib/tasks/reference_data/update_variant_annotations_table_with_updated_reference_dataset.py
```diff
@@ -0,0 +1,50 @@
...
+            self.rdc,
+            annotations_ht_globals,
+            rdc_ht_globals,
+            self.dataset_type,
+        )
+        return len(updated_datasets_for_rdc) == 0
+
+    def update_table(self, ht: hl.Table) -> hl.Table:
+        rdc_ht = self.rdc_annotation_dependencies[f'{self.rdc.value}_ht']
+        rdc_datasets = self.rdc.datasets(self.dataset_type)
```
bpblanken, line 43: Do we want to touch every `dataset` here or just the ones that require an update?

bpblanken: As in... do we want to use: 

```
rdc_datasets = get_datasets_to_update(
             self.rdc,
             annotations_ht_globals,
             rdc_ht_globals,
             self.dataset_type,
         )
```

jklugherz: I think I implemented it the way I did because I thought I needed to join every dataset in the rdc - see line 49 `ht = ht.join(rdc_ht, 'left')`..

but can I do `ht = ht.join(rdc_ht[dataset], 'left')`? would it save time/computation to do a join for each dataset that needs an update?

v03_pipeline/lib/tasks/update_variant_annotations_table_with_new_samples.py
jklugherz:
```
        # 5) Fix up the globals.
        ht = ht.annotate_globals(
            paths=hl.Struct(),
            versions=hl.Struct(),
            enums=hl.Struct(),
        )
        ht = self.fix_globals(ht)
```
I realized this would probably work.

what's special about the enums in `annotate_enums`?

bpblanken: Everything in `annotate_enums` is specific to fields that are annotated from the callset VCF and aren't coming from the reference data.  

It actually looks like we've introduced weirdness by slotting the clinvar enums into that file... which should be able to resolve by the upcoming work to make cached reference dataset query a task.

v03_pipeline/lib/tasks/reference_data/update_variant_annotations_table_with_updated_reference_dataset.py
bpblanken: Yeah.  I think we only want to re-join against the field that needs it.  It _should_ reduce work!

v03_pipeline/lib/tasks/reference_data/update_variant_annotations_table_with_updated_reference_dataset.py
```diff
@@ -0,0 +1,62 @@
...
+    BaseVariantAnnotationsTableTask,
+):
+    rdc = luigi.EnumParameter(enum=ReferenceDatasetCollection)
+    _datasets_to_update: ClassVar[list[str]] = []
+
+    def complete(self) -> bool:
+        self._datasets_to_update.clear()
+
+        if not super().complete():
+            self._datasets_to_update.extend(
```
bpblanken, line 26: This might be cleaner w/ just setting to the list.  It should always be empty at this point right?

```
self._datasets_to_update = self.rdc.datasets(self.dataset_type)
```

jklugherz: we still need `extend` now that the task checks multiple rdcs

v03_pipeline/lib/model/reference_dataset_collection.py
```diff
@@ -94,3 +94,40 @@ def for_reference_genome_dataset_type(
         if not Env.ACCESS_PRIVATE_REFERENCE_DATASETS:
             return [rdc for rdc in rdcs if rdc.access_control == AccessControl.PUBLIC]
         return rdcs
+
+    @classmethod
+    def for_dataset(cls, dataset: str) -> 'ReferenceDatasetCollection':
```
bpblanken, line 99: It'd be cool if we could use the existing method so we don't have to duplicate lists.  It's basically just calling `datasets(dataset_type)` for all possible`DatasetType`?

v03_pipeline/lib/tasks/reference_data/update_variant_annotations_table_with_updated_reference_dataset.py
bpblanken: What about on line 34?

v03_pipeline/lib/model/reference_dataset_collection.py
jklugherz: ok I cleaned this up to use `datasets` function, and I added `dataset_type` as an arg to `for_dataset` since we have it from the calling task.

bpblanken (approved):
:shipit:

PR #679, opened 2024-01-30
refactor globals validation so that we can re-use this pattern

jklugherz:
I tried to think of this in a composition-y way and settled on adding classmethods to create a `Globals` object - one from a hail table and one from a config dict.

The Globals objects are used fns that perform the equality checks between the two objects. 
These classes will be used in the next step of the reference data work, namely step 4 here: https://docs.google.com/document/d/19VEv7Iy6H9kyT23tvPzQtCNK0d4n_w_TAGY6XW6tdI4/edit


v03_pipeline/lib/reference_data/compare_globals.py
```diff
@@ -15,131 +19,123 @@
...
+        return cls(paths, versions, enums, selects)
+
+    @classmethod
+    def from_ht(
+        cls,
+        ht: hl.Table,
+        rdc: ReferenceDatasetCollection,
+        dataset_type: DatasetType,
+    ):
+        rdc_globals_struct = hl.eval(ht.globals)
```
jklugherz, line 70: related comment: https://github.com/broadinstitute/seqr-loading-pipelines/pull/676/files#r1471680353

v03_pipeline/lib/reference_data/compare_globals.py
```diff
@@ -15,131 +19,123 @@
 logger = logging.getLogger(__name__)
 
 
+def struct_to_dict(struct: hl.Struct) -> dict:
```
jklugherz, line 22: related comment: https://github.com/broadinstitute/seqr-loading-pipelines/pull/676/files#r1471678310

jklugherz: ok I ended up just removing this function and using `dict()`!

v03_pipeline/lib/reference_data/compare_globals.py
```diff
@@ -16,130 +20,114 @@
...
+        enums = dict(rdc_globals_struct.enums)
+
+        selects = {}
+        for dataset in rdc.datasets(dataset_type):
+            if dataset in ht.row:
+                select = ht[dataset]
+                if isinstance(select, hl.StructExpression):
+                    selects[dataset] = set(select)
+                else:
+                    selects[dataset] = set()
```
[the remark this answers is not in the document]
jklugherz: this was the case in 1 of the test rdc tables, I don't expect it to happen in real life.

I put this code in here instead of creating new test tables or updating the ones used in the current tests, but that might be the wrong call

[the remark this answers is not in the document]
bpblanken: ah, the result of my laziness 🤣 .  how much work would it be to change it?  we can maybe make a backlog ticket and punt on it... we probably should have test data that aligns to our real data.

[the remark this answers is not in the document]
jklugherz: I think not too much work now that I am pretty familiar with the structure of the data and the tests. I'll do it as part of this PR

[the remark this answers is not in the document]
jklugherz: ok I just had to change the test_combined_1.ht table - it has 1 row with the correct # and shape of annotations

v03_pipeline/lib/reference_data/compare_globals.py
```diff
@@ -16,130 +20,114 @@
...
+        }
+
+        results = []
+        for check, result in checks.items():
+            if result is False:
+                logger.info(f'{check} mismatch for {dataset}, {self.rdc.value}')
+            results.append(result)
+        return all(results)
+
+    def _compare_versions(self, dataset: str) -> bool:
```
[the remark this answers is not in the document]
jklugherz: good idea! I implemented this with additional logic to handle the case of comparing the `selects`

v03_pipeline/lib/reference_data/compare_globals.py
```diff
@@ -16,130 +20,114 @@
...
+    @classmethod
+    def from_dataset_configs(
+        cls,
+        rdc: ReferenceDatasetCollection,
+        dataset_type: DatasetType,
+        reference_genome: ReferenceGenome,
+    ):
+        paths, versions, enums, selects = {}, {}, {}, {}
+        for dataset in rdc.datasets(dataset_type):
+            dataset_config = CONFIG[dataset][reference_genome.v02_value]
```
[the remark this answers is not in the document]
jklugherz: I'd rather keep it like it is now until I do the `CONFIG` refactor into models, which I think I'll make a backlog ticket for as I've been avoiding messing with config structure as I work thru this epic

v03_pipeline/lib/reference_data/compare_globals.py
```diff
@@ -15,131 +19,98 @@
...
+        enums = dict(rdc_globals_struct.enums)
+
+        selects = {}
+        for dataset in rdc.datasets(dataset_type):
+            if dataset in ht.row:
+                selects[dataset] = set(ht[dataset])
+        return cls(paths, versions, enums, selects)
+
+
+class GlobalsValidator:
```
bpblanken, line 77: I think this stands out to me as being a class here.  It doesn't look like we're doing anything "state"-ey with it (Maybe storing the {"dataset": result} somewhere is an option?) and we're just initializing it only to call a single method with no arguments.

v03_pipeline/lib/reference_data/compare_globals.py
```diff
@@ -15,131 +19,98 @@
...
+                ].get(dataset)
+            if result is False:
+                logger.info(f'{field.name} mismatch for {dataset}, {self.rdc.value}')
+            results.append(result)
+        return all(results)
+
+    def _compare_selects(self, dataset: str) -> bool:
+        ht1_selects = self.ht1_globals.selects.get(dataset)
+        ht2_selects = self.ht2_globals.selects.get(dataset)
+        if ht1_selects is None or ht2_selects is None:
```
bpblanken, line 114: If both of these are `None` we would want to return `True` here right?  But realistically neither should ever be `None`?

jklugherz: I added this here because the `symmetric_difference` operation throws an error if one of the sets is `None`, however you're right that if they're both `None` we'd want to return `True`. 

The case where one is `None` and one is a set would happen when we're creating the rdc table for the first time, where the `Globals` attributes for the rdc table would be empty. I can't think of a case where they'd both be `None` irl

jklugherz: defaulting to an empty set handles these cases

bpblanken: awesome! cool.  I think that makes sense.  I was wondering if we could just make use of `set`  equality in python here.

In that case we might not need to special case `_compare_selects`?

```
>>> set([3,2,2]) == {2,3}
True
```

jklugherz: you're right it really is that simple!!

v03_pipeline/lib/reference_data/compare_globals.py
jklugherz: honestly yes the class is unnecessary I just made them functions again

bpblanken:
👍 from me whenever the tests pass.  this is awesome work, and thank you for managing the update to the mocked combined table!

bpblanken (approved):

PR #683, opened 2024-02-02
Benb/prefilter more fields

bpblanken:
Add more fields to this table so we can be more specific.

I don't think table size is a particular concern right now:

```
308.24 MiB   gs://seqr-reference-data/v03/GRCh38/SNV_INDEL/cached_reference_dataset_queries/high_af_variants.ht
```


bpblanken:
This increased the size to ```
317.99 MiB   gs://seqr-reference-data/v03/GRCh38/SNV_INDEL/cached_reference_dataset_queries/high_af_variants.ht
```

PR #685, opened 2024-02-06
Family Entries Refactor

bpblanken:


v03_pipeline/lib/tasks/write_metadata_for_run.py
```diff
@@ -75,15 +75,20 @@ def run(self) -> None:
             'callsets': self.callset_paths,
             'run_id': self.run_id,
             'sample_type': self.sample_type.value,
-            'families': {},
-            'families_failed_missing_samples': {},
-            'families_failed_relatedness_check': {},
-            'families_failed_sex_check': {},
+            'family_samples': {},
```
bpblanken, line 78: I renamed these fields to be consistent with this structure across the tables.  It'll require a change to the check-new-samples job though.

v03_pipeline/lib/tasks/write_family_table.py
```diff
@@ -79,26 +79,20 @@ def create_table(self) -> hl.Table:
...
+            callset_mt,
+            get_fields(
+                callset_mt,
+                self.dataset_type.genotype_entry_annotation_fns,
+                **self.param_kwargs,
             ),
-        ).rows()
-        ht = ht.filter(ht.entries.any(self.dataset_type.sample_entries_filter_fn))
-        ht = globalize_sample_ids(ht)
+        )
```
bpblanken, line 90: This small chunk of changes allows the family table structure to stay exactly the same.

v03_pipeline/lib/tasks/write_metadata_for_run_test.py
```diff
@@ -37,12 +37,12 @@ def test_write_metadata_for_run_task(self) -> None:
                 json.load(f),
                 {
                     'callsets': [TEST_VCF],
-                    'families_failed_missing_samples': {
+                    'family_samples_failed_missing_samples': {
```
bpblanken, line 40: Looking at this with fresh eyes, there's probably a case to reformat this to:

```
"failed_family_samples" : {
    "missing_samples":
    "relatedness_check":
    "sex_check":
```

or something

v03_pipeline/lib/tasks/write_metadata_for_run.py
hanars: can you point me to an example of a run with the updated metadata?

v03_pipeline/lib/tasks/write_metadata_for_run.py
```diff
@@ -75,15 +75,20 @@ def run(self) -> None:
...
+            'family_samples_failed_missing_samples': {},
```
hanars, line 79: to clarify, is `family_samples` only the families and samples that succeed or is it all families and samples in the run, and then the failed ones are listed in the failed directory

v03_pipeline/lib/misc/family_entries.py
```diff
@@ -0,0 +1,154 @@
...
+            for family_guid, sample_ids in hl.eval(mt.family_samples).items()
+            for s in sample_ids
+        },
+    )
+    ht = mt.select_rows(
+        filters=mt.filters.difference(dataset_type.excluded_filters),
+        family_entries=(
+            # NB: we're sorted by both family and sample when this runs.
+            # However, the sort is not guaranteed once the entries
+            # table is editted and families are spliced out and re-appended.
```
hanars, line 23: if the sort is not maintained in the long run, is there an advantage to doing it at all, or is it just a slight performance drain?

v03_pipeline/lib/misc/family_entries.py
```diff
@@ -0,0 +1,154 @@
...
+            ht.family_entries.map(
+                lambda fe: hl.or_missing(
+                    fe.any(dataset_type.family_entries_filter_fn),
+                    fe,
+                ),
+            )
+        ),
+    )
+    # Only keep rows where at least one family is not missing.
+    return ht.filter(ht.family_entries.any(lambda fe: ~hl.is_missing(fe)))
```
hanars, line 54: nitpick, but I feel like `hl.is_defined` is cleaner than `~hl.is_missing`, especially because you then no longer need the lambda

v03_pipeline/lib/misc/family_entries.py
```diff
@@ -0,0 +1,154 @@
...
+                        family_guid=ht.family_guids[i],
+                    ),
+                ),
+            )
+        ),
+    )
+    return ht.drop('family_guids', 'family_samples')
+
+
+def splice_new_callset_family_guids(
```
hanars, line 96: `splice` is a little confusing of a name to me, it makes me think of splicing into an array but this is updating arrays and a dictionary. Maybe `merge` or `combine` instead?

v03_pipeline/lib/misc/family_entries.py
bpblanken: I'm not convinced we need it at all, but it does add some determinism to the method (since we group by family_guid into a dictionary and then flatten the values).  

It's not (or at least shouldn't be) a performance drain.

v03_pipeline/lib/misc/family_entries.py
bpblanken: Yep that's fair.  I can just rename it to "remove_"

v03_pipeline/lib/tasks/write_metadata_for_run.py
bpblanken: I'd need to actually run this first, but I can when we have it!

v03_pipeline/lib/tasks/write_metadata_for_run.py
bpblanken: `family_samples` is only the families and samples that succeed.

hanars (approved):

PR #687, opened 2024-02-07
migration of project tables to `family_entries` schema.

bpblanken:


v03_pipeline/bin/migration.py
```diff
@@ -0,0 +1,587 @@
...
+
+
+Env.HAIL_TMPDIR = 'gs://seqr-scratch-temp'
+
+MIGRATIONS = [
+    (
+        DatasetType.SNV_INDEL,
+        ReferenceGenome.GRCh37,
+        [
+            'gs://seqr-hail-search-data/v03/GRCh37/SNV_INDEL/projects/R0200_bonnemann_int92_ext123s.ht',
```
hanars, line 35: I feel like it would be more precise if we di a `gsutil ls` and then ran this on all the projects that are there, instead of hardcoding a list?

bpblanken: Yeah, that's reasonable.  I was just.... incredibly lazy here.

PR #688, opened 2024-02-08
delete bin/ scripts and fn

jklugherz:


v03_pipeline/lib/reference_data/dataset_table_operations.py
```diff
@@ -205,25 +205,3 @@ def join_hts(
         joined_ht = joined_ht.join(dataset_ht, 'outer')
         joined_ht = annotate_dataset_globals(joined_ht, dataset, dataset_ht)
     return joined_ht
-
-
-def update_existing_joined_hts(
```
bpblanken, line 210: can `join_hts` also go?

jklugherz: it's used in a few v02 scripts

jklugherz: image

jklugherz: I could move it into a file in the v02 directory so that when that stuff gets deleted (are there plans to do that?) we won't have to go searching in the v03 directory

bpblanken: let's just leave it for now I guess!  I... think nothing in v02 will even work anymore but might as well keep them around just in case.

bpblanken (approved):

PR #690, opened 2024-02-09
Bugfixes for reference data update

bpblanken:


jklugherz (approved):
thanks for figuring these out!

jklugherz (approved):
everything here lgtm, no changes needed from me

PR #696, opened 2024-02-16
add UpdatedCachedReferenceDatasetQuery

bpblanken:


v03_pipeline/lib/reference_data/compare_globals.py
```diff
@@ -103,9 +99,13 @@ def validate_globals_match(
     ht1_globals: Globals,
     ht2_globals: Globals,
     dataset: str,
+    validate_selects: bool = True,
 ) -> bool:
     results = []
     for field in dataclasses.fields(Globals):
+        if field.name == 'selects' and not validate_selects:
```
jklugherz, line 106: I added this because crdq queries are more like annotations table queries and will never match the rdc or raw dataset selects

v03_pipeline/lib/reference_data/config.py
```diff
@@ -372,10 +372,12 @@ def custom_mpc_select(ht):
     },
     'gnomad_qc': {
         '37': {
+            'version': 'v2',
             'custom_import': import_matrix_table,
             'source_path': 'gs://gnomad/sample_qc/mt/gnomad.joint.high_callrate_common_biallelic_snps.pruned.mt',
         },
         '38': {
+            'version': 'v3.1',
```
jklugherz, line 380: `gnomad_qc` is the only dataset without a version in either the config or the source table, so I created a version field for them here in the config so that the globals validation works for this dataset (`clinvar` and `gnomad_genomes` have versions)

bpblanken: 👍

v03_pipeline/lib/tasks/reference_data/updated_cached_reference_dataset_query.py
```diff
@@ -0,0 +1,127 @@
...
+    def requires(self) -> luigi.Task:
+        if Env.REFERENCE_DATA_AUTO_UPDATE and not self.crdq.query_raw_dataset:
+            return UpdatedReferenceDatasetCollectionTask(
+                self.reference_genome,
+                self.dataset_type,
+                self.sample_type,
+                ReferenceDatasetCollection.COMBINED,
+            )
+        if self.crdq.query_raw_dataset:
+            # NB: the config contents are not guaranteed to be hail tables/matrix tables.
```
bpblanken, line 72: Your thoroughness actually fixed the bug here right so we don't need the comment any more right?

v03_pipeline/lib/tasks/reference_data/updated_cached_reference_dataset_query.py
```diff
@@ -87,6 +88,7 @@ def create_table(self) -> hl.Table:
         if self.crdq.query_raw_dataset:
             query_ht = import_ht_from_config_path(
                 CONFIG[dataset][self.reference_genome.v02_value],
+                dataset,
```
jklugherz, line 91: since I mock `import_ht_from_config_path` in tests I didn't catch this until later :/

v03_pipeline/lib/reference_data/compare_globals.py
```diff
@@ -34,9 +34,14 @@ def from_dataset_configs(
         rdc: ReferenceDatasetCollection,
         dataset_type: DatasetType,
         reference_genome: ReferenceGenome,
+        single_dataset: str | None = None,
     ):
         paths, versions, enums, selects = {}, {}, {}, {}
-        for dataset in rdc.datasets(dataset_type):
+        for dataset in (
```
jklugherz, line 40: Each crdq uses 1 dataset. We don't need to build the Globals object with all datasets for the rdc (which will always be ReferenceDatasetCollection.COMBINED), so I added back this optional parameter

v03_pipeline/lib/reference_data/compare_globals.py
```diff
@@ -51,7 +51,7 @@ def from_dataset_configs(
             dataset_ht_globals = hl.eval(dataset_ht.globals)
             paths[dataset] = dataset_ht_globals.path
             versions[dataset] = dataset_ht_globals.version
-            enums[dataset] = dataset_ht_globals.enums
+            enums[dataset] = dict(dataset_ht_globals.enums)
```
jklugherz, line 54: This was a bug I didn't catch until now because my test cases were wrong, they were enums that were already dicts: `enums={'test_enum': ['C', 'D']}`

v03_pipeline/lib/reference_data/compare_globals.py
bpblanken: would another option here be to pass in `datasets` as a `list[str]` rather than both the `rdc` and the `single_dataset`?

v03_pipeline/lib/reference_data/compare_globals.py
```diff
@@ -68,11 +69,10 @@ def from_ht(
         dataset_type: DatasetType,
     ):
         rdc_globals_struct = hl.eval(ht.globals)
-        paths = dict(rdc_globals_struct.paths)
-        versions = dict(rdc_globals_struct.versions)
+        paths = dict(rdc_globals_struct.get('paths', {}))
```
jklugherz, line 72: added defaults here because some of the raw datasets that are used by crdqs don't necessarily have the standard paths/versions/enums globals

v03_pipeline/lib/reference_data/compare_globals_test.py
```diff
@@ -113,6 +114,37 @@ def test_create_globals_from_dataset_ht_configs(
             },
         )
 
+    @mock.patch.dict('v03_pipeline.lib.reference_data.compare_globals.CONFIG', CONFIG)
+    @mock.patch(
+        'v03_pipeline.lib.reference_data.dataset_table_operations.hl.read_table',
+    )
+    def test_create_globals_from_dataset_configs_single_dataset(self, mock_read_table):
+        mock_read_table.return_value = B_TABLE
```
jklugherz, line 122: This test case mocks `hl.read_table` instead of `import_ht_from_config_path` so we have coverage of that function being called. (I couldn't do this in the first test in this file because dataset 'a' has a custom_import,  so I had to mock the call to `import_ht_from_config_path`). This test also covers the case of using the `single_dataset` param.

v03_pipeline/lib/reference_data/compare_globals.py
jklugherz: yeah that works, I think that's what was done in the original implementation of this fn a few months back

v03_pipeline/lib/tasks/reference_data/updated_cached_reference_dataset_query.py
bpblanken: we really need type checking :/

v03_pipeline/lib/tasks/reference_data/updated_cached_reference_dataset_query.py
jklugherz: yeah, the 2 raw datasets are a hail table and a matrix table

v03_pipeline/lib/reference_data/compare_globals.py
jklugherz: I changed the param to `datasets` as a `list[str]`

v03_pipeline/lib/reference_data/compare_globals.py
```diff
@@ -34,14 +34,10 @@ def from_dataset_configs(
         rdc: ReferenceDatasetCollection,
         dataset_type: DatasetType,
         reference_genome: ReferenceGenome,
-        single_dataset: str | None = None,
+        datasets: list[str] | None = None,
```
bpblanken, line 37: sorry to push harder on this... I think we want either `datasets` or `rdc` but not both here.  `rdc`  being passed is basically just a proxy for the list of `datasets` to build.

honestly what we really want is for the `CachedReferenceDatasetQuery` object to be passed in here instead of an `rdc`.  So maybe something like:

```
    @classmethod
    def from_dataset_configs(
        cls,
         dc: DatasetCollection,
         dataset_type: DatasetType,

...
    for dataset in dc.datasets(dataset_type):
```

with both `CachedReferenceDatasetQuery` and `ReferenceDatasetCollection` implementing the `DatasetCollection` interface (a "datasets" method).

bpblanken: This is a little more of a broader change, so lemme know if you want to chat about it.

bpblanken: And it's also annoying because the `CachedReferenceDatasetQuery` has `dataset` instead of `datasets` now, but it's probably a good idea to have the behavior be shared.

v03_pipeline/lib/reference_data/compare_globals.py
bpblanken: Would it be better to just go through and fix them now?  I think we probably want to hard fail in the future if these fields are missing.

jklugherz: good idea. I'll annotate globals for the existing crdq tables

v03_pipeline/lib/reference_data/compare_globals.py
jklugherz: this makes sense and shouldn't be too big of a change!

v03_pipeline/lib/reference_data/compare_globals.py
jklugherz: I removed the {} defaults but still need to update the most recent crdq tables before this is merged

v03_pipeline/lib/reference_data/compare_globals.py
bpblanken: 👍 that seems reasonable!

v03_pipeline/lib/reference_data/dataset_table_operations.py
```diff
@@ -80,7 +80,7 @@ def import_ht_from_config_path(
         if 'custom_import' in config
         else hl.read_table(path)
     )
-    return ht.select_globals(
+    return ht.annotate_globals(
```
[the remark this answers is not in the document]
bpblanken: I think this is probably correct!  If the gnomad custom selects rely on globals then we would've been broken when we update gnomad.

bpblanken:
The one missing thing here... do we want to add this task as a requires/dependency of one of the tasks in luigi?  Or just slot the task into airflow somewhere?

bpblanken (commented):
approved!

PR #700, opened 2024-02-22
Lookup table refactor.

bpblanken:


v03_pipeline/lib/misc/family_entries.py
```diff
@@ -97,26 +97,23 @@ def remove_new_callset_family_guids(
     ht: hl.Table,
     family_guids: list[str],
 ) -> hl.Table:
+    family_guids = hl.set(family_guids)
```
jklugherz, line 100: The `family_guids` param comes from a `.collect()` call [here](https://github.com/broadinstitute/seqr-loading-pipelines/blob/19875488b36caae6d8ff019fdabd9fb0ef2389eb/v03_pipeline/lib/tasks/update_project_table.py#L109) - would it improve performance to keep it in hail instead of turning the guids into a python array then to a hail set?

bpblanken: Yep good catch! Keeping it hail makes sense I think.  I refactored one of the internal methods to be hail instead of vanilla python so should preserve the hail everywhere.

bpblanken: I just found [key_set](https://hail.is/docs/0.2/hail.expr.DictExpression.html#hail.expr.DictExpression.key_set) which is pretty much exactly what we want.

v03_pipeline/lib/misc/lookup.py
```diff
@@ -0,0 +1,235 @@
...
+    family_guids: list[str],
+) -> hl.Table:
+    if project_guid not in hl.eval(ht.globals.project_families):
+        return ht
+    family_guids = hl.set(family_guids)
+    project_i = ht.project_guids.index(project_guid)
+    family_indexes_to_keep = hl.array(
+        hl.enumerate(ht.globals.project_families[project_guid])
+        .filter(lambda item: ~family_guids.contains(item[1]))
+        .map(lambda item: item[0]),
```
jklugherz, line 74: this logic is shared with 'family_entries.py', maybe it belongs in a helper function

jklugherz: also why do we need to do this? why are there extra families inside of a 1project_families` project?

bpblanken: The field name is `family_guids` there rather than `ht.globals.project_families[project_guid]`, though you're right that the rest of the logic is the same.  There's an additional nested layer here because the lookup table contains all projects rather than just one.  

The idea here is that we want to remove families from the table if they've been re-called and are present in the new callset.

hanars: its out of scope for this PR but at some point in the future we will want to add the ability to just remove families from seqr without reloading them and we will probably use this helper function. Since theres nothing in this function thats specifically related to the new callset, might be more maintainable to just name it `remove_family_guids` or something like that

v03_pipeline/lib/misc/lookup.py
```diff
@@ -0,0 +1,234 @@
...
+                ),
+            ],
+        ),
+    ).rows()
+    return globalize_ids(ht, project_guid)
+
+
+def globalize_ids(ht: hl.Table, project_guid: str) -> hl.Table:
+    row = ht.take(1)
+    has_project_stats = row and len(row[0].project_stats) > 0
```
hanars, line 46: very minor, but its confusing that `row` is an array of length one. I would either change the name to `rows` or change the definition to `row = ht.take(1)[0]`

v03_pipeline/lib/misc/lookup.py
```diff
@@ -0,0 +1,234 @@
...
+        hl.enumerate(ht.globals.project_families[project_guid])
+        .filter(lambda item: ~family_guids.contains(item[1]))
+        .map(lambda item: item[0]),
+    )
+    ht = ht.annotate(
+        project_stats=(
+            hl.enumerate(ht.project_stats).starmap(
+                lambda i, fs: (
+                    hl.if_else(
+                        i != project_i,
```
hanars, line 80: probably out of scope, but since we are often running the pipeline on multiple projects at once it feels kind of inefficient that we need to run this operation once per project, it feels like it would be more efficient to run it for all the projects at once

v03_pipeline/lib/misc/lookup.py
```diff
@@ -0,0 +1,234 @@
...
+                            lambda j: ht.project_stats[i][j],
+                        ),
+                    )
+                ),
+            )
+        ),
+    )
+    return ht.annotate_globals(
+        project_families=hl.dict(
+            ht.project_families.items().map(
```
hanars, line 92: nitpick but would be more consistent to use `starmap`

v03_pipeline/lib/misc/lookup.py
```diff
@@ -0,0 +1,234 @@
...
+        # 2) The project has not been loaded before, the row is present
+        #    on the left but missing on the right.
+        # 3) The project as not been loaded before, the row is present on
+        #    both the left and right.
+        # 4) The project has been loaded before, the row is missing on the
+        #    left but present on the right.
+        # 5) The project has been loaded before, the row is present on the
+        #    left but missing on the right.
+        # 6) The project has been loaded before, the row is present on both
+        #    the left and right.
```
hanars, line 130: So its not 100% the same but I had to deal with a similar-ish scenario for merging different project hts while merging project tables in seqr and ended up with the following code to join `project_ht` onto `families_ht`. Its much simpler than what you have here, and I think it covers all the same cases you have but its possible theres an edge case that I don't need to handle the way you do
```
entry_type = families_ht.family_entries.dtype.element_type
default=hl.range(num_families_ht_families).map(lambda i: hl.missing(entry_type))
default_1=hl.range(num_project_ht_families).map(lambda i: hl.missing(entry_type))
families_ht = families_ht.join(project_ht, how='outer')
families_ht = families_ht.select_globals(
     family_guids=families_ht.family_guids.extend(families_ht.family_guids_1)
)
families_ht = families_ht.select(
    family_entries=hl.bind(
       lambda a1, a2: a1.extend(a2),
       hl.or_else(families_ht.family_entries, default),
       hl.or_else(families_ht.family_entries_1, default_1),
    ),
)
```

v03_pipeline/lib/misc/lookup.py
bpblanken: `ht.take(1)[0]` will `IndexError` in the empty table case.  `row =  ht.take(1)[0] if ht.count() else None` is maybe clearer though?

v03_pipeline/lib/misc/lookup.py
bpblanken: 👍 should be a straightforward change I think.  Would save I/O at least!

v03_pipeline/lib/misc/lookup.py
bpblanken: yes, that's pretty much what I've got [here](https://github.com/broadinstitute/seqr-loading-pipelines/blob/main/v03_pipeline/lib/misc/family_entries.py#L126-L147) in the entries table.  

The difference is coming from the nesting of the families within the projects here.  With the entries tables, though it's also an array of arrays, the inner array (family) is always complete.  So merging the old families with the new families is either the "extend" or the population with the array of missing families.

Here, we're handling projects being added wholesale to the end of the existing array AND families being added in at the index of the project.  The top three cases in this `case` statement are more or less what we have in the family entries table, and the bottom three cases are for the second.  It was tricky to split into two separate blocks of three cases each (though I tried!).  Maybe worth trying again?

hanars: no I think this is okay as is, just wanted to make sure we weren't overlooking an obvious improvement

PR #701, opened 2024-02-26
Delete old pipeline runs.

bpblanken:
This will also need an airflow change to run this task, which is not too much effort.


jklugherz (approved):
lgtm. will the airflow task run on a schedule?

bpblanken:
@jklugherz I was just gonna slot this into airflow to run at the start of the pipeline!  It should be relatively cheap!

PR #702, opened 2024-02-26
Add reference genome check to VEP.  

bpblanken:


hanars (approved):

v03_pipeline/lib/vep.py
```diff
@@ -1,20 +1,29 @@
...
+    reference_genome: ReferenceGenome,
     vep_config_json_path: str | None,
 ) -> hl.Table:
     if not dataset_type.veppable:
         return ht
     config = (
         vep_config_json_path
         if vep_config_json_path is not None
-        else 'file:///vep_data/vep-gcloud.json'
+        else f'file:///vep_data/vep-{reference_genome.value}-gcloud.json'
```
jklugherz, line 24: where does the vep config json live?

jklugherz (approved):

v03_pipeline/lib/vep.py
bpblanken: It's basically created for us by hail's initialization of the VEP environment on dataproc.  Our local installation provides a local VEP config set [here](https://github.com/broadinstitute/seqr-loading-pipelines/tree/main/docker/vep_configs), which actually already has the reference genome embedded in the path.

PR #704, opened 2024-02-27
Add 'reasons' key to `metadata.json` with sex/relatedness check failures.

bpblanken:
Adds a `reasons` key to the failures list of a runs `metadata.json`.


v03_pipeline/lib/misc/family_loading_failures.py
```diff
@@ -10,83 +12,92 @@ def passes_relatedness_check(
...
+            f'Sample {sample_id} has expected relation "{relation.value}" to {other_id} but has coefficients {coefficients or []}',
+        )
+    return True, None
 
 
 def passes_all_relatedness_checks(  # noqa: C901
     relatedness_check_lookup: dict[tuple[str, str], list],
     sample: Sample,
-) -> bool:
+) -> tuple[bool, str | None]:
```
bpblanken, line 37: aw, I just noticed that this is still written such that only the first failure will be returned, even though we have the ability to return all relationship failures.  I will quickly fix that!

v03_pipeline/lib/misc/family_loading_failures.py
```diff
@@ -99,7 +111,9 @@ def build_relatedness_check_lookup(
         j=remap_lookup.get(relatedness_check_ht.j, relatedness_check_ht.j),
     )
     return {
-        (r.i, r.j): list(r.drop('i', 'j').values())
+        # NB: samples are sorted in the original ibd but not necessarily
```
jklugherz, line 114: what's ibd?

bpblanken: It's the relatedness matrix produced by [`hl.identity_by_descent`](https://hail.is/docs/0.2/methods/relatedness.html#hail.methods.identity_by_descent).  There's some examples of what it looks like there or in our tests.

v03_pipeline/lib/tasks/write_remapped_and_subsetted_callset.py
```diff
@@ -127,9 +127,9 @@ def create_table(self) -> hl.MatrixTable:
 
         loadable_families = (
             families
-            - families_failed_missing_samples
-            - families_failed_relatedness_check
-            - families_failed_sex_check
+            - families_failed_missing_samples.keys()
+            - families_failed_relatedness_check.keys()
+            - families_failed_sex_check.keys()
```
jklugherz, line 132: does it happen often that all families fail the checks? is it worth changing the error message on line 135 to include info about which families failed which checks?

jklugherz (approved):
I like that you added a comment with the structure of the test tables!

v03_pipeline/lib/misc/family_loading_failures.py
```diff
@@ -10,83 +12,93 @@ def passes_relatedness_check(
...
             relatedness_check_lookup,
             sample.sample_id,
             half_sibling_id,
-            Relation.SIBLING,
-        ):
-            return False
+            Relation.HALF_SIBLING,
+        )
+        if not success1 and not success2:
+            failure_reasons.append(reason)
```
matren395, line 90: a bit turned around my logic here - fail if they pass tests for both sibling and half-sibling, yeah? 

Would there ever be a case where they would fail at being half-siblings but pass at being full siblings?

matren395 (commented):
makes sense enough to be , but one quick question about half sibling logic

v03_pipeline/lib/misc/family_loading_failures.py
bpblanken: yeah, this is specific to the "half-siblings" parsed from the pedigree (meaning a single parent in common).  the logic here means that we won't fail the family if a half-sibling in the pedigree is actually a full sibling in the ibd.  the relationship is a "success" if either of the checks passes.  

we check the full siblings [here](https://github.com/broadinstitute/seqr-loading-pipelines/pull/704/files/0812e75ecbedd78dbed33b7e23760cb1dceaa4a0#diff-81bda30e9edc96dbaf6b9533d4db17f8afa17df35de4b37c3c05e1856e6779dbR64), where both parents are shared.

PR #707, opened 2024-02-28
fix test tables for new clinvar assertions

jklugherz:
fix test cases for my accidental dev [commit](https://github.com/broadinstitute/seqr-loading-pipelines/commit/a493221cb42ff5b338b973b7768a116f6e6f9326) which adds new assertions and gold star lookup values


bpblanken:
@hanars so you see this.  We encountered changes to the "assertions" modifiers that were parsing away from the pathogenicities.  If we continue to see frequent naming changes to these we should probably change the behavior to not hard-fail and instead dynamically map the enums.  We're not doing anything on the search end other than returning in the JSON response and displaying in the UI right?

hanars:
yeah, seqr has nothing hard coded around assertions so theres no harm if these dynamically update

v03_pipeline/lib/tasks/reference_data/update_variant_annotations_table_with_updated_reference_dataset_test.py
```diff
@@ -388,7 +388,7 @@ def test_update_vat_with_updated_rdc_mito_38(
                     clinvar_mito=None,
                     dbnsfp_mito=hl.Struct(
                         SIFT_score=None,
-                        MutationTaster_pred='N',
```
jklugherz, line 391: fixed this column in the test table too

bpblanken (approved):

bpblanken:
merging so I can pull into my other branch

PR #713, opened 2024-03-01
add new clinvar pathogenicity and contig filter workaround

jklugherz:
bug 1: 
Hail 0.2.128 crashes when we filter the imported clinvar table with `ht.filter(ht.locus.contig != 'MT')`. If we just don't include that contig when we import the vcf, the MT loci will be ignored and we don't need the filter.

bug 2: 
```
HailUserError: Error summary: HailException: Key "low_penetrance/Established_risk_allele" not found in dictionary. Keys: ["Affects","association","association_not_found","confers_sensitivity","drug_response","low_penetrance","no_classification_for_the_single_variant","no_classifications_from_unflagged_records","not_provided","other","protective","risk_factor",null]
```
The CNLSIG that caused the error: [
                            'Pathogenic/Likely_pathogenic/Pathogenic',
                            '_low_penetrance/Established_risk_allele',
                        ]

We added a new .replace() line in parsed_clnsig() which adds a new pathogenicity       'Pathogenic/Likely_pathogenic/Established_risk_allele' above 'Pathogenic/Likely_pathogenic/Likely_risk_allele'. 

This requires no seqr changes because the new pathogenicity is within the `pathogenic` range.


v03_pipeline/lib/annotations/enums.py
```diff
@@ -198,6 +198,7 @@
 CLINVAR_PATHOGENICITIES = [
     'Pathogenic',
     'Pathogenic/Likely_pathogenic',
+    'Pathogenic/Likely_pathogenic/Established_risk_allele',
```
jklugherz, line 201: @hanars wanted to bring this new pathogenicity to your attention even though it is within seqr's pathogenic filtering range: https://github.com/broadinstitute/seqr/blob/master/hail_search/constants.py#L75

hanars: yes should nto be an issue for seqr but appreciate the heads up!

v03_pipeline/lib/reference_data/clinvar.py
```diff
@@ -120,14 +125,25 @@ def download_and_import_latest_clinvar_vcf(
             reference_genome=reference_genome.value,
             drop_samples=True,
             skip_invalid_loci=True,
-            contig_recoding=reference_genome.contig_recoding(include_mt=True),
+            contig_recoding=reference_genome.contig_recoding(include_mt),
```
bpblanken, line 128: 🤔 I think there's a subtly here that makes this maybe not work the way we think.  The contig recoding is saying when I see `MT`, turn that into `chrM`.  So for `GRCh38`, if we don't do that, `MT` will stay `MT` and we'll drop those rows when we [filter for the standard contigs](https://github.com/broadinstitute/seqr-loading-pipelines/blob/a1564d7d4c1ac03a28129e60bd84e9603b3ced5c/v03_pipeline/lib/reference_data/dataset_table_operations.py#L59).  However, for `GRCh37`, the `MT` is already an `MT` and we don't use the recode, so we'll keep that row because `MT` is actually a standard contig.  

Looks like we can just do `~(ht.locus.contig == 'MT')` or `hl.filter_intervals(ht, [hl.parse_locus_interval('MT')]` here.  

Let me know if that makes sense.

jklugherz: Ah I just read in the table, filtered for 'MT' and yep I see them. I'll use `~(ht.locus.contig == 'MT')` instead.

bpblanken (approved):

v03_pipeline/lib/annotations/enums.py
hanars: 😬

PR #715, opened 2024-03-04
add a few more logs

jklugherz:


v03_pipeline/lib/logger.py
```diff
@@ -20,7 +20,7 @@
         '': {
             'level': 'INFO',
             'handlers': ['default'],
-            'propagate': True,
+            'propagate': False,
```
jklugherz, line 23: I was hoping this would eliminate the log duplicates but not sure

bpblanken (approved):

v03_pipeline/lib/logger.py
bpblanken: yeah, `propagate` is set to true here for something that we don't have yet (testing logs in the unit tests).  so fine to disable for now.

PR #718, opened 2024-03-07
use dataset configs in UpdateVATWithUpdatedRDC complete()

jklugherz:
Before this change, when complete() runs, there are no datasets to update since the rdc tables haven't been updated by the time this task runs. (requires() won't run because complete() is already true)

By looking at the dataset configs instead of rdc globals, we can actually run requires()

### DIFF
--- a/v03_pipeline/lib/reference_data/compare_globals.py
+++ b/v03_pipeline/lib/reference_data/compare_globals.py
@@ -22,8 +22,8 @@ logger = get_logger(__name__)
 
 @dataclasses.dataclass
 class Globals:
-    paths: dict[str]
-    versions: dict[str]
+    paths: dict[str, str]
+    versions: dict[str, str]
     enums: dict[str, dict[str, list[str]]]
     selects: dict[str, set[str]]
 
@@ -33,12 +33,11 @@ class Globals:
     @classmethod
     def from_dataset_configs(
         cls,
-        rdc: ReferenceDatasetCollection,
-        dataset_type: DatasetType,
         reference_genome: ReferenceGenome,
+        datasets: list[str],
     ):
         paths, versions, enums, selects = {}, {}, {}, {}
-        for dataset in rdc.datasets(dataset_type):
+        for dataset in datasets:
             dataset_config = CONFIG[dataset][reference_genome.v02_value]
             dataset_ht = import_ht_from_config_path(dataset_config, reference_genome)
 
--- a/v03_pipeline/lib/reference_data/compare_globals_test.py
+++ b/v03_pipeline/lib/reference_data/compare_globals_test.py
@@ -13,46 +13,61 @@ from v03_pipeline.lib.reference_data.compare_globals import (
     get_datasets_to_update,
 )
 
+CONFIG = {
+    'a': {
+        '38': {
+            'custom_import': None,
+            'source_path': 'a_path',  # 'a' has a custom import
+            'select': {
+                'test_select': 'info.test_select',
+                'test_enum': 'test_enum',
+            },
+            'version': 'a_version',
+            'enum_select': {'test_enum': ['A', 'B']},
+        },
+    },
+    'b': {  # b is missing version
+        '38': {
+            'path': 'b_path',
+            'select': {
+                'test_select': 'info.test_select',
+                'test_enum': 'test_enum',
+            },
+            'enum_select': {'test_enum': ['C', 'D']},
+            'custom_select': lambda ht: {'field2': ht.info.test_select_2},
+        },
+    },
+}
+
+B_TABLE = hl.Table.parallelize(
+    [],
+    schema=hl.tstruct(
+        locus=hl.tlocus('GRCh38'),
+        alleles=hl.tarray(hl.tstr),
+        info=hl.tstruct(
+            test_select=hl.tint,
+            test_select_2=hl.tint,
+        ),
+        test_enum=hl.tstr,
+    ),
+    globals=hl.Struct(
+        version='b_version',
+        path='b_path',
+        enums=hl.Struct(test_enum=['C', 'D']),
+    ),
+    key=['locus', 'alleles'],
+)
+
 
 class CompareGlobalsTest(unittest.TestCase):
-    @mock.patch.dict(
-        'v03_pipeline.lib.reference_data.compare_globals.CONFIG',
-        {
-            'a': {
-                '38': {
-                    'custom_import': None,
-                    'source_path': 'a_path',  # 'a' has a custom import
-                    'select': {
-                        'test_select': 'info.test_select',
-                        'test_enum': 'test_enum',
-                    },
-                    'version': 'a_version',
-                    'enum_select': {'test_enum': ['A', 'B']},
-                },
-            },
-            'b': {  # b is missing version
-                '38': {
-                    'path': 'b_path',
-                    'select': {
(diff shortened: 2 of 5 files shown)


v03_pipeline/lib/reference_data/compare_globals.py
```diff
@@ -33,12 +33,11 @@ def __getitem__(self, name: str):
     @classmethod
     def from_dataset_configs(
         cls,
-        rdc: ReferenceDatasetCollection,
-        dataset_type: DatasetType,
         reference_genome: ReferenceGenome,
+        datasets: list[str],
```
jklugherz, line 37: I made this change in crdq pr and added to this one so that I don't make merge conflicts for myself

v03_pipeline/lib/reference_data/compare_globals_test.py
```diff
@@ -105,6 +110,37 @@ def test_create_globals_from_dataset_ht_configs(
...
+            datasets=['b'],
+        )
+
+        self.assertTrue(
+            dataset_config_globals.versions == {'b': 'b_version'},
+        )
+        self.assertTrue(
+            dataset_config_globals.paths == {'b': 'b_path'},
+        )
+        print(dataset_config_globals.enums)
```
bpblanken, line 133: stray print!

v03_pipeline/lib/tasks/reference_data/update_variant_annotations_table_with_updated_reference_dataset_test.py
```diff
@@ -130,6 +135,136 @@ def test_update_vat_with_updated_rdc_snv_indel_38(
                 updates=hl.empty_set(hl.tstruct(callset=hl.tstr, project_guid=hl.tstr)),
             ),
         )
+
+        mock_globals_from_config = Globals(
```
bpblanken, line 139: is it possible to reduce the "real-ness" of this mock at all?  It's basically just the real config.

jklugherz: yeah I need to think a bit more about this, 
I want to mock the `Globals` created from `from_dataset_configs()` because 
- the data in the tables may (and does rn) diverge from what's in the config and we don't want to have to update our test tables if we update something in the configs?
- `from_dataset_configs()` also reads in the dataset raw table which is why I lean towards just mocking the return value of that fn

bpblanken: yeah, is one option copying the config as is and changing the one two things that the tests require to be different?

v03_pipeline/lib/reference_data/compare_globals.py
bpblanken: Should we also change `from_ht` below to use a `datasets` list rather than the `rdc`?  Just to share an "interface"?

jklugherz: yes, I like that

v03_pipeline/lib/tasks/reference_data/update_variant_annotations_table_with_updated_reference_dataset_test.py
jklugherz: ok I just mocked the `get_datasets_to_update()` call to return an empty list. these tests are really testing the `update_table()` functionality

v03_pipeline/lib/reference_data/compare_globals_test.py
jklugherz: oop good catch

v03_pipeline/lib/tasks/update_variant_annotations_table_with_new_samples_test.py
```diff
@@ -1436,8 +1465,14 @@ def test_sv_update_vat(
...
 
-    def test_gcnv_update_vat(self, mock_update_rdc_task: None) -> None:
+    def test_gcnv_update_vat(
+        self,
+        mock_update_rdc_task,
+        mock_vat_get_datasets_to_update,
+        mock_vat_globals_from_dataset_configs,
+    ) -> None:
         mock_update_rdc_task.return_value = MockCompleteTask()
+        mock_vat_get_datasets_to_update.return_value = []
```
bpblanken, line 1475: I'm a little concerned by the need to mock the other dataset types here.  There's no reference data, so the `datasets` array should be empty already, making those tasks complete by default.  Is there another approach here?

v03_pipeline/lib/tasks/reference_data/update_variant_annotations_table_with_updated_reference_dataset_test.py
```diff
@@ -130,6 +138,8 @@ def test_update_vat_with_updated_rdc_snv_indel_38(
                 updates=hl.empty_set(hl.tstruct(callset=hl.tstr, project_guid=hl.tstr)),
             ),
         )
+        mock_get_datasets_to_update.return_value = []
```
bpblanken, line 141: I'm a little confused by this mock too.  We set `datasets_to_update` to be the empty list, but the rest of the test includes updating [all the datasets](https://github.com/broadinstitute/seqr-loading-pipelines/pull/718/files#diff-d7e252ad0022625cd78f557ba25857241d79b35a8a6d9a99b32daa1672b7385cR165-R182)

v03_pipeline/lib/tasks/update_variant_annotations_table_with_new_samples_test.py
```diff
@@ -74,6 +74,12 @@
 }
 
 
+@patch(
```
bpblanken, line 77: I guess another solution here... can we just mock `UpdateVariantAnnotationsTableWithUpdatedReferenceDataset` into the `MockCompleteTask`?  that might be a clean option?

jklugherz: oh yes that's cleaner

v03_pipeline/lib/tasks/reference_data/update_variant_annotations_table_with_updated_reference_dataset_test.py
jklugherz: These tests begin with a nonexistent annotations table, so  on the first `complete()` call, `initialize_table()` is called then `update_table()` adds all the datasets to the vat.  

During the next `complete()` call, if this were real life, `get_datasets_to_update()` should return an empty list

jklugherz: maybe I need a test case which also updates an existing vat

v03_pipeline/lib/reference_data/compare_globals.py
jklugherz: that makes sense, I still want to pass `datasets` to `get_datasets_to_update` to be explicit because the code doesn't enforce that `ht1_globals` and `ht2_globals` have the same datasets

bpblanken: Isn't that a good thing though?

- In production world we want a discrepancy to go noticed.
- In tests, we can force the behavior by mocking the config correctly rather than mocking the `rdc.datasets`, which I think actually simplifies things.

PR #721, opened 2024-03-10
use datasets in Globals constructors

jklugherz:


v03_pipeline/lib/reference_data/compare_globals.py
```diff
@@ -64,17 +61,21 @@ def from_dataset_configs(
...
-        dataset_type: DatasetType,
+        datasets: list[str],
     ):
         rdc_globals_struct = hl.eval(ht.globals)
         paths = dict(rdc_globals_struct.paths)
         versions = dict(rdc_globals_struct.versions)
         # enums are nested structs
         enums = {k: dict(v) for k, v in rdc_globals_struct.enums.items()}
 
+        for global_dict in [paths, versions, enums]:
```
jklugherz, line 72: I added this because in `UpdateVariantAnnotationsTableWithUpdatedReferenceDataset` as it's written now, we compare the annotations ht to the rdc hts one by one, and `get_datasets_to_update` now explicitly finds differing datasets between the two `Globals` objects, so I want to filter out datasets not in the rdc I'm checking when I call `from_ht` on the annotations ht. 

I think that once I change the call to `from_ht` on the rdc hts in that task to `from_dataset_configs`, I can use all the datasets for all rdcs in the `datasets` param and remove this block of code. And without this block of code we can also catch datasets that need to be dropped.

I know I split out that work but I sorta want to add it back in to this PR (and close the other one) so I don't check in code that I'll immediately change?

v03_pipeline/lib/reference_data/compare_globals.py
```diff
@@ -86,30 +87,21 @@ def from_ht(
...
-        result = ht1_globals[field.name].get(dataset) == ht2_globals[field.name].get(
-            dataset,
-        )
-        if result is False:
-            logger.info(f'{field.name} mismatch for {dataset}, {rdc.value}')
-        results.append(result)
-    return all(results)
+        ht1_datasets = set(ht1_globals[field.name].keys())
+        ht2_datasets = set(ht2_globals[field.name].keys())
+        datasets_to_update.update(ht1_datasets.symmetric_difference(ht2_datasets))
```
bpblanken, line 98: nit, but maybe the set ops are cleaner?  `^` for symmetric difference and `&` for intersection?

v03_pipeline/lib/reference_data/compare_globals.py
bpblanken: yeah this makes sense.  I think this is good to merge though, so we can just proceed there?

bpblanken (approved):

PR #723, opened 2024-03-11
use dataset configs in UpdateVATWithUpdatedRDC complete() take 2

jklugherz:


v03_pipeline/lib/reference_data/compare_globals.py
```diff
@@ -96,6 +91,16 @@ def get_datasets_to_update(
         datasets_to_update.update(
             ht1_globals[field.name].keys() ^ ht2_globals[field.name].keys(),
         )
+        if field.name == 'enums':
```
bpblanken, line 94: good catch on these weird enum fields.  I think it might be better to just exclude these when we construct the globals [here](https://github.com/broadinstitute/seqr-loading-pipelines/pull/723/files#diff-2460cc720d61d0c78e554d81ddfcd480509c4b260270709c2952a6bcb091d729R70)?  they're basically all non-reference-data related so can be considered separate from this process we're doing to compare reference data globals.

v03_pipeline/lib/tasks/update_variant_annotations_table_with_new_samples_test.py
```diff
@@ -1436,8 +1460,13 @@ def test_sv_update_vat(
             ],
         )
 
-    def test_gcnv_update_vat(self, mock_update_rdc_task: None) -> None:
+    def test_gcnv_update_vat(
+        self,
+        mock_update_rdc_task,
+        mock_update_vat_with_rdc_task,
```
bpblanken, line 1466: If we lift the mocking down from the class onto the individual functions we can exclude the mocks from SV and GCNV right?

v03_pipeline/lib/tasks/reference_data/update_variant_annotations_table_with_updated_reference_dataset_test.py
```diff
@@ -34,6 +35,560 @@
...
+                key=['locus', 'alleles'],
+                globals=hl.Struct(),
+            ),
+        },
+    },
+    'screen': {
+        '38': {
+            **CONFIG['screen']['38'],
+            'source_path': 'gs://seqr-reference-data/GRCh38/ccREs/GRCh38-ccREs.ht',
+            'custom_import': lambda *_: hl.Table.parallelize(
```
bpblanken, line 430: Are the `custom_import`'s here being used at all?  It feels like the mocking here is a little heavy, so it'd  be great if we could simplify in some way.

bpblanken: Ah, I'm following now.  We're actually going through and using these stubbed custom_imports as the table imports to check the versions.

v03_pipeline/lib/reference_data/compare_globals.py
bpblanken: Actually, won't `from_ht` above handle this by filtering to `datasets`?

bpblanken: (one hint here, you can add a `for dataset in datasets:` loop to `from_ht` which'll make it look similar to `from_dataset_configs` and it'll be super nice!

v03_pipeline/lib/tasks/reference_data/update_variant_annotations_table_with_updated_reference_dataset_test.py
jklugherz: Exactly, it's a clunky dict but allows us to execute as much of the actual code as possible, including most of `from_dataset_configs()`. It's better than mocking the return value of `from_dataset_configs`.

v03_pipeline/lib/tasks/update_variant_annotations_table_with_new_samples_test.py
jklugherz: good point, I was in a copy paste frenzy but these mocks definitely don't go here

v03_pipeline/lib/tasks/reference_data/update_variant_annotations_table_with_updated_reference_dataset_test.py
bpblanken: yep!  I'm on board!  this is great.

v03_pipeline/lib/reference_data/compare_globals.py
jklugherz: In `from_ht`, after I took out [these lines](https://github.com/broadinstitute/seqr-loading-pipelines/pull/723/files/2df1a3f7232386df9fafd63253b3d14b8923a214#diff-2460cc720d61d0c78e554d81ddfcd480509c4b260270709c2952a6bcb091d729L72-L76), only the `selects` are filtered to `datasets` using the `for dataset in datasets:` loop.
I prefer not to filter the datasets in `from_ht` at all because any extra dataset in these globals that isn't one of those special annotation enums is a dataset that needs to be dropped needs to be dropped.

bpblanken: oof, yah that's correct about the dropping 🤔.  What about filtering `enums` to only the datasets that are in `paths` or `versions`?  not sure if that's better.

jklugherz: I could also filter `enums` to `datasets` so we'd still catch extra datasets in `paths` and `versions`

bpblanken: 👍 I don't think there should be anything in paths/versions that's not in enums.  But that's basically just a heuristic and not enforced.

v03_pipeline/lib/reference_data/compare_globals.py
```diff
@@ -67,12 +67,9 @@ def from_ht(
...
         versions = dict(rdc_globals_struct.versions)
         # enums are nested structs
-        enums = {k: dict(v) for k, v in rdc_globals_struct.enums.items()}
-
-        for global_dict in [paths, versions, enums]:
-            for dataset in list(global_dict.keys()):
-                if dataset not in datasets:
-                    global_dict.pop(dataset)
+        enums = {
+            k: dict(v) for k, v in rdc_globals_struct.enums.items() if k in datasets
```
bpblanken, line 71: Do you think it's more accurate to filter to `datasets` or filter to `paths`?  I think they're gonna be the same in practice.  Just curious.

jklugherz: I think I like filtering to `paths` better. If there is an erroneous enum not in `datasets`, at least we'll see one of those 'mismatch' logs for it which is more helpful for debugging

bpblanken (approved):
🚢

PR #724, opened 2024-03-13
crdq wrapper task

jklugherz:


v03_pipeline/lib/tasks/reference_data/write_cached_reference_dataset_query.py
```diff
@@ -0,0 +1,40 @@
...
+    DatasetType,
+    ReferenceGenome,
+    SampleType,
+)
+from v03_pipeline.lib.tasks.reference_data.updated_cached_reference_dataset_query import (
+    UpdatedCachedReferenceDatasetQuery,
+)
+
+
+class WriteCachedReferenceDatasetQuery(luigi.Task):
```
bpblanken, line 14: yay!  I'm glad you remembered we talked about this because I sure didn't.

v03_pipeline/lib/tasks/reference_data/write_cached_reference_dataset_query.py
```diff
@@ -0,0 +1,40 @@
...
+    reference_genome = luigi.EnumParameter(enum=ReferenceGenome)
+    dataset_type = luigi.EnumParameter(enum=DatasetType)
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.dynamic_crdq_tasks = set()
+
+    def complete(self) -> bool:
+        return len(self.dynamic_crdq_tasks) >= 1 and all(
```
bpblanken, line 23: I kinda think we could just have these `return False` and force all the dynamic dependencies to run.  `complete()` isn't reevaluated after the task runs because of `check_complete_on_run` being set to false.

bpblanken:
Looks mostly correct to me!

v03_pipeline/lib/tasks/reference_data/write_cached_reference_dataset_query.py
jklugherz: ohh I hadn't internalized that `complete()` only runs once irl because we always run it a second time via `self.assertTrue(task.complete())` in tests 

I had the idea of using another instance variable `self.checked_for_tasks = False` which we can update to `True` `run()` is called and just return the variable in `complete()`

jklugherz: this also allows the task to complete or a reference genome + dataset type that has no dependencies

bpblanken: yeah: 

```
ohh I hadn't internalized that complete() only runs once irl because we always run it a second time via self.assertTrue(task.complete()) in tests
``` 

is probably weird.

bpblanken: Your approach definitely makes sense though.

bpblanken:
We want this to be a `requires` of `UpdateVariantAnnotationsTableWithNewSamplesTask` right?

jklugherz:
@bpblanken I was actually thinking about making it another airflow task in the v3 pipeline, since it's not really an upstream requirement of `UpdateVariantAnnotationsTableWithNewSamplesTask`.. what do you think?

bpblanken:
Yep! I think that makes sense.

bpblanken (approved):

PR #725, opened 2024-03-13
DRAGEN has haploid calls?  

bpblanken:


hanars:
what does it mean at like a biological level for Dragen to have a haploid call? If these calls look substantially different from GATK calls, this feels like something where we need a biologist to weigh in, not for us to just shoehorn the data into our existing schema

bpblanken:
this is direct from hail: https://discuss.hail.is/t/pl-of-haploid-call-induced-error-when-split-multi-hts-in-vcf-produced-by-dragen-pipeline/3116

bpblanken:
(this isn't going into prod, I'm just trying to get the sex/relatedness to work... likely not correctly if the X chromosomes are weird).

hanars:
are we sure this is x chromosome only, and not that we are suddenly being delivered the y chromosome? The discussion you linked to seems to assume all these calls are on the y chromosome

bpblanken:
not just X chromosome.  there's haploid calls on both X/Y.

hanars (approved):
okay sounds like we need this, but add a comment explaining hwy and maybe add a linkl to the ticket the hail team has to track adding support for this: https://github.com/hail-is/hail/issues/13149

PR #730, opened 2024-03-14
resolve parent being parsed as aunt/nephews bug

bpblanken:


v03_pipeline/lib/misc/pedigree.py
```diff
@@ -107,56 +124,34 @@ def parse_collateral_lineage(
...
-            ) or (
-                samples[sample_i].father
-                and samples[sample_i].father == samples[sample_j].father
+            if (sample_i.mother and sample_i.mother == sample_j.mother) or (
+                sample_i.father and sample_i.father == sample_j.father
             ):
-                samples[sample_i].half_siblings.append(
-                    sample_j,
-                )
+                sample_i.half_siblings.append(sample_j.sample_id)
```
matren395, line 146: remind me, I don't work with 'continue' much - full siblings aren't classified as half-siblings as well, right ?

v03_pipeline/lib/misc/pedigree.py
```diff
@@ -107,56 +124,34 @@ def parse_collateral_lineage(
...
-                samples[sample_i].mother,
-                samples[sample_i].father,
-                samples[sample_i].maternal_grandmother,
-                samples[sample_i].maternal_grandfather,
-                samples[sample_i].paternal_grandmother,
-                samples[sample_i].paternal_grandfather,
-            }:
+        for sample_i, sample_j in itertools.combinations(samples.values(), 2):
+            # If sample is already related from direct relationships, continue
+            if sample_i.is_in_direct_lineage(sample_j):
```
matren395, line 129: quick summary - what's the logic of what it was doing before versus after?

v03_pipeline/lib/misc/pedigree_test.py
```diff
@@ -205,6 +205,107 @@ def test_parse_lineage(self) -> None:
...
+                    sample_id='sample_5',
+                    sex=Sex.FEMALE,
+                    mother='sample_3',
+                    father=None,
+                    maternal_grandmother='sample_1',
+                    maternal_grandfather='sample_2',
+                    paternal_grandmother=None,
+                    paternal_grandfather=None,
+                    siblings=[],
+                    half_siblings=[],
```
matren395, line 303: also half-sibling of sample_4, or is this supposed to throw an error ?

matren395 (commented):
review! a few content questions and then one thing I'm not clear if is right or no

v03_pipeline/lib/misc/pedigree_test.py
bpblanken: There's a [comment here](https://github.com/broadinstitute/seqr-loading-pipelines/pull/730/files#diff-b905a71b20c6c80c544cc674af65cc044d5ff12fe2b2b2801ae10e4350bbfbf0R123-R126) about this, but it's correct as is I think.  It was a design choice to only list each relationship a single time when doing these checks to avoid having everything be present twice (i.e. parent and child, sibling and sibling).

v03_pipeline/lib/misc/pedigree.py
bpblanken: yep!  `continue` means to skip all subsequent steps... it's basically the same as an `else` (which might be clearer... `continue` is just a style choice).

v03_pipeline/lib/misc/pedigree.py
bpblanken: the bug was in how we were checking for sample being in a direct lineage.  We were only checking a single direction (sample_i is mother/father/grandparent of sample_j) when we need to check both.  

The below logic which checks for "aunt-uncle" makes the assumption that samples aren't direct related, so that check needs to actually be correct.

matren395:
Hmm fair for all! lgtm then

matren395 (approved):
thanks for the explanations!

matren395 (approved):
okay idk why it didn't approve before

PR #736, opened 2024-03-19
Lookup table touchups

bpblanken:


v03_pipeline/lib/misc/lookup.py
```diff
@@ -38,7 +38,18 @@ def compute_callset_lookup_ht(
             ],
         ),
     ).rows()
-    return globalize_ids(ht, project_guid)
+    ht = globalize_ids(ht, project_guid)
```
bpblanken, line 41: I included this as part of the migration... and it works!

PR #739, opened 2024-03-20
clinvar submitters and conditions

jklugherz:
fetches clinvar submission summary via ftp and uses the `Submitter` and `ReportedPhenotypeInfo` columns, joining on `rsid` from the clinvar vcf to get submitters and conditions for each VariationID


hanars (commented):
looks good to me but would want Ben's eyes on this as well

v03_pipeline/lib/reference_data/clinvar.py
```diff
@@ -59,7 +61,7 @@ def safely_move_to_gcs(tmp_file_name, gcs_tmp_file_name):
 
 def parsed_clnsig(ht: hl.Table):
     return (
-        hl.delimit(ht.info.CLNSIG)
+        hl.delimit(ht.CLNSIG)
```
bpblanken, line 64: These should maybe stay the same?

v03_pipeline/lib/reference_data/clinvar.py
```diff
@@ -150,3 +152,37 @@ def _parse_clinvar_release_date(local_vcf_path: str) -> str:
...
+            min_partitions=3,  # recommended 2-4 partitions per core
+        )
+
+
+def download_and_import_clinvar_submission_summary() -> hl.Table:
+    # https://ftp.ncbi.nlm.nih.gov/pub/clinvar/tab_delimited/README - submission_summary.txt
+    logger.info('Getting clinvar submission summary')
+    ht = download_and_import_clinvar_txt_file(
+        url=CLINVAR_SUBMISSION_SUMMARY_URL,
+        types={'#VariationID': hl.tstr},
```
bpblanken, line 179: We can just slot these on `hl.import_table` right?  Also `hl.tstr` is the default so the type isn't necessary.

v03_pipeline/lib/reference_data/clinvar.py
```diff
@@ -150,3 +152,37 @@ def _parse_clinvar_release_date(local_vcf_path: str) -> str:
...
+        hl_filter='^(#[^:]*:|^##).*$',  # removes all comments except for the header line
```
bpblanken, line 180: Would we be able to use the `comment` arg instead of a regex?

v03_pipeline/lib/reference_data/config.py
```diff
@@ -52,10 +54,21 @@ def clinvar_custom_select(ht):
...
     return selects
 
 
+def join_clinvar_to_submission_summary(ht, selects: dict):
+    submission_ht = download_and_import_clinvar_submission_summary()
+    ht = ht.info.annotate(
+        submitters=submission_ht[ht.rsid].Submitters,
+        conditions=submission_ht[ht.rsid].Conditions,
+    )
+    selects['submitters'] = ht.submitters
```
bpblanken, line 67: it's maybe _slightly_ easier to understand this if the `selects` stay in `clinvar_custom_select`!

v03_pipeline/lib/reference_data/clinvar.py
jklugherz: 1. I split the download into two functions because I was originally downloading 2 files (the variant_summary.txt.gz file as well, but we can actually get all the info from submission_summary). I want to keep these two functions separate because in tests I can mock the function that calls `import_table` but still execute the hail code that does the selecting and aggregating of submitters and conditions.

2. iirc (thank you for this acronym) the `impute=True` in `import_table` would impute the `#VariationID` as an integer, but I coerce it to a string to join on the table from the clinvar vcf

v03_pipeline/lib/reference_data/clinvar.py
jklugherz: I tried a bunch of combinations of different `comment` and `filter` and even `find_replace` to keep all of the comments except for the one with the column names (also starts with # which made it difficult). I even checked out the source code for `import_table` to see what order the params were being used.

This filter regex was the best solution I found to filter out the comments but keep the column names. I guess it's not resilient to any change in the formatting of the file comments though 

here's what the header looks like:
```
##Overview of interpretation, phenotypes, observations, and methods reported in each current submission 
##Explanation of the columns in this report
#VariationID:                   the identifier assigned by ClinVar and used to build the URL, namely https://ncbi.nlm.nih.gov/clinvar/VariationID
#ClinicalSignificance:          the germline classification on this submitted record
#DateLastEvaluated:             the last date the classification on this record was evaluated by the submitter
#Description:                   an optional free text description comment describing the rationale for the classification
#SubmittedPhenotypeInfo:        the name(s) or identifier(s) submitted as the condition for the classification 
#ReportedPhenotypeInfo:         the MedGen identifier/name combinations that the submitted condition for the classification maps to. 'na' means there is no public identifer in MedGen for the condition.
#ReviewStatus:                  the level of review for this submitted record; see http//www.ncbi.nlm.nih.gov/clinvar/docs/variation_report/#review_status
#CollectionMethod:              the method by which the submitter collected the data for the classification; see https://www.ncbi.nlm.nih.gov/clinvar/docs/spreadsheet/#collection
#OriginCounts:                  the allele origin reported by the submitter and the number of observations for each origin. ‘na’ means that the number of observations was not provided by the submitter.
#Submitter:                     the submitter of this record
#SCV:                            the accession and current version assigned by ClinVar to this submitted record
#SubmittedGeneSymbol:           the gene symbol reported in this submitted record
#ExplanationOfInterpretation:   more details if the germline classification (ClinicalSignificance) is 'other' or 'drug response'
#SomaticClinicalImpact:         the somatic classification of clinical impact on this submitted record
#Oncogenicity:                  the somatic classification of oncogenicity on this submitted record
#VariationID	ClinicalSignificance	DateLastEvaluated	Description	SubmittedPhenotypeInfo	ReportedPhenotypeInfo	ReviewStatus	CollectionMethod	OriginCounts	Submitter	SCV	SubmittedGeneSymbol	ExplanationOfInterpretation	SomaticClinicalImpact	Oncogenicity
2	Pathogenic	-	-	OMIM:613647	C3150901:Hereditary spastic paraplegia 48	criteria provided, single submitter	clinical testing	unknown:2	Paris Brain Institute, Inserm - ICM	SCV001451119.1	-	-	-	-
2	Pathogenic	Jun 29, 2010	-	SPASTIC PARAPLEGIA 48, AUTOSOMAL RECESSIVE	C3150901:Hereditary spastic paraplegia 48	no assertion criteria provided	literature only	germline:na	OMIM	SCV000020155.3	AP5Z1	-
```

v03_pipeline/lib/reference_data/clinvar.py
jklugherz: So I was getting this error I didn't quite understand `Error: ExpressionException: Cannot combine expressions from different source objects` and I was able to resolve it by annotating `submitters` and `conditions` to `ht.info` before   parsing the other `selects`. In `parsed_clnsig` and other subsequent accesses to `ht`, the context? of ht is actually  really `ht.info` so I had to remove `info`.

I stopped getting the error after that change.

v03_pipeline/lib/reference_data/config.py
jklugherz: This was another choice that makes mocking easier - our tests were already running the `clinvar_custom_select` code so it was easiest to put the submitter/condition stuff in a new function, and in tests mock the new function to return `ht.info`.

v03_pipeline/lib/reference_data/clinvar.py
bpblanken: 👍 to keeping the functions separate.  what about moving the args given to `download_and_import_clinvar_txt_file` to `hl.import_table` though?  

and yes, that looks right about the `impute=True`!  Could we just pass `{'#VariationID': hl.tstr, 'Submitter': hl.tstr, 'ReportedPhenotypeInfo': hl.str} instead of `impute`?  I think it's faster!

v03_pipeline/lib/reference_data/config.py
jklugherz: oh wait I see what you mean here. I can annotate without passing and adding to `selects`

v03_pipeline/lib/reference_data/clinvar.py
bpblanken: Ah... I think I see what's happening here.   The selects aren't actually evaluated until `get_dataset_ht`... so the annotation that's happening when we call `join_clinvar_to_submission_summary` inside `clinvar_custom_select` is subtly creating a new `Table` under the hood and giving the `different source objects` error.

I think the easiest way out is to just modify the `custom_import` for clinvar to do both the downloading of the vcf and the annotation with the submitters.  Then our "select" becomes super straightforward.  Is that doable?

v03_pipeline/lib/reference_data/clinvar.py
bpblanken: 👍 makes sense to me!  thanks for the explanation!

v03_pipeline/lib/reference_data/clinvar.py
jklugherz: It's doable, and it does also make sense to import everything in the `custom_import` function and not sneak one into the `custom_select`

bpblanken: Hope the mocking is also doable!

v03_pipeline/lib/reference_data/clinvar.py
jklugherz: moved the args to `import_table`, got rid of `impute` and passed the types!

bpblanken (approved):

PR #745, opened 2024-03-26
add allele registry step in update vat with samples task

jklugherz:
options for where in pipeline to register the alleles (I implemented option B here):

A) in a new airflow/luigi task downstream of the annotations table tasks
- doesn't really jive with luigi expecting a file output, and I'm not sure how to get the variants to register (I could pass the variants to register between tasks with the airflow xcom, or use the callset from the input variable to re-create the `new_variants_ht`)

B) within the update annotations table task
- I can easily use the `new_variants_ht` to get what I need to call the register endpoint with
- However, this logic doesn't necessarily belong in the task and really a new step _should_ mean a new pipeline task

### DIFF
--- a/v03_pipeline/lib/tasks/update_variant_annotations_table_with_new_samples.py
+++ b/v03_pipeline/lib/tasks/update_variant_annotations_table_with_new_samples.py
@@ -1,10 +1,12 @@
 import functools
 import math
+from itertools import chain
 
 import hail as hl
 import luigi
 
 from v03_pipeline.lib.annotations.fields import get_fields
+from v03_pipeline.lib.misc.clingen_allele_registry import register_alleles
 from v03_pipeline.lib.misc.math import constrain
 from v03_pipeline.lib.misc.util import callset_project_pairs
 from v03_pipeline.lib.model import Env, ReferenceDatasetCollection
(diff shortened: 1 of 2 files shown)


bpblanken:
yeah, just in a brief look at this, my gut is leaning towards option A.  The UpdateAnnotations task is already more crowded than it should be (VEP at some point was its own task... but got moved here when we moved things around).

rather than Xcoms I'd recommend a new intermediate file, similar to what we're doing with the `WriteImportedCallset`, of just the new variants.  It probably also makes sense for VEP to be there!

v03_pipeline/lib/tasks/update_variant_annotations_table_with_new_samples.py
```diff
@@ -48,6 +50,7 @@ class UpdateVariantAnnotationsTableWithNewSamplesTask(BaseVariantAnnotationsTabl
         default=True,
         parsing=luigi.BoolParameter.EXPLICIT_PARSING,
     )
+    allele_registry_password = luigi.Parameter()
```
bpblanken, line 53: Also, just a heads up we definitely want this to be an env var and not param.  Luigi does support "hidden" params but our old elasticsearch pass was definitely getting passed in plaintext within dataproc.

PR #749, opened 2024-03-29
change clinvar submitters table partitions

jklugherz:
make them equal to the clinvar vcf min partitions to hopefully reduce flakiness of the join. this is mostly a guess

yesterday's test [run](https://console.cloud.google.com/dataproc/jobs/UpdateVariantAnnotationsTableWithUpdatedReferenceDataset_20240328_9a449403/monitoring?region=us-central1&project=seqr-project)  succeeded but after multiple attempts


bpblanken (approved):
Yeah, I think this makes sense?

PR #757, opened 2024-04-05
Benb/force task option

bpblanken:


v03_pipeline/lib/tasks/__init__.py
```diff
@@ -15,7 +14,6 @@
 )
 
 __all__ = [
-    'DeleteOldRunsTask',
```
bpblanken, line 18: This is unrelated code removal.

v03_pipeline/lib/tasks/write_new_variants_table.py
```diff
@@ -142,31 +148,35 @@ def requires(self) -> list[luigi.Task]:
...
-                            _,
-                        ) in callset_project_pairs(
-                            self.callset_paths,
-                            self.project_guids,
-                            self.project_remap_paths,
-                            self.project_pedigree_paths,
-                        )
-                    ],
+        return (
+            not self.force
```
jklugherz, line 152: do we still need this if the new variants table is in the runs directory?

jklugherz (approved):
this is awesome!!

v03_pipeline/lib/tasks/write_new_variants_table.py
bpblanken: Nope!  Good call!

PR #759, opened 2024-04-08
allele registry

jklugherz:


v03_pipeline/lib/misc/allele_registry.py
```diff
@@ -0,0 +1,76 @@
...
+        f'Registering {len(hgvs_expressions)} alleles to the Clingen Allele Registry',
+    )
+
+    # adapted from https://reg.clinicalgenome.org/doc/scripts/request_with_payload.py
+    login = Env.ALLELE_REGISTRY_LOGIN
+    password = Env.ALLELE_REGISTRY_PASSWORD
+    identity = hashlib.sha1((login + password).encode('utf-8')).hexdigest()  # noqa: S324
+    gb_time = str(int(time.time()))
+    token = hashlib.sha1((URL + identity + gb_time).encode('utf-8')).hexdigest()  # noqa: S324
+    request = URL + '&gbLogin=' + login + '&gbTime=' + gb_time + '&gbToken=' + token
```
bpblanken, line 49: `urllib.parse.urlencode` might be nice here!

v03_pipeline/lib/misc/allele_registry.py
```diff
@@ -0,0 +1,76 @@
...
+
+    for i in range(0, len(hgvs_expressions), MAX_REQUEST_SIZE):
```
bpblanken, line 51: Would [itertools.batched](https://docs.python.org/3/library/itertools.html#itertools.batched) be helpful?

jklugherz: ahh I like `batched` but it's only available in python 3.12

v03_pipeline/lib/misc/allele_registry.py
jklugherz: I tried this but even though the url was the same when I printed it out I was getting 403 error with `urllib.parse.urlencode`

v03_pipeline/lib/misc/allele_registry.py
```diff
@@ -0,0 +1,145 @@
...
+                AlleleRegistryError.from_api_response(allele_response, base_url),
+            )
+        else:
+            # Example allele_response:
+            # {'@id': 'http://reg.genome.network/allele/CA520798109',
+            #  'externalRecords': {'gnomAD_4': [{'id': '1-10109-AACCCT-A'}]}}
+            caid = allele_response['@id'].split('/')[-1]
+            if 'externalRecords' in allele_response:
+                gnomad_id = allele_response['externalRecords']['gnomAD_4'][0]['id']
+                id_map[gnomad_id] = caid
```
jklugherz, line 136: I'm currently using gnomad_id from the api response to map to CAID when [annotating](https://github.com/broadinstitute/seqr-loading-pipelines/pull/759/files#diff-2677a48e0bab4da76f54ec6184fa2179f8a688a291a46faeaf012eef58e74274R248) new variants with CAID. 

However not all variants have a gnomad ID, some have other IDs. I'm thinking of alternatives/better ways to do this mapping.

Attached is an example allele registry json response with no fields filtered for this variant: 1:10109	["AACCCT","A"]
[allele_reg_json_response_example.json](https://github.com/broadinstitute/seqr-loading-pipelines/files/15030753/allele_reg_json_response_example.json)

hanars: So in seqr we have this line of code that takes a a gnomad style ID thats been split on a "-" (literally called `variant_id.split('-')` upstream somewhere) and turns it into a locus and allele, which should be the key of the table you are working with. You could then turn the allele response into its own table keyed by locus-allele so you can join them
https://github.com/broadinstitute/seqr/blob/master/hail_search/queries/mito.py#L202

v03_pipeline/lib/misc/allele_registry.py
```diff
@@ -0,0 +1,145 @@
...
+        return cls(
+            base_url=base_url,
+            error_type=response['errorType'],
+            description=response['description'],
+            message=response['message'],
+            input_line=response.get('inputLine'),
+        )
+
+    @property
+    def loggable_message(self) -> str:
```
bpblanken, line 40: Would just implementing `__str__` or `__rpr__` work for this?

jklugherz: `__str__` is correct here because the logger will convert an object to a string, so I can just do `logger.error(e)` where `e` is an instance of `AlleleRegistryError`. thanks for the suggestion!

v03_pipeline/lib/misc/allele_registry.py
jklugherz: I like this approach, (left) joining two hail tables to annotate CAID definitely feels cleaner than annotating a new field from a dictionary.  Performance-wise I'm guessing the join is faster this way too?

v03_pipeline/lib/misc/allele_registry.py
```diff
@@ -0,0 +1,169 @@
...
+):
+    num_rows = ht.count()
+    num_chunks = math.ceil(num_rows / chunk_size)
+    logger.info(
+        f'Registering {num_rows} allele(s) in chunks of {chunk_size} in {num_chunks} request(s).',
+    )
+    for start_idx in range(0, num_rows, chunk_size):
+        end_idx = start_idx + chunk_size
+        if end_idx == chunk_size:
+            chunk_ht = ht.head(chunk_size)
```
bpblanken, line 63: this code is super cool and would be a great addition to hail.

jklugherz: I got the idea from this [discussion](https://discuss.hail.is/t/way-for-slicing-on-hail-table/2841/2) - it seems like it's kind of a hail anti-pattern (and an expensive operation?) to chunk out tables like this. 

(I tried the suggestion from the discussion to split the table by chromosomes using `parse_locus_interval` but there's way more than 1 million variants per chromosome, so it didn't seem like the right solution for our case.)
```
filtered = hl.filter_intervals(s37_ht, [hl.parse_locus_interval('1-2', reference_genome='GRCh37')])
# 7 million variants in chrom 1, this solution won't work
```

bpblanken: oh yes it's definitely a hail anti pattern to localize 🤣 

could you do something like `repartition` to many small pieces and then `export` to tsv with the `header_per_shard` option to get exported chunks?  I guess that's not too dissimilar from exporting to vcf.

v03_pipeline/lib/tasks/write_new_variants_table.py
```diff
@@ -234,6 +235,15 @@ def create_table(self) -> hl.Table:
...
             new_variants_ht = new_variants_ht.join(rdc_ht, 'left')
 
+        # Register the new variant alleles to the Clingen Allele Registry
+        # and annotate new_variants table with CAID.
+        if self.dataset_type.should_send_to_allele_registry:
+            for ar_ht in register_alleles_in_chunks(
+                new_variants_ht,
+                self.reference_genome,
+            ):
+                new_variants_ht = new_variants_ht.join(ar_ht, 'left')
```
bpblanken, line 245: There's nothing in `ar_ht` that's not in `new_variants_ht` right?

bpblanken: fwiw I do think the `left` join is correct and safest!  I guess the alternative would be `outer` which probably doesn't make sense.

jklugherz: > There's nothing in ar_ht that's not in new_variants_ht right?
that's correct. it's also possible for `ar_ht` is to be missing variants that are in the `new_variants_ht`, hence the `left` join

v03_pipeline/lib/misc/allele_registry.py
jklugherz: This is a good idea and probably the sparkier way to do it but I'd rather keep the solution as it is now instead of refactoring both the chunking and exporting (tsv vs vcf) parts of the code

v03_pipeline/lib/misc/allele_registry.py
```diff
@@ -0,0 +1,201 @@
...
+import requests
+from requests import HTTPError
+
+from v03_pipeline.lib.logger import get_logger
+from v03_pipeline.lib.model import Env, ReferenceGenome
+from v03_pipeline.lib.reference_data.clinvar import safely_move_to_gcs
+
+MAX_VARIANTS_PER_REQUEST = 1000000
+ALLELE_REGISTRY_URL = 'https://reg.genome.network/alleles?file=vcf&fields=none+@id+externalRecords.gnomAD_4.id'
+HTTP_REQUEST_TIMEOUT = 420
```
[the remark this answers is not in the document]
jklugherz: done... I had a timeout at 300 seconds so I bumped it up to 420 to be safe :/

the example Larry gave me from the AR tech lead was that registering ~3.5M variants in one request took them 4.26 minutes.

v03_pipeline/lib/misc/allele_registry.py
```diff
@@ -0,0 +1,201 @@
...
+        if use_gcs_filesystem:
+            raw_vcf_file_name = os.path.join(
+                Env.HAIL_TMPDIR,
+                os.path.basename(raw_vcf.name),
+            )
+            safely_move_to_gcs(raw_vcf.name, raw_vcf_file_name)
+        else:
+            raw_vcf_file_name = raw_vcf.name
+
+        # Reformat the VCF created by hail's 'export_vcf' function: replace the header and remove any 'chr' prefix
```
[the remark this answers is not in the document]
jklugherz: It removes 'chr' if it exists (which it does on the raw 38 vcfs) because that's what the allele registry wants. I added a comment

v03_pipeline/lib/misc/allele_registry.py
```diff
@@ -1,10 +1,11 @@
 import dataclasses
 import hashlib
 import math
-import tempfile
 import time
+import uuid
 
 import hail as hl
+import hailtop.fs as hfs
```
bpblanken, line 8: yay!

v03_pipeline/lib/model/environment.py
```diff
@@ -32,3 +34,5 @@ class Env:
     REFERENCE_DATASETS: str = REFERENCE_DATASETS
     VEP_CONFIG_PATH: str | None = VEP_CONFIG_PATH
     VEP_CONFIG_URI: str | None = VEP_CONFIG_URI
+    SHOULD_REGISTER_ALLELES: bool = SHOULD_REGISTER_ALLELES
+    ALLELE_REGISTRY_SECRET_NAME: str | None = ALLELE_REGISTRY_SECRET_NAME
```
bpblanken, line 38: nit but can we alphabetize?

v03_pipeline/lib/model/constants.py
```diff
@@ -1,3 +1,4 @@
+SEQR_GCP_PROJECT_ID = 'seqr-project'
```
bpblanken, line 1: can we just make this an env var?  I think it's one on the airflow end already.

v03_pipeline/lib/misc/allele_registry.py
```diff
@@ -0,0 +1,215 @@
...
+        res = requests.put(
+            url=build_url(base_url, reference_genome),
+            data=data,
+            timeout=HTTP_REQUEST_TIMEOUT_S,
+        )
+        return handle_api_response(res, base_url, reference_genome)
+
+
+def build_url(base_url: str, reference_genome: ReferenceGenome) -> str:
+    login, password = _get_ar_credentials_from_secret_manager()
```
bpblanken, line 110: have we tested this yet?  do we need to do something like [this stack overflow suggests](https://stackoverflow.com/a/75878048)?

jklugherz: good catch. I tested it in my julia cluster and got "Request is missing required authentication credential."
let me see if I can create a cluster with the cloud-platform scope 

in the meantime I added the scope here: https://github.com/broadinstitute/seqr-pipeline-airflow/pull/241/files#diff-e02c8035fb33918e7057de3b717c24d2f123f31e383b6558acc2808bb24851c9R168

bpblanken (approved):

PR #760, opened 2024-04-10
add annotation clinical_significances for clinvar submission

jklugherz:


v03_pipeline/lib/reference_data/clinvar_test.py
```diff
@@ -191,6 +197,12 @@ def test_join_to_submission_summary_ht(self, mock_download):
                         'PreventionGenetics, part of Exact Sciences',
                         'Invitae',
                     ],
+                    clinical_significances=[
```
bpblanken, line 200: this looks like something that we can maybe enum map?  

the process for that would look like:
- using hail to aggregate to a set and inspecting the values.
- adding the list submitters to the `enum_select` field nested under `clinvar` in `config.py`.

hanars:
I don't think this does what we want. The update based on the seqr ticket was to show the clinical significance associated with each submitter. This is aggregating the submitters and the clinical significances independently so the relationship will be lost. 

Lynn said on that ticket that we should only add this if it would be easy, and I said there that since this required upstream pipeline changes its not easy and we should not do it. The fact that there seems to be a bunch of complexity and iteration needed for this in the pipeline makes me feel that even more. I do not think we want to add this functionality now at all, and should leave this out of the MVP.

bpblanken:
FWIW I think Julia was able to do this before the comment about not proceeding:

Screenshot 2024-04-11 at 12 14 21 PM

Screenshot 2024-04-11 at 12 14 41 PM

let's discuss in standup!

PR #761, opened 2024-04-16
Delete "family" tasks

bpblanken:
Adds the three remaining tasks to handle deletion of list of families from a project.

- Delete families from the lookup table.
- Delete families from the annotations table (by re-annotating gt_stats).
- Delete families from the project table.


v03_pipeline/lib/tasks/update_project_table_with_deleted_families.py
```diff
@@ -0,0 +1,31 @@
...
+
+
+class UpdateProjectTableWithDeletedFamiliesTask(BaseProjectTableTask):
+    family_guids = luigi.ListParameter()
+
+    def complete(self) -> bool:
+        return super().complete() and hl.eval(
+            hl.bind(
+                lambda family_guids: (
+                    hl.all(
```
jklugherz, line 15: why don't we also need a line like [this](https://github.com/broadinstitute/seqr-loading-pipelines/pull/761/files#diff-ba28fb0f9bee045c745fd70e50d7ec743dff2b98cc03b7299469dc00501e294bR30) here?

bpblanken: ah!  good question.  that line is handling the case where the project is missing from the lookup table (which contains all of our projects!).  Within the project table, we don't have that notion.  If a project is missing, the table itself is missing (which would be handled by the `super().complete()`!

v03_pipeline/lib/tasks/update_lookup_table_with_deleted_families.py
```diff
@@ -0,0 +1,50 @@
...
+            family_guids=self.family_guids,
+        )
+
+    def complete(self) -> bool:
+        return super().complete() and hl.eval(
+            hl.bind(
+                lambda family_guids: (
+                    hl.is_missing(family_guids)  # The project itself is missing
+                    | hl.all(
+                        hl.array(list(self.family_guids)).map(
```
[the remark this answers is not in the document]
bpblanken: It's necessary in order to make the `.map` work within hail.   `map(func, list(self.family_guids))` _would_ work though, with the vanilla python `map`.   Or even `map(func, self.family_guids)`.  The real annoyance is `luigi.ListParameter` turning lists into tuples!

v03_pipeline/lib/tasks/update_lookup_table_with_deleted_families.py
```diff
@@ -0,0 +1,50 @@
...
+    UpdateVariantAnnotationsTableWithDeletedFamiliesTask,
+)
+
+
+class UpdateLookupTableWithDeletedFamiliesTask(BaseLookupTableTask):
+    project_guid = luigi.Parameter()
+    family_guids = luigi.ListParameter()
+
+    def requires(self) -> luigi.Task:
+        return UpdateVariantAnnotationsTableWithDeletedFamiliesTask(
```
jklugherz, line 18: why is `UpdateVariantAnnotationsTableWithDeletedFamiliesTask` a requirement of ` UpdateLookupTableWithDeletedFamiliesTask` and not vice versa? 

I ask because I noticed that the UpdateVariantAnnotationsTable calls `remove_family_guids` on the lookup table, which wouldn't need to happen if the requirements were reversed

v03_pipeline/lib/tasks/update_variant_annotations_table_with_deleted_families.py
```diff
@@ -0,0 +1,96 @@
...
+            lookup_table_path(
+                self.reference_genome,
+                self.dataset_type,
+            ),
+        )
+
+    def complete(self) -> bool:
+        if not self.dataset_type.has_lookup_table:
+            return True
+        return super().complete() and (
```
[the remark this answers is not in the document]
jklugherz: > The pipeline essentially treats a project as an immutable collection of families

can you tell me more about where in the pipeline this 'assumption' is?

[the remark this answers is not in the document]
bpblanken: Yeah... it's basically encoded in how we're using '(project_guid, callset_path)' as the completion token for an update of the annotations and lookup table.  If the families inside a project change, we are not "done" and need to actually re-do the task.  My initial guess is we should be using something like `(project_guid, callset_path, "hash of all family_guids concatted together")`.

v03_pipeline/lib/tasks/update_lookup_table_with_deleted_families.py
bpblanken: Yes, this is actually pretty tricky and probably needs better in-code comments.  We're using the lookup table itself to determine which rows of the annotations table require an update.  The rows where the family is defined are the rows that require re-annotation with `gt_stats`.  If we update the lookup table first, that information is lost and we'd have to re-annotate the entire annotations table (not the worst thing, but is computationally more expensive).

jklugherz (approved):

v03_pipeline/lib/tasks/update_lookup_table_with_deleted_families.py
jklugherz: ah nice a comment would be helpful

v03_pipeline/lib/tasks/update_variant_annotations_table_with_deleted_families.py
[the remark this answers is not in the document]
jklugherz: ahhh I get it, thanks. I like the hash idea.

PR #767, opened 2024-04-22
Delete Project & Family Table Tasks.

bpblanken:


v03_pipeline/lib/tasks/base/base_update.py
```diff
@@ -1,7 +1,7 @@
 import hail as hl
 
 from v03_pipeline.lib.misc.io import write
-from v03_pipeline.lib.tasks.base.base_hail_table_task import BaseHailTableTask
+from v03_pipeline.lib.tasks.base.base_hail_table import BaseHailTableTask
```
bpblanken, line 4: I ended up doing a some renaming of the `base` tasks since we now have a `base` task that isn't an `update`.  It sprawled... happy to move the renames to a different PR so this one becomes smaller.

v03_pipeline/lib/tasks/base/base_delete_table.py
```diff
@@ -0,0 +1,19 @@
...
+
+logger = get_logger(__name__)
+
+
+class BaseDeleteTableTask(BaseHailTableTask):
+    def complete(self) -> bool:
+        logger.info(f'DeleteTableTask: checking if {self.output().path} exists')
+        return (
+            not GCSorLocalTarget(self.output().path).exists()
+            and not GCSorLocalFolderTarget(self.output().path).exists()
```
jklugherz, line 15: Would there ever be a case where `GCSorLocalTarget(self.output().path).exists()` is false but `GCSorLocalFolderTarget(self.output().path).exists()` is true?

v03_pipeline/lib/tasks/delete_project_table.py
```diff
@@ -0,0 +1,18 @@
+import luigi
+
+from v03_pipeline.lib.paths import project_table_path
+from v03_pipeline.lib.tasks.base.base_delete_table import BaseDeleteTableTask
+from v03_pipeline.lib.tasks.files import GCSorLocalTarget
+
+
+class DeleteProjectTableTask(BaseDeleteTableTask):
```
jklugherz, line 8: Should `DeleteProjectFamilyTablesTask` be a dependency of this task?

v03_pipeline/lib/tasks/base/base_delete_table.py
bpblanken: yes!  The "Folder" target checks that there's a `_SUCCESS` file inside the path.  So in many cases they do not give the same answer.

v03_pipeline/lib/tasks/delete_project_table.py
bpblanken: oof, excellent question.  my initial thought was no, that we should be able to delete a project able in isolation.  however, I'm re-thinking that.

bpblanken: I'm gonna go ahead and add the dependency.  It's not clear to me what the right move is... but might as well put it there for completeness!

jklugherz (approved):

PR #770, opened 2024-04-25
Ensure rows are deleted after deleting samples!

bpblanken:


v03_pipeline/lib/misc/lookup.py
```diff
@@ -140,6 +140,7 @@ def remove_project(
             )
         ),
     )
+    ht = ht.filter(hl.any(ht.project_stats.map(hl.is_defined)))
```
bpblanken, line 143: subtle difference between the two filters.... this one only checks whether a "project" is defined, the other goes deeper and checks for the presence of at least one family.

v03_pipeline/lib/tasks/update_variant_annotations_table_with_deleted_families.py
```diff
@@ -25,72 +21,33 @@ def __init__(self, *args, **kwargs):
...
-        )
+        # We don't have the concept of families being present or not present
+        #  in annotations table a done flag to prevent the task from looping over itself.
+        return super().complete() and self.done
 
     def update_table(self, ht: hl.Table) -> hl.Table:
         if not self.dataset_type.has_lookup_table:
             return ht
         lookup_ht = hl.read_table(self.input().path)
-        project_i = hl.eval(lookup_ht.globals.project_guids.index(self.project_guid))
```
bpblanken, line 66: I decided to just remove this optimization (to only annotate changed rows) because the logic got even more confusing with the lookup table potentially changing its row count.

v03_pipeline/lib/misc/lookup.py
```diff
@@ -98,15 +98,15 @@ def remove_family_guids(
...
-        project_guid,
-    )  # double reference because new expression
+    ht = ht.filter(
+        hl.any(ht.project_stats.map(lambda fs: hl.any(fs.map(hl.is_defined)))),
+    )
     return ht.annotate_globals(
         project_families=hl.dict(
-            hl.enumerate(ht.project_families.items()).starmap(
-                lambda i, item: (
+            ht.project_families.items().map(
```
bpblanken, line 106: I don't think `ht.project_families.items()` is guaranteed to return items in order... so this is more correct if less consistent with the rest of the code.

hanars (approved):

PR #774, opened 2024-04-30
vets filters

bpblanken:


v03_pipeline/lib/misc/io.py
```diff
@@ -11,6 +11,44 @@
...
+                    ~is_snp & (split_cs > VETS_INDEL_CUTOFF),
+                    hl.if_else(
+                        hl.is_defined(mt.filters),
+                        mt.filters.add(VETS_INDEL_FILTER),
+                        hl.set([VETS_INDEL_FILTER]),
+                    ),
+                )
+                .default(
+                    mt.filters,
+                )
```
matren395, line 45: much prettier than multiple separate annotate statements

v03_pipeline/lib/misc/io.py
```diff
@@ -11,6 +11,44 @@
...
+
+def annotate_vets(mt: hl.MatrixTable) -> hl.MatrixTable:
+    if not hasattr(mt, 'info') or not hasattr(mt.info, 'CALIBRATION_SENSITIVITY'):
+        return mt
+    return mt.annotate_rows(
+        filters=hl.bind(
+            lambda is_snp, split_cs: (
+                hl.case()
+                .when(
+                    is_snp & (split_cs > VETS_SNP_CUTOFF),
```
matren395, line 28: briefly checked - their header documentation for the vcf just uses '>' and not '>=' so gtg here

v03_pipeline/lib/tasks/write_imported_callset.py
```diff
@@ -85,6 +86,7 @@ def create_table(self) -> hl.MatrixTable:
         mt = select_relevant_fields(mt, self.dataset_type)
         if self.dataset_type.has_multi_allelic_variants:
             mt = split_multi_hts(mt)
+        mt = annotate_vets(mt)
```
matren395, line 89: VETS is allele-specific, so it needs to be ran on a split MT , which it looks like is covered here. And if the prior 'if' statement isn't triggered, then it wouldn't be a problem... I don't think?

v03_pipeline/lib/misc/io_test.py
```diff
@@ -17,3 +24,97 @@ def test_compute_hail_n_partitions(self) -> None:
...
+                    hl.set(['PASS']),
+                ],
+                'info': [
+                    hl.Struct(CALIBRATION_SENSITIVITY=['0.999']),
+                    hl.Struct(CALIBRATION_SENSITIVITY=['0.995']),
+                    hl.Struct(CALIBRATION_SENSITIVITY=['0.999']),
+                    hl.Struct(CALIBRATION_SENSITIVITY=['0.98']),
+                    hl.Struct(CALIBRATION_SENSITIVITY=['0.99']),
+                    hl.Struct(CALIBRATION_SENSITIVITY=['0.991']),
+                ],
```
matren395, line 103: At the moment, VETS is only ran on variants that pass all other filters and are in high-quality genotypes, essentially. Variants with 'AC0' and 'ExcessHet' don't have VETS scores. This is by design, and the same behavior as VQSR. 

Would we want to include a test case to future-proof ourselves if this behavior changes, or is that a future Ben/Dan problem

v03_pipeline/lib/misc/io_test.py
```diff
@@ -17,3 +24,97 @@ def test_compute_hail_n_partitions(self) -> None:
...
+            },
+            cols={'s': ['sample_1']},
+            entries={'HL': [[0.0], [0.0], [0.0], [0.0], [0.0], [0.0]]},
+        ).key_rows_by('locus', 'alleles')
+        dragen_mt = split_multi_hts(dragen_mt)
+        dragen_mt = annotate_vets(dragen_mt)
+        self.assertListEqual(
+            dragen_mt.filters.collect(),
+            [
+                {'high_CALIBRATION_SENSITIVITY_SNP', 'PASS'},
```
matren395, line 113: Hmm give me a second to rethink this - is this the desired behavior? If it's failing VETS than it shouldn't be a PASS I don't think .

bpblanken: my guess is we should actually remove `PASS`.  That's straightforwards!

v03_pipeline/lib/tasks/write_imported_callset.py
bpblanken: yep, this logic was placed explicitly after the `split_multi`.  The `if` statement will always be true for `SNV_INDEL`.  it might be even clearer to slot `annotate_vets` under an `if` statement as well, but the function will be a no-op if the `info` field is missing.

v03_pipeline/lib/misc/io_test.py
bpblanken: that might be outside our purview... 

would we actually want to add code that says `if AC0 is present don't check VETs`?  that's a line of code we can add a test for!

v03_pipeline/lib/misc/io_test.py
bpblanken: This is implemented!

v03_pipeline/lib/misc/vets.py
```diff
@@ -0,0 +1,40 @@
...
+        return mt
+    return mt.annotate_rows(
+        filters=hl.bind(
+            lambda is_snp, split_cs: (
+                hl.case()
+                .when(
+                    is_snp & (split_cs > VETS_SNP_CUTOFF),
+                    hl.if_else(
+                        hl.is_defined(mt.filters),
+                        mt.filters.add(VETS_SNP_FILTER).remove(PASS_FILTER),
```
matren395, line 21: would Hail's set remove throw an error if 'PASS' isn't present? I know sometimes callsets can be inconsistent if it's passed as 'PASS' versus just None

v03_pipeline/lib/misc/vets_test.py
```diff
@@ -0,0 +1,104 @@
...
+                        reference_genome='GRCh38',
+                    ),
+                ],
+                'alleles': [
+                    ['A', 'T'],
+                    ['A', 'T'],
+                    ['A', 'T'],
+                    ['AC', 'T'],
+                    ['AT', 'TC'],
+                    ['AG', 'TG'],
```
matren395, line 70:
```suggestion
                    ['AC', 'T'],
                    ['AT', 'ATC'],
                    ['AG', 'ATG'],
```

petty note - don't think variants in Hail can look like that after min-rep, but I'm aware this is super minor to note

v03_pipeline/lib/misc/vets.py
bpblanken:
```
In [5]: hl.set(['abc']).remove('10').remove('abc').remove('s').show()
+----------+
| <expr>   |
+----------+
| set<str> |
+----------+
| {}       |
+----------+

```

nope! we're all good.

matren395 (approved):
lgtm! Arrows are facing the right way, their documentation is also non-inclusive (> not >=) , and the scores are defined in their documentation rather than being depending on the callset (as I understand it).

v03_pipeline/lib/misc/vets.py
```diff
@@ -0,0 +1,40 @@
...
+                .when(
+                    ~is_snp & (split_cs > VETS_INDEL_CUTOFF),
+                    hl.if_else(
+                        hl.is_defined(mt.filters),
+                        mt.filters.add(VETS_INDEL_FILTER).remove(PASS_FILTER),
+                        hl.set([VETS_INDEL_FILTER]),
+                    ),
+                )
+                .default(
+                    mt.filters,
```
hanars, line 34: do you need to add the `.remove(PASS_FILTER)` to this case as well?

v03_pipeline/lib/tasks/write_imported_callset.py
hanars: I think there should be a comment here explaining why we are adding this one specific annotation here when this function otherwise does not do that. Alternatively, maybe make it a dataset-type-specific, generically named function, something like `annotate_imported_mt`

v03_pipeline/lib/misc/vets.py
bpblanken: I don't think so!  that would remove `PASS` from the variants that do actually pass!

hanars: GATK has an empty array for filters if they pass. There is no special handling in the hail backend if PASS is included, seqr will filter out variants as if they failed a filter if theres anything in the filter list

bpblanken: ah, ok!  adding `.remove(PASS_FILTER)` was a complete mistake then.  It ended up in the tests as an example of an existing filter to preserve, but `PASS` isn't actually present in these vcfs.  I'll fix!

v03_pipeline/lib/misc/vets.py
```diff
@@ -0,0 +1,40 @@
+import hail as hl
+
+PASS_FILTER = 'PASS'  # noqa: S105
```
hanars, line 3: I don't think this is used anymore?

hanars (approved):

matren395:
As a very interested party that I'm in communication with, any objections if I shared this code with Konrad K. ? This is a public repo and the code is in dev , but just lmk if that's not the best choice for whatever reason

bpblanken:
@matren395 nope!  no objections.

PR #778, opened 2024-05-02
Implement DRAGEN predicted sex import.

bpblanken:


v03_pipeline/lib/methods/sex_check.py
```diff
@@ -1,58 +0,0 @@
...
-BIALLELIC: int = 2
-XX_FSTAT_THRESHOLD: float = (
-    0.5  # F-stat threshold below which a sample will be called XX
-)
-XY_FSTAT_THRESHOLD: float = (
-    0.75  # F-stat threshold above which a sample will be called XY.
-)
-
-
-def call_sex(mt: hl.MatrixTable) -> hl.Table:
```
bpblanken, line 25: Removed in favor of importing the file.

v03_pipeline/lib/misc/callsets.py
```diff
@@ -6,13 +6,14 @@
...
 
-def get_callset_ht(
+def get_callset_ht(  # noqa: PLR0913
     reference_genome: ReferenceGenome,
     dataset_type: DatasetType,
     callset_paths: list[str],
     project_guids: list[str],
     project_remap_paths: list[str],
     project_pedigree_paths: list[str],
+    imputed_sex_paths: list[str] | None,
```
matren395, line 16: so we're storing callset-wide metric tables, yeah? Seems like the right implementation. Thoughts on some new `gs://seqr-loading-temp/v03/GRCh38/SNV_INDEL/callset_metrics/` ?

v03_pipeline/lib/misc/callsets.py
```diff
@@ -49,18 +51,34 @@ def callset_project_pairs(
...
```
matren395, line 54: if this logic works for callset-wide callset (duh) pairs , it should absolutely work for callset-wide predicted sex metrics yea ?

v03_pipeline/lib/misc/family_loading_failures_test.py
```diff
@@ -178,16 +178,16 @@ def test_all_relatedness_checks(self):
...
-                {'s': 'ROS_006_18Y03228_D1', 'sex': 'F'},
-                {'s': 'ROS_007_19Y05919_D1', 'sex': 'F'},
-                {'s': 'ROS_007_19Y05939_D1', 'sex': 'F'},
-                {'s': 'ROS_007_19Y05987_D1', 'sex': 'F'},
+                {'s': 'ROS_006_18Y03226_D1', 'predicted_sex': 'M'},
+                {'s': 'ROS_006_18Y03227_D1', 'predicted_sex': 'F'},
+                {'s': 'ROS_006_18Y03228_D1', 'predicted_sex': 'F'},
+                {'s': 'ROS_007_19Y05919_D1', 'predicted_sex': 'F'},
+                {'s': 'ROS_007_19Y05939_D1', 'predicted_sex': 'F'},
+                {'s': 'ROS_007_19Y05987_D1', 'predicted_sex': 'F'},
```
matren395, line 186: could test for an 'F' inappropriately called 'M' : maybe just swap the sex for the last one , F->M instead of M->F

matren395 (commented):
commenting on as much as I follow - just a few questions

v03_pipeline/lib/misc/callsets.py
bpblanken: we do store the "sex_check" table, which is pretty much just this now.  

in general all the "paths" used here are considered "arbitrary", so we don't have a way of computing them from the inputs in the way we do for the rest of the pieces of the pipeline.

v03_pipeline/lib/misc/callsets.py
bpblanken: I think so!

v03_pipeline/lib/misc/family_loading_failures_test.py
bpblanken: good idea!

matren395 (commented):
in that case, it looks good to me ! I'll hold off approving til you think it's ready though

matren395 (approved):
if it's good to you then lgtm !!!

PR #780, opened 2024-05-08
Validate imputed sex results against the callset.

bpblanken:
Add some validation over the ploidys in the callset, that they align with what's reported in the imputed sex file!


v03_pipeline/lib/misc/validation.py
```diff
@@ -56,6 +56,30 @@ def validate_expected_contig_frequency(
...
+            )
+            | (
+                # At least one call is haploid but the sex is Female/Unknown
+                hl.agg.any(~mt.GT.is_diploid())
+                & (sex_check_ht[mt.s].predicted_sex != Sex.MALE.value)
+            )
+        ),
+    )
+    discrepant_rate = mt.aggregate_cols(hl.agg.fraction(mt.discrepant))
+    if discrepant_rate:
```
bpblanken, line 78: I set this up on purpose so that it will throw if the `discrepant_rate` is greater than 0... so the most sensitive possible threshold.

v03_pipeline/lib/misc/validation.py
```diff
@@ -56,6 +56,30 @@ def validate_expected_contig_frequency(
...
+    mt: hl.MatrixTable,
+    sex_check_ht: hl.Table,
+) -> None:
+    mt = mt.select_cols(
+        discrepant=(
+            (
+                # All calls are diploid but the sex is Male/Unknown
+                hl.agg.all(mt.GT.is_diploid())
+                & (sex_check_ht[mt.s].predicted_sex != Sex.FEMALE.value)
+            )
```
matren395, line 69: hmm I remember from the docs that only confirmed male XY samples have ploidy altered, and I don't think anything else has ploidy touched. I think we'd expect Unknown to be fully diploid. Do we want to throw an error if it's unknown regardless?

v03_pipeline/lib/misc/validation.py
matren395: that tracks , but qq - wouldn't this just be checking if the discrepant_rate is none? or for float values, does 'if' check if it's non-zero?

v03_pipeline/lib/misc/validation.py
```diff
@@ -56,6 +56,30 @@ def validate_expected_contig_frequency(
...
+                # At least one call is haploid but the sex is Female/Unknown
+                hl.agg.any(~mt.GT.is_diploid())
+                & (sex_check_ht[mt.s].predicted_sex != Sex.MALE.value)
+            )
+        ),
+    )
+    discrepant_rate = mt.aggregate_cols(hl.agg.fraction(mt.discrepant))
+    if discrepant_rate:
+        msg = f'{discrepant_rate:.2%} of samples have misaligned ploidy with their provided imputed sex.  Contact the methods team to investigate.'
+        raise SeqrValidationError(msg)
```
matren395, line 80: hmm questions: 
- what exactly does the failure mode look like here? it would totally fail the loading during the validation step, yeah? 
- what would the methods team investigating this look like ? would we exclude this family from loading by editing what all is subsetted and/or removing it from the airtable? trying to take a step back

v03_pipeline/lib/misc/validation_test.py
```diff
@@ -29,6 +32,57 @@ def _mt_from_contigs(contigs):
 
 
 class ValidationTest(unittest.TestCase):
+    def test_validate_imputed_sex_ploidy(self) -> None:
+        sex_check_ht = hl.read_table(TEST_SEX_CHECK_1)
+        mt = hl.MatrixTable.from_parts(
+            rows={
+                'locus': [
+                    hl.Locus(
+                        contig='chr1',
```
matren395, line 41:
```suggestion
                        contig='chrX',
```

v03_pipeline/lib/misc/validation_test.py
```diff
@@ -29,6 +32,57 @@ def _mt_from_contigs(contigs):
...
+                    ],
+                ],
+            },
+        ).key_rows_by('locus')
+        validate_imputed_sex_ploidy(mt, sex_check_ht)
+        mt = hl.MatrixTable.from_parts(
+            rows={
+                'locus': [
+                    hl.Locus(
+                        contig='chr1',
```
matren395, line 62:
```suggestion
                        contig='chrX',
```

v03_pipeline/lib/misc/validation_test.py
```diff
@@ -29,6 +32,57 @@ def _mt_from_contigs(contigs):
...
+                    [
+                        hl.Call(alleles=[0], phased=False),
+                        hl.Call(alleles=[0], phased=False),
+                    ],
+                ],
+            },
+        ).key_rows_by('locus')
+        self.assertRaisesRegex(
+            SeqrValidationError,
+            '50.00% of samples have misaligned ploidy',
```
matren395, line 80: hmm, what's in the validation ht? both would be flagged as having haploid calls, but I'm assuming _1 is predicted female and _2 is predicted male?

matren395 (commented):
first review - let me get some tea and look at it again

v03_pipeline/lib/misc/validation.py
bpblanken: `0.0` is falsey!

v03_pipeline/lib/misc/validation.py
bpblanken: yeah.. the failure mode is actually quite different for this.... it'll totally fail loading.  These "validation" errors occur at the very entrance to the pipeline, when we're trying to import the callset from VCF.  

I just pulled the exception message from [here](https://github.com/broadinstitute/seqr-loading-pipelines/pull/780/files/772e2b7b3aebd99964ab8b979f8fe8e46df54e83#diff-bfb21e2b0d4e857967483c2ef096c5814dc2644bb9aea28496af70c52dd7fe37R19), so maybe not the clearest... but to me, if there's an issue with the ploidy that points to an issue upstream that we'd need to at least email our DSP friends about.

v03_pipeline/lib/misc/validation_test.py
bpblanken: I've got a print out [here](https://github.com/broadinstitute/seqr-loading-pipelines/blob/main/v03_pipeline/lib/tasks/write_remapped_and_subsetted_callset_test.py#L26-L47) but yes `HG00731_1` is female and `HG00732_1` is male.

v03_pipeline/lib/misc/validation.py
matren395: hmm that's cool what what I thought was up, but def wanted to check!

v03_pipeline/lib/misc/validation.py
matren395: that's fair - and honestly if our observed ploidy disagrees with their metadata for that exact thing, then I could see that deserving a full stop and reporting to our DSP friends:) 

what would the fix be for this, though? Either on their end or ours?

v03_pipeline/lib/misc/validation_test.py
matren395: yeah that tracks 

would we want a test for unknown?

matren395 (commented):
methods stuff makes sense I think, just responding to comments

v03_pipeline/lib/misc/validation.py
bpblanken: hrm, I think you're correct that we should be more explicit about Unknowns.  

We have [this function](https://github.com/broadinstitute/seqr-loading-pipelines/pull/780/files#diff-bfb21e2b0d4e857967483c2ef096c5814dc2644bb9aea28496af70c52dd7fe37R14) that will throw even _earlier_ in the pipeline if there's too many Unknown in the imputed sex file... but now that we're not running a statistical method ourselves, maybe we should just throw if anything is Unknown?

matren395 (approved):

PR #781, opened 2024-05-09
add `check_ref` and `dont_skip` to VEP

bpblanken:


v03_pipeline/var/vep_config/vep-GRCh38.json
```diff
@@ -2,10 +2,10 @@
...
+    "/vep --format vcf -json --hgvs --biotype --canonical --mane --numbers --minimal --allele_number --no_stats --cache --offline --assembly GRCh38 --fasta /opt/vep/.vep/Homo_sapiens.GRCh38.dna.primary_assembly.fa.gz --check_ref --dont_skip --plugin LoF,loftee_path:/plugins,gerp_bigwig:/opt/vep/.vep/gerp_conservation_scores.homo_sapiens.GRCh38.bw,human_ancestor_fa:/opt/vep/.vep/human_ancestor.fa.gz,conservation_file:/opt/vep/.vep/loftee.sql --plugin UTRAnnotator,file=/opt/vep/.vep/uORF_5UTR_GRCh38_PUBLIC.txt --plugin SpliceRegion,Extended --plugin AlphaMissense,file=/opt/vep/.vep/AlphaMissense_hg38.tsv.gz --dir_plugins /plugins -o STDOUT | sed s/5utr/fiveutr/g"
```
bpblanken, line 5: To describe the behavior here:

- adding `--check_ref` only will lead to no output if the reference allele is wrong.
- adding `--dont_skip` only appears to do nothing.
- adding both leads to a `check_ref="failed"` flag in the output.

There is additionally a `--lookup_ref` which will overwrite `the supplied reference allele with the sequence stored` in the supplied FASTA.  

I've been trying to get VEP to hard fail on _literally any error_ and cannot.... it just logs to a warnings file :/

jklugherz (approved):

PR #783, opened 2024-05-10
Bugfix to ensure the validation/gnomad_qc table is updated before being used, + a unit test!

bpblanken:
Bug appeared when updating gnomad!


v03_pipeline/lib/tasks/reference_data/update_cached_reference_dataset_queries.py
```diff
@@ -11,7 +11,7 @@
 )
 
 
-class WriteCachedReferenceDatasetQuery(luigi.Task):
+class UpdateCachedReferenceDatasetQueries(luigi.Task):
```
bpblanken, line 14: I renamed this to clarify that it's a wrapper around `UpdatedCachedReferenceDatasetQuery`.

jklugherz (approved):
thanks for fixing this!

PR #785, opened 2024-05-14
Benb/validate with allele type

bpblanken:


v03_pipeline/lib/misc/validation.py
```diff
@@ -11,6 +11,19 @@ class SeqrValidationError(Exception):
...
+    mt: hl.MatrixTable,
+) -> None:
+    ht = mt.rows()
+    ht = ht.filter(
+        hl.numeric_allele_type(ht.alleles[0], ht.alleles[1])
+        == hl.genetics.allele_type.AlleleType.UNKNOWN,
+    )
+    if ht.count() > 0:
+        msg = f'Alleles with Unknown AlleleType are present in the callset: {ht.alleles.collect()}'
+        raise SeqrValidationError(msg)
```
jklugherz, line 24: What's the next step for humans to take if this is the case? do we ask for a new callset? 
Would we ever want to soft fail on unknown alleles?

bpblanken: great question.  I think historically we've sent emails to our collaborators with instructions for them to resolve.  we can also filter them out manually.

hanars: > Would we ever want to soft fail on unknown alleles?

If we find that a huge percentage of AnVIL requests are failing on this and its creating an unmanageable support burden for our team we might reconsider the hard-fail behavior. We had a similar thing with unexpected contigs where we used to hard fail and then so much external data had them that we instead decided to just filter them out and not report an error.
Our intuition here is that the presence of unknown allele types probably means the way the data was called does not result in the data format we expect so filtering them out would basically result in only a fraction of data being available in seqr, and that the preferred user experience would be to fix the data and actually have all the calls in seqr. So hard failing feels preferable for now, and we can reconsider later if we have to

jklugherz (approved):

PR #788, opened 2024-05-16
Benb/vep annotations

bpblanken:


v03_pipeline/lib/annotations/vep_test.py
```diff
@@ -0,0 +1,231 @@
...
+                        transcript_id='ENST00000327044',
+                        mane_select='NM_001005221.2',
+                        mane_plus_clinical=None,
+                        biotype_id=39,
+                        consequence_term_ids=[12],
+                        exon=[15, 19],
+                        intron=None,
+                        is_lof_nagnag=None,
+                        lof_filter_ids=[0, 1],
+                        alphamissense=hl.Struct(pathogenicity=0.10000000149011612),
```
bpblanken, line 157: A few notes on the schema:

- New fields are `exon`, `intron`, `mane_select`, `mane_plus_clinical`, `alphamissense`, and `utrannotator`.  
- `SpliceRegion` annotations are included in `consequence_term_ids`.  
- I dropped the alphamissense class annotation in favor of dynamic computation (probably in the UI?) given [this formula](https://github.com/Ensembl/VEP_plugins/blob/release/112/AlphaMissense.pm#L52-L58).
- UTRAnnotator [consequences are defined as an enum](https://github.com/broadinstitute/seqr-loading-pipelines/pull/788/files#diff-dbaf5c430bf58fa8c17ba736be2b20759057d1c25a60ee777c54dd89cb0a1784R121-R126).

v03_pipeline/lib/annotations/vep.py
```diff
@@ -0,0 +1,145 @@
...
+def _transcript_consequences_select(
+    reference_genome: ReferenceGenome,
+) -> Callable[[hl.StructExpression], hl.StructExpression]:
+    if reference_genome == ReferenceGenome.GRCh38:
+        return lambda c: c.select(
+            *SELECTED_ANNOTATIONS,
+            *MANE_SELECT_ANNOTATIONS,
+            biotype_id=BIOTYPE_LOOKUP[c.biotype],
+            consequence_term_ids=_consequence_term_ids(c),
+            exon=c.exon.split('/').map(hl.parse_int32),
```
bpblanken, line 72: TODO: use a struct instead of a list for performance reasons.

v03_pipeline/lib/annotations/vep_test.py
hanars: > I would like to nest the different plugins under hl.Structs (which would mean moving is_lof_nagnag and lof_filter_ids under loftee, but left that undone to preserve consistency with GRCh37 VEP.

If we think this is the better approach we can change it here, its not too hard to have the seqr UI be able to handle both formats and we have no plans to use this as a search field

hanars: > SpliceRegion annotations are included in consequence_term_ids

Is there a reason not to keep it as a separate annotation? We are going to want these to be displayed differently so its gonna require some rough logic to pull these aprat. Also, I am worried how this enumeration affects the `consequence_term_id` which is currently a sorted list that is used for sorting search results

hanars: > UTRAnnotator annotation is a big unknown from me on utility. 

We should ask Lynn tolook into this and figure out what we want

hanars: For `mane_select`, we are currently adding this information from the seqr reference data db, and we just have a boolean indicator of whether or not a transcript is mane sleect, and then we also include the refseq id for all transcripts regardless of whether they are mane or not. It looks like here we are assigning the refseq id as the `mane_sleect` value,  and there can be multiple transcripts wth this value populated? s that correct?

bpblanken: 1) I'll update to be nested under a `loftee` key.  I think it's cleaner.
2) There was actually a bug here.  The three consequence terms that `SpliceRegion` adds (`splice_donor_5th_base_variant`, `splice_donor_region_variant`, and `splice_polypyrimidine_tract_variant`) [as documented here](https://github.com/Ensembl/VEP_plugins/blob/release/112/SpliceRegion.pm#L43-L58) were added as VEP consequence terms in [version 105](https://github.com/Ensembl/ensembl-vep/releases/tag/release%2F105.0) and were added to our consequence_terms [straight from the VEP consequence terms page](https://www.ensembl.org/info/genome/variation/prediction/predicted_data.html).  

I dove into this and realized I had a misunderstanding of what SpliceRegion was doing, and thought that the plugin itself was adding those fields to the `consequence_terms`.   Instead, they're being annotated by [VEP itself](https://github.com/Ensembl/ensembl-variation/blob/release/108/modules/Bio/EnsEMBL/Variation/Utils/VariationEffect.pm#L540-L544), and what we really want are the [two extended SpliceRegion flags](https://github.com/Ensembl/VEP_plugins/blob/release/112/SpliceRegion.pm#L148-L153) which will come in as a separate field.  I will update.  

3) We're pulling `mane_select` straight off of the VEP.  It looks like in practice each transcript has a unique id or is `null`.  The duplicates are just bad mock data.

bpblanken: The actual best I could do for `SpliceRegion` is to add a boolean if `extended_intronic_splice_region_variant` is present.  Need to ask Lynn is this is even worth doing if we're getting the other three (`splice_donor_5th_base_variant`, `splice_donor_region_variant`, `splice_polypyrimidine_tract_variant`) as VEP builtins.

PR #789, opened 2024-05-20
Reference Data Update Validity Test

bpblanken:


v03_pipeline/lib/reference_data/compare_globals.py
```diff
@@ -69,32 +73,52 @@ def from_ht(
...
-                    set(ht[dataset])
+                    {k: v.dtype for k, v in dict(ht[dataset]).items()}
                     if isinstance(ht[dataset], hl.StructExpression)
-                    else set()
+                    else {}
                 )
         return cls(paths, versions, enums, selects)
 
 
+def validate_selects_types(
```
bpblanken, line 83: The idea here is to just ensure that any fields shared between the existing table and an "update" are identically typed.  

imo we're getting very close to needing a statically defined schema for these, but trying to squeeze that in was much messier.

v03_pipeline/lib/tasks/reference_data/update_variant_annotations_table_with_updated_reference_dataset_test.py
```diff
@@ -45,7 +45,7 @@
         hl.tstruct(
             locus=hl.tlocus('GRCh38'),
             alleles=hl.tarray(hl.tstr),
-            PHRED=hl.tint32,
+            PHRED=hl.tfloat32,
```
bpblanken, line 48: we had a few types in our tests/mock data that weren't right and doing this work caught them 🤩 !

v03_pipeline/lib/reference_data/compare_globals_test.py
```diff
@@ -183,11 +194,12 @@ def test_from_rdc_or_annotations_ht(self):
         self.assertTrue(
             rdc_globals.enums == {'screen': {'region_type': ['C', 'D']}},
         )
+        print(rdc_globals.selects)
```
hanars, line 197: I assume this is left in by mistake?

hanars (approved):

PR #797, opened 2024-06-03
Import fields with fully qualified names.

bpblanken:


v03_pipeline/lib/annotations/sv.py
```diff
@@ -72,7 +82,7 @@ def _sv_types(ht: hl.Table) -> hl.ArrayExpression:
 
 
 def algorithms(ht: hl.Table, **_: Any) -> hl.Expression:
-    return hl.str(',').join(ht.info.ALGORITHMS)
+    return hl.str(',').join(ht['info.ALGORITHMS'])
```
hanars, line 85: this change feels like its less readable/ maintainable than the current syntax. Whats the motivation for this change?

bpblanken: selecting individual fields `info.ALGORITHMS` rather than just `info` at was substantially easier than selecting `info` and trying to type assert fields from a nested struct.

One potential option would be to select the nested fields and follow hail's default behavior of lifting the select to the top unnested level:

```
ht = ht.select(ht.info.ALGORITHMS)
ht.ALGORITHMS
```

leaving this line as:

```
hl.str(',').join(ht.ALGORITHMS)
```

I thought it was more maintainable to leave the `info` in here so that the structure of the import was preserved, but it definitely _looks_ worse.

hanars: yeah I might wait for @jklugherz to come back to see if she has a preference, since both options have their cons. Personally i feel like lifting things to the top level is more readabe, but thats not like an objective choice

bpblanken: yep that seems reasonable!

v03_pipeline/lib/tasks/write_imported_callset.py
```diff
@@ -119,7 +119,31 @@ def create_table(self) -> hl.MatrixTable:
...
+                        and self.dataset_type.check_sex_and_relatedness
+                    )
+                    else []
+                ),
+                *(
+                    # this field is never required, the pipeline
+                    # will run smoothly even in its absence, but
+                    # will trigger special handling if it is present.
+                    ['info.CALIBRATION_SENSITIVITY']
+                    if hasattr(mt, 'info')
```
jklugherz, line 141: Do we need this line? If we just check `hasattr(mt.info, 'CALIBRATION_SENSITIVITY')` and `info` doesn't exist on `mt`, won't we will get a readable `AttributeError`?

I guess the difference is that if `info` doesn't exist on `mt` and we keep this line `hasattr(mt, 'info')`, the error would occur inside `select_relevant_fields`, which might be preferable.

v03_pipeline/lib/misc/io.py
```diff
@@ -147,16 +148,25 @@ def import_callset(
...
-    optional_row_fields = [
-        row_field
-        for row_field in dataset_type.optional_row_fields
-        if hasattr(mt, row_field)
-    ]
-    mt = mt.select_rows(*dataset_type.row_fields, *optional_row_fields)
-    mt = mt.select_cols(*dataset_type.col_fields)
-    return mt.select_entries(*dataset_type.entries_fields)
+    mt = mt.select_rows(
+        **{field: parse_nested_field(mt, field) for field in dataset_type.row_fields},
```
jklugherz, line 155: This is minor but the code to create and unpack a dictionary of fields is repeated here so you could create a helper function that does this, takes arguments 'mt' and 'field_type' or something and returns 
```
**{field: parse_nested_field(mt, field) for field in dataset_type[field_type]}
```
where field_type could be row_fields, col_fields, or entries_fields

v03_pipeline/lib/annotations/sv.py
jklugherz: hmm before I can comment on this I am confused why the original syntax needs to be changed at all

bpblanken: The drive behind the change was changing the initial select to be nested, so selecting `info.ALGORITHMS` when we import the callset rather than just `info`.  This leads to the fields being unnested from an inner struct at "formatting annotation" time.  

The decision here is whether or not it's cleaner and more readable to leave the `info.` prefix in the unnested annotation names or leave hail's default behavior of renaming the field to the leaf function.  The actual change is [here](https://github.com/broadinstitute/seqr-loading-pipelines/pull/797/files#diff-dea15076a8ab788d04e8e4a78eabb479d472e5f389152cea68d0adabe99e141dR166).  My preference was to leave `info` in the field names so that the imported callset would at least pseudo-match the structure of the raw callset, but think it _barely_ matters because we're changing the structure anyways.  

We have occasionally run into issues with the default behavior (e.g. `info.AF` and `AF` both being present on the old pipeline tables) but don't think that's going to be a serious issue going forwards.

v03_pipeline/lib/tasks/write_imported_callset.py
bpblanken: ah yeah, I think we don't want to throw an error if the field isn't present.  it's basically a flag for whether or not the callset is DRAGEN and is not expected in non-DRAGEN callsets.

v03_pipeline/lib/annotations/sv.py
jklugherz: I think it's best to leave it how you have it, doing a select to lift the nested field to the top is more confusing imo

jklugherz (approved):

v03_pipeline/lib/misc/io.py
bpblanken: yeah, I actually don't know on this one.  I'm pretty lukewarm on one line functions because of the additional layer of indirection but can see the case here.

My guess is there's some better deduplication that can happen with the reference data selection [here](https://github.com/broadinstitute/seqr-loading-pipelines/pull/797/files/1172bf69c49bd71077309c293cbb9bbca27d46d1#diff-8e147d1242ec201292bdadc21e13034e1e0e4ce60ccccb053f1833ee3abd02fdR112).  I didn't go as in-depth as I could have on unification.

PR #798, opened 2024-06-04
Add Type Assertions to imported fields.

bpblanken:


v03_pipeline/lib/tasks/write_imported_callset.py
```diff
@@ -122,27 +141,14 @@ def create_table(self) -> hl.MatrixTable:
...
-                    ['info.CALIBRATION_SENSITIVITY']
-                    if hasattr(mt, 'info')
-                    and hasattr(mt.info, 'CALIBRATION_SENSITIVITY')
-                    else []
-                ),
-            ],
+            self.additional_row_fields(mt),
+        )
+        # This validation isn't override-able.  If a field is the wrong
+        # type, the pipeline will likely hard-fail downstream.
```
hanars, line 147: or even worse, the pipeline will succeeed, data will update, and then search will totally break in seqr!

hanars (approved):

v03_pipeline/lib/misc/io.py
```diff
@@ -148,7 +140,7 @@ def import_callset(
 def select_relevant_fields(
     mt: hl.MatrixTable,
     dataset_type: DatasetType,
-    additional_row_fields: None | list[str] = None,
+    additional_row_fields: None | dict[str, hl.expr.types.HailType | set] = None,
```
jklugherz, line 143: These type hints are just for us right? I remember you said something about why we can't use mypy but can you remind me

jklugherz (approved):
yahoo!

v03_pipeline/lib/misc/io.py
bpblanken: yeah... they're a best attempt rn (I'm pretty sure this type isn't even fully correct).

the issues with `mypy` were mostly with `luigi` and `hail`.  we have way too many cases where functions return something like `luigi.Target(x).exists()` or `hl.eval(hl.is_defined(ht.a))` where we know the type is `bool`, but neither library has the python typing integrations to make something like `mypy` work correctly.  There's definitely a potential path towards using the `mypy` non-strict mode and typing things as `Any`... I just opted against it because it's potentially even more confusing to have "bad" type checking rather than "no" type checking.

PR #807, opened 2024-06-12
Remove concept of private crdqs

bpblanken:


jklugherz (approved):
they were always public? :)

PR #809, opened 2024-06-12
Parameter refactoring + cleanup.

bpblanken:
List of changes:
- removed `callset_paths` in favor of `callset_path`.
- removed `liftover_ref_path` param in favor of an env var.
- `imputed_sex_path` and `filters_path` are computed and enabled via env vars (and optionally disabled via a "skip" param).
- Make use of [luigi.util.inherits](https://luigi.readthedocs.io/en/stable/api/luigi.util.html) to reduce parameter copying.


v03_pipeline/lib/tasks/base/base_loading_run_params.py
```diff
@@ -0,0 +1,36 @@
...
+    callset_path = luigi.Parameter()
+    ignore_missing_samples_when_remapping = luigi.BoolParameter(
+        default=False,
+        parsing=luigi.BoolParameter.EXPLICIT_PARSING,
+    )
+    force = luigi.BoolParameter(
+        default=False,
+        parsing=luigi.BoolParameter.EXPLICIT_PARSING,
+    )
+    skip_check_sex_and_relatedness = luigi.BoolParameter(
```
bpblanken, line 21: Some context on this:

- I wanted options that defaulted to `False` here, so that we didn't need to pass a flag for our default case in airflow (which we are doing now for s&r, which has led to some confusion). 
- I wanted to limit these to options that would either be manually specified while debugging a run or would be enabled dependent on sample source.
- `skip_` prefix for consistency.  
- There are _also_ env vars for [the two features](https://github.com/broadinstitute/seqr-loading-pipelines/pull/809/files#diff-d715bb96893f31a9b3249e1c50b7eea3e3311ea34c07e8f719010dfccda03499R31-R32) that we don't want to enable for local users, so we wouldn't have to hardcode  `skip_expect_filters=True` and `skip_check_sex_and_relatedness=True` (though we could).  
- The "env var only" approach doesn't work well because env vars are set at the dataproc cluster level and I haven't been able to get job env vars to override cluster env vars correctly (one afternoon already wasted on this a few months ago).

v03_pipeline/lib/paths.py
```diff
@@ -198,6 +216,24 @@ def sex_check_table_path(
...
 
+def valid_filters_path(
+    dataset_type: DatasetType,
+    sample_type: SampleType,
+    callset_path: str,
+) -> str | None:
+    if (
+        not Env.EXPECT_WES_FILTERS
+        or not dataset_type.expect_filters(sample_type)
+        or 'part_one_outputs' not in callset_path
```
jklugherz, line 227: what are 'part_one_outputs' and 'part_two_outputs'?

bpblanken: @matren395 can give some context here!  I think it's just that internal exome callsets have always been delivered in two different folders, and we run an extra step to join the `filters` part onto the first part!

matren395: That's what I'd believe this is too, yes!

WES callsets are delivered in two parts usually, with two folders for Part_One_Outputs and Part_Two_Outputs. Part_One_Outputs are the "normal" sharded VCFs, one set of sharded VCFs per chromosome, complete with tons of annotations and information. Part_Two_Outputs is I believe just one VCF per chromosome, containing the sites to filter the Part_One VCFs to .

v03_pipeline/lib/tasks/base/base_loading_run_params.py
```diff
@@ -0,0 +1,36 @@
...
+        default=False,
+        parsing=luigi.BoolParameter.EXPLICIT_PARSING,
+    )
+    skip_validation = luigi.BoolParameter(
+        default=False,
+        parsing=luigi.BoolParameter.EXPLICIT_PARSING,
+    )
+    is_new_gcnv_joint_call = luigi.BoolParameter(
+        default=False,
+        description='Is this a fully joint-called callset.',
```
jklugherz, line 35: do we also need `parsing=luigi.BoolParameter.EXPLICIT_PARSING` here?

v03_pipeline/lib/tasks/update_lookup_table.py
```diff
@@ -73,51 +47,34 @@ def complete(self) -> bool:
...
                 project_pedigree_path,
-                imputed_sex_path,
-            ) in callset_project_pairs(
-                self.callset_paths,
+            ) in zip(
                 self.project_guids,
                 self.project_remap_paths,
                 self.project_pedigree_paths,
-                self.imputed_sex_paths,
+                strict=False,
```
jklugherz, line 65: `strict=False` [makes](https://docs.python.org/3/library/functions.html#zip) it so that if any of the 3 lists are of different length we'll never know - do we expect these to ever be of different lengths?

jklugherz (approved):

v03_pipeline/lib/tasks/update_lookup_table.py
bpblanken: yeah! I think this is worth changing.  `ruff` complains if you `zip` without the param... and I do think it's impossible for the lists to not be the same length with the way it's been simplified now....but throwing an error is definitely clearer!

PR #829, opened 2024-07-03
Benb/split import and validate

bpblanken:
Splits the Import and Validate steps into two disjoint tasks.

### DIFF
--- a/v03_pipeline/lib/tasks/validate_callset.py
+++ b/v03_pipeline/lib/tasks/validate_callset.py
@@ -1,0 +1,162 @@
+import hail as hl
+import luigi
+import luigi.util
+
+from v03_pipeline.lib.misc.validation import (
+    validate_allele_type,
+    validate_expected_contig_frequency,
+    validate_imported_field_types,
+    validate_imputed_sex_ploidy,
+    validate_no_duplicate_variants,
+    validate_sample_type,
+)
+from v03_pipeline.lib.model import CachedReferenceDatasetQuery
+from v03_pipeline.lib.model.environment import Env
+from v03_pipeline.lib.paths import (
+    cached_reference_dataset_query_path,
+    imported_callset_path,
+    sex_check_table_path,
+)
+from v03_pipeline.lib.tasks.base.base_loading_run_params import BaseLoadingRunParams
+from v03_pipeline.lib.tasks.base.base_update import BaseUpdateTask
+from v03_pipeline.lib.tasks.files import CallsetTask, GCSorLocalTarget, HailTableTask
+from v03_pipeline.lib.tasks.reference_data.updated_cached_reference_dataset_query import (
+    UpdatedCachedReferenceDatasetQuery,
+)
+from v03_pipeline.lib.tasks.write_imported_callset import WriteImportedCallsetTask
+from v03_pipeline.lib.tasks.write_sex_check_table import WriteSexCheckTableTask
+
+
+@luigi.util.inherits(BaseLoadingRunParams)
+class ValidateCallsetTask(BaseUpdateTask):
+    def complete(self) -> luigi.Target:
+        if not self.force and super().complete():
+            mt = hl.read_matrix_table(self.output().path)
+            return hasattr(mt, 'validated_sample_type') and hl.eval(
+                self.sample_type.value == mt.validated_sample_type,
+            )
+        return False
+
+    def output(self) -> luigi.Target:
+        return GCSorLocalTarget(
+            imported_callset_path(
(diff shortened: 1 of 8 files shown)


v03_pipeline/lib/misc/allele_registry.py
```diff
@@ -159,16 +159,20 @@ def handle_api_response(
...
+            allele_info = next(
+                record
+                for record in allele_response['genomicAlleles']
+                if record['referenceGenome'] == reference_genome.value
+            )
+            chrom = allele_info['chromosome']
+            pos = allele_info['coordinates'][0]['end']
+            ref = allele_info['coordinates'][0]['referenceAllele']
+            alt = allele_info['coordinates'][0]['allele']
+        except (KeyError, StopIteration):
```
jklugherz, line 173: oof so `referenceGenome` was missing from the response? 

Looking back to https://reg.clinicalgenome.org/doc/AlleleRegistry_1.01.xx_api_v1.pdf, looks like `referenceGenome` in `alleleDefinition` is populated conditionally. 

We could handle it more explicitly but I think a simply try/except is the better option.

v03_pipeline/lib/tasks/validate_callset.py
```diff
@@ -0,0 +1,149 @@
...
+from v03_pipeline.lib.tasks.files import CallsetTask, GCSorLocalTarget, HailTableTask
+from v03_pipeline.lib.tasks.reference_data.updated_cached_reference_dataset_query import (
+    UpdatedCachedReferenceDatasetQuery,
+)
+from v03_pipeline.lib.tasks.write_imported_callset import WriteImportedCallsetTask
+from v03_pipeline.lib.tasks.write_sex_check_table import WriteSexCheckTableTask
+
+
+@luigi.util.inherits(BaseLoadingRunParams)
+class ValidateCallsetTask(BaseUpdateTask):
```
jklugherz, line 32: is there anything in here that wasn't pulled from `WriteImportedCallsetTask`?

v03_pipeline/lib/misc/allele_registry.py
bpblanken: 🤷 yes!  I think this is the laziest answer... if any keys are missing just let it be unmappable.  manual intervention on this is expensive (I think?)

v03_pipeline/lib/tasks/validate_callset.py
bpblanken: nope!

bpblanken: lifted line by line.

v03_pipeline/lib/tasks/write_imported_callset.py
```diff
@@ -146,64 +100,19 @@ def create_table(self) -> hl.MatrixTable:
...
             self.dataset_type,
-            self.additional_row_fields(mt),
-        )
-        # This validation isn't override-able.  If a field is the wrong
-        # type, the pipeline will likely hard-fail downstream.
-        validate_imported_field_types(
-            mt,
-            self.dataset_type,
-            self.additional_row_fields(mt),
+            additional_row_fields(
```
jklugherz, line 103: Can you delete `additional_row_fields` in this file since you moved it?

v03_pipeline/lib/tasks/write_relatedness_check_table_test.py
```diff
@@ -57,7 +57,7 @@ def setUp(self) -> None:
 
         # Force imported callset to be complete
         ht = import_vcf(TEST_VCF, ReferenceGenome.GRCh38)
-        ht = ht.annotate_globals(sample_type=SampleType.WGS.value)
+        ht = ht.annotate_globals(validated_sample_type=SampleType.WGS.value)
```
jklugherz, line 60: is `validated_sample_type` used anywhere?

v03_pipeline/lib/tasks/write_imported_callset.py
bpblanken: yes!  that snuck through good catch!

v03_pipeline/lib/tasks/write_relatedness_check_table_test.py
bpblanken: yes!  we check it in the `complete` method of `ValidateCallset`.

bpblanken: The change to this block was necessary because we're "mocking" the upstream task completion.

jklugherz (approved):

PR #830, opened 2024-07-08
use cached clinvar submission table based on etag

jklugherz:
- HEAD the clinvar submission summary to get the etag
- if gs://seqr-scratch-temp/clinvar-submissions-{etag}.ht exists, use that
- if not, follow existing flow of getting submissions from ftp, adding an additional step to save imported submissions to gs://seqr-scratch-temp/clinvar-submissions-{etag}.ht

### DIFF
--- a/v03_pipeline/lib/paths.py
+++ b/v03_pipeline/lib/paths.py
@@ -284,8 +284,8 @@ def new_variants_table_path(
     )
 
 
-def clinvar_dataset_path(etag: str) -> str:
+def clinvar_dataset_path(reference_genome: ReferenceGenome, etag: str) -> str:
     return os.path.join(
         Env.HAIL_TMPDIR,
-        f'clinvar-{etag}.ht',
+        f'clinvar-{reference_genome.value}-{etag}.ht',
     )
--- a/v03_pipeline/lib/paths.py
+++ b/v03_pipeline/lib/paths.py
@@ -284,8 +284,8 @@ def new_variants_table_path(
     )
 
 
-def clinvar_submission_summary_path(etag: str) -> str:
+def clinvar_dataset_path(etag: str) -> str:
     return os.path.join(
         Env.HAIL_TMPDIR,
-        f'clinvar-submissions-{etag}.ht',
+        f'clinvar-{etag}.ht',
     )
--- a/v03_pipeline/lib/paths.py
+++ b/v03_pipeline/lib/paths.py
@@ -282,3 +282,10 @@ def new_variants_table_path(
         run_id,
         'new_variants.ht',
     )
+
+
+def clinvar_submission_summary_path(etag: str) -> str:
+    return os.path.join(
+        Env.HAIL_TMPDIR,
+        f'clinvar-{etag}.ht',
+    )
--- a/v03_pipeline/lib/paths.py
+++ b/v03_pipeline/lib/paths.py
@@ -287,5 +287,5 @@ def new_variants_table_path(
 def clinvar_submission_summary_path(etag: str) -> str:
     return os.path.join(
         Env.HAIL_TMPDIR,
-        f'clinvar-{etag}.ht',
+        f'clinvar-submissions-{etag}.ht',
     )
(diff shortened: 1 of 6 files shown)


v03_pipeline/lib/paths.py
```diff
@@ -282,3 +282,10 @@ def new_variants_table_path(
         run_id,
         'new_variants.ht',
     )
+
+
+def clinvar_submission_summary_path(etag: str) -> str:
```
bpblanken, line 287: Cool!  I like this, but instead of just the submission summary can we also do this for the vcf itself?  

We'd need to deal with the `ftp://` protocol that's embedded in `config.py` in a potentially janky way, but that's the real `ht` (the clinvar vcf annotated with submission summary) we want!

v03_pipeline/lib/tasks/reference_data/updated_reference_dataset_collection_test.py
```diff
@@ -221,7 +221,7 @@ def test_update_task_with_empty_reference_data_table(
                     paths=hl.Struct(
                         primate_ai='gs://seqr-reference-data/GRCh38/primate_ai/PrimateAI_scores_v0.2.liftover_grch38.ht',
                         cadd='gs://seqr-reference-data/GRCh38/CADD/CADD_snvs_and_indels.v1.6.ht',
-                        clinvar='ftp://ftp.ncbi.nlm.nih.gov/pub/clinvar/vcf_GRCh38/clinvar.vcf.gz',
+                        clinvar='{protocol}://ftp.ncbi.nlm.nih.gov/pub/clinvar/vcf_GRCh38/clinvar.vcf.gz',
```
jklugherz, line 224: @bpblanken By changing the clinvar source path to include a template, the clinvar path in the rdc ht globals changes too. I don't think this will have negative consequences, it could just be confusing. There's other ways to handle replacing 'ftp' with 'https' to get the etag, lmk if you think I should do something else

bpblanken: one silly option is that we just use `https` in the download.  it's... slightly slower I think?  but maybe easier to reason about.

v03_pipeline/lib/reference_data/clinvar.py
```diff
@@ -110,12 +112,25 @@ def parsed_and_mapped_clnsigconf(ht: hl.Table):
...
 
+def get_clinvar_ht(
+    clinvar_url: str,
+    reference_genome: ReferenceGenome,
+):
+    etag = requests.head(clinvar_url, timeout=10).headers.get('ETag').strip('"')
+    try:
+        logger.info(f'Try using cached clinvar ht with etag {etag}')
+        ht = hl.read_table(clinvar_dataset_path(reference_genome, etag))
+    except hl.utils.FatalError:
```
bpblanken, line 123: I think it's probably cleaner to use [hailtop.fs.exists](https://hail.is/docs/0.2/fs_api.html#hailtop.fs.exists) here.  The `FatalError` stands out to m!

v03_pipeline/lib/reference_data/clinvar.py
```diff
@@ -110,12 +112,25 @@ def parsed_and_mapped_clnsigconf(ht: hl.Table):
...
+        logger.info('Cached clinvar ht not found, downloading latest clinvar vcf')
+        ht = download_and_import_latest_clinvar_vcf(clinvar_url, reference_genome)
+        ht.write(clinvar_dataset_path(reference_genome, etag), overwrite=True)
```
bpblanken, line 126: I think you can use our helper `write` from `io.py` here and return the result from that method (not a huge deal though!)

jklugherz: I chose not to use `write` for the submission write because I wanted to avoid the `repartition` but now since we don't need the entire clinvar ht to remain 2000 partitions I can use `write`!

bpblanken: cool.  I guess, are we concerned that it'll be only one partition when we go to join the reference data though?  not the worst thing to add an arg to `write` that overrides the default partitioning.

jklugherz: ah yes I added an arg to `write`. There's the concern that the join will break if the number of partitions on the clinvar table changes, because we know 2000 works and hail is hail.

bpblanken (approved):

PR #834, opened 2024-07-09
Benb/migration task

bpblanken:
I went back and forth on a few things on this late last week, but it's ready now!


v03_pipeline/lib/migration/misc_test.py
```diff
@@ -0,0 +1,44 @@
...
+
+    def tearDown(self):
+        if os.path.isdir(self.tmpdir.name):
+            shutil.rmtree(self.tmpdir.name)
+
+    def test_list_migrations(self):
+        self.assertEqual(
+            list_migrations(self.tmpdir.name),
+            [
+                ('0000_migration', ANY),
```
jklugherz, line 38: what's the second item in the tuple returned by list_migrations? is it possible to assert on something more specific here

bpblanken: Yeah, it's an implementation of the `BaseMigration` class.  I can make it more specific for sure!

v03_pipeline/lib/tasks/base/base_migrate.py
```diff
@@ -0,0 +1,49 @@
...
+    migration_name = luigi.Parameter()
+
+    @property
+    def migrations_path(self):
+        raise NotImplementedError
+
+    def requires(self) -> luigi.Task | None:
+        # Require the previous migration
+        defined_migrations = [x[0] for x in list_migrations(self.migrations_path)]
+        for i, migration in enumerate(defined_migrations):
```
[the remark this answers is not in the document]
jklugherz: The MigrateAllVariantAnnotationsTablesTask and MigrateLookupTableTask tasks create implementations of BaseMigrateTask for each migration in the migrations directories - and then each of those tasks recursively creates migration task requirements here. Is that a problem, or does luigi cache/(memoize?) the result of identical tasks?

v03_pipeline/lib/migration/misc_test.py
bpblanken: @jklugherz I have no idea if this was the best fix but it is fixed!

jklugherz (approved):

PR #840, opened 2024-07-19
Checkpoint no longer necessary with improved caching

bpblanken:


v03_pipeline/lib/reference_data/dataset_table_operations.py
```diff
@@ -36,10 +35,7 @@ def update_or_create_joined_ht(
             continue
 
         # Join the new one!
-        hl._set_flags(use_new_shuffle=None, no_whole_stage_codegen='1')  # noqa: SLF001
```
jklugherz, line 39: Were you able to find a different solution to the shuffle problem?

bpblanken: eh... I think it's wrong to remove it actually, good call!  Instead I should have moved it to the clinvar file. 🤦

jklugherz (approved):

PR #842, opened 2024-07-22
add sample_type to family and project table paths

jklugherz:
This will require a one-time migration to move all existing family tables into their sampletype subfolder. This won't be too bad as sample_type is already in the table globals.


v03_pipeline/lib/tasks/delete_family_tables.py
```diff
@@ -1,10 +1,12 @@
 import luigi
 
+from v03_pipeline.lib.model import SampleType
 from v03_pipeline.lib.tasks.base.base_hail_table import BaseHailTableTask
 from v03_pipeline.lib.tasks.delete_family_table import DeleteFamilyTableTask
 
 
 class DeleteFamilyTablesTask(BaseHailTableTask):
+    sample_type = luigi.EnumParameter(enum=SampleType)
```
jklugherz, line 9: @bpblanken fyi, if we want to delete all family or project tables for a family or project, we have to run the delete dag once with each sample type

v03_pipeline/lib/tasks/delete_project_family_tables.py
```diff
@@ -21,21 +23,24 @@ def complete(self) -> bool:
...
-        family_guids = hl.eval(project_ht.globals.family_guids)
-        for family_guid in family_guids:
-            self.dynamic_delete_family_table_tasks.add(
-                DeleteFamilyTableTask(
-                    reference_genome=self.reference_genome,
-                    dataset_type=self.dataset_type,
-                    family_guid=family_guid,
-                ),
             )
+            if hfs.exists(project_ht_path):
```
jklugherz, line 33: @bpblanken I think what was happening before was that `HailTableTask` was actually never `complete()` if the project table didn't exist and that's what was causing the infinite loop?

yielding inside the sample_type loop works now if the project table exists.

bpblanken: 🤔 definitely strange behavior!  I think we do actually want this task to fail though if the project able isn't there right?

jklugherz: I don't think so, all of the existing seqr projects will have project tables in only one sample_type directory.

v03_pipeline/lib/tasks/delete_family_tables.py
```diff
@@ -18,12 +19,14 @@ def complete(self) -> bool:
...
     def run(self):
-        for family_guid in self.family_guids:
-            self.dynamic_delete_family_table_tasks.add(
-                DeleteFamilyTableTask(
-                    reference_genome=self.reference_genome,
-                    dataset_type=self.dataset_type,
-                    family_guid=family_guid,
-                ),
-            )
+        for sample_type in SampleType:
```
hanars, line 22: potentially out of scope here, but I think we would actually want to be explicit about the sample type (or multiple sample types) we are deleting, so potentially we could delete one but not the other

v03_pipeline/lib/tasks/delete_project_family_tables.py
```diff
@@ -21,21 +23,24 @@ def complete(self) -> bool:
         )
 
     def run(self):
-        project_table_task: luigi.Target = yield HailTableTask(
-            project_table_path(
+        for sample_type in SampleType:
```
hanars, line 26: same as above, I think ideally we would want to explicitly set which sample type(s) we are deleting

jklugherz:
TODO: change path from v03 to 3.1

v03_pipeline/lib/tasks/delete_project_family_tables.py
jklugherz: TODO: raise error if there is not at least 1 project table! else the task will never complete!

v03_pipeline/lib/tasks/delete_family_tables.py
jklugherz: talked yesterday, this was intentional so that we won't have to remember to run the dag twice for both sample types

v03_pipeline/lib/tasks/delete_project_family_tables.py
jklugherz: ^ done

v03_pipeline/lib/model/definitions.py
```diff
@@ -17,6 +17,7 @@ class Sex(Enum):
 class PipelineVersion(Enum):
     V02 = 'v02'
     V03 = 'v03'
+    V3_1 = 'v3.1'
```
bpblanken, line 20: 😆

bpblanken (approved):
🚀

PR #845, opened 2024-07-24
Add mito local constraint

bpblanken:


v03_pipeline/lib/reference_data/config.py
```diff
@@ -533,4 +536,14 @@ def custom_mpc_select(ht):
             'custom_import': import_locus_intervals,
         },
     },
+    'local_constraint_mito': {
```
bpblanken, line 539: I've left `high_constraint_region_mito` and can remove in a follow up so that we can have a clean handover with a change of the naming.

hanars (approved):

v03_pipeline/lib/reference_data/config.py
bpblanken: Ticket here: https://github.com/broadinstitute/seqr-loading-pipelines/issues/846

PR #856, opened 2024-07-30
Resolve the assumption in the pipeline that remap/pedigree files are immutable.

bpblanken:


v03_pipeline/lib/tasks/update_project_table.py
```diff
@@ -28,7 +29,13 @@ def complete(self) -> bool:
             and super().complete()
             and hl.eval(
                 hl.read_table(self.output().path).updates.contains(
-                    self.callset_path,
+                    hl.Struct(
```
bpblanken, line 32: Another possible approach here is to store the set of families loaded for each callset on the globals.  This works, but it way more storage and annoyance.

bpblanken: We also cannot just check that the families in the pedigree are already in the project table because they may have been loaded from a previous callset!

hanars: I like the hash approach you did, its functionally checking the same thing as storing the full families/samples list but easier to maintain

v03_pipeline/lib/tasks/write_remapped_and_subsetted_callset.py
```diff
@@ -36,7 +37,17 @@ class WriteRemappedAndSubsettedCallsetTask(BaseWriteTask):
     project_pedigree_path = luigi.Parameter()
 
     def complete(self) -> luigi.Target:
-        return not self.force and super().complete()
+        return (
+            not self.force
+            and super().complete()
+            and hl.eval(
+                hl.read_matrix_table(self.output().path).globals.remap_pedigree_hash
```
hanars, line 44: did you add a test for when this is not the case?

bpblanken: I think it might be covered by something that exists, but I should add an explicit one for sure!

v03_pipeline/lib/misc/io.py
```diff
@@ -200,6 +202,17 @@ def import_pedigree(pedigree_path: str) -> hl.Table:
...
 
+def remap_pedigree_hash(remap_path: str, pedigree_path: str) -> hl.Int32Expression:
+    sha256 = hashlib.sha256()
+    if hfs.exists(remap_path):
+        with hfs.open(remap_path) as f1:
+            sha256.update(f1.read().encode('utf8'))
+    with hfs.open(pedigree_path) as f2:
+        sha256.update(f2.read().encode('utf8'))
+    # maximum 4 byte int
+    return hl.int32(int(sha256.hexdigest()[:8], 16))
```
jklugherz, line 213: What's the reasoning behind converting the hash value string to an int?

v03_pipeline/migrations/annotations/0002_add_remap_pedigree_hash.py
```diff
@@ -0,0 +1,28 @@
...
+                (ReferenceGenome.GRCh38, DatasetType.SNV_INDEL),
+                (ReferenceGenome.GRCh38, DatasetType.MITO),
+                (ReferenceGenome.GRCh38, DatasetType.GCNV),
+                (ReferenceGenome.GRCh38, DatasetType.SV),
+            ),
+        )
+
+    @staticmethod
+    def migrate(ht: hl.Table) -> hl.Table:
+        return ht.annotate_globals(
```
jklugherz, line 24: Do you need to add this migration to the `migrations` globals? 
I don't see it in the BaseMigration but that behavior could belong there.

v03_pipeline/lib/misc/io.py
bpblanken: Compression!  Ints are smaller.  We don't really need to use sha256 here since we're only taking 32 bits, but that's the func we're using elsewhere so I stuck with it.

v03_pipeline/migrations/annotations/0002_add_remap_pedigree_hash.py
bpblanken: It's in [BaseMigrateTask](https://github.com/broadinstitute/seqr-loading-pipelines/blob/dev/v03_pipeline/lib/tasks/base/base_migrate.py#L47).  There's a distinction between the `BaseMigration` (which is an interface that the migration definitions implement) and the `BaseMigrateTask`, which is something that a "table" (lookup/annotations or in theory projects) inherit from.

v03_pipeline/lib/misc/io.py
bpblanken: it's definitely a micro optimization though that will have very limited implications.  It wasn't hard though!

jklugherz (approved):

PR #857, opened 2024-07-30
Add new SV annotations for VCF export.

bpblanken:
The actual export will be on a subsequent PR.


v03_pipeline/lib/annotations/sv.py
```diff
@@ -81,6 +81,41 @@ def _sv_types(ht: hl.Table) -> hl.ArrayExpression:
...
+        ALGORITHMS=ht.algorithms,
+        END=ht.start_locus.position,
+        CHR2=ht.end_locus.contig,
+        END2=ht.end_locus.position,
+        SVTYPE=hl.array(SV_TYPES)[ht.sv_type_id],
+        SVLEN=ht.sv_len,
+    )
+
+
+def locus(ht: hl.Table, **_: Any) -> hl.LocusExpression:
```
hanars, line 115: Is the plan to keep both `locus` and `start_locus`, or to rename the field?

bpblanken: Keep only `start_locus` for our annotations table, then export to `locus`, `alleles`, `info`

hanars (approved):

PR #867, opened 2024-08-08
lookup sample_type migration

jklugherz:
companion PR to handle new lookup table structure inside the pipeline: https://github.com/broadinstitute/seqr-loading-pipelines/pull/868


v03_pipeline/lib/migration/base_migration.py
```diff
@@ -12,5 +12,9 @@ class BaseMigration(ABC):
 
     @staticmethod
     @abstractmethod
-    def migrate(ht: hl.Table) -> hl.Table:
+    def migrate(
```
bpblanken, line 15: 😮

v03_pipeline/migrations/lookup/0003_add_sample_type.py
```diff
@@ -0,0 +1,106 @@
...
+                    project_guid,
+                )
+                if not hfs.exists(project_ht_path):
+                    continue
+
+                project_has_ht = True
+                key = (project_guid, sample_type.value)
+                project_guids[i] = key
+                project_families[key] = project_families.pop(project_guid)
+                for update in updates:
```
jklugherz, line 79: I'm not sure if we actually need `sample_type` in updates

v03_pipeline/migrations/lookup/0003_add_sample_type.py
```diff
@@ -0,0 +1,93 @@
...
+        Old Global fields:
+            'project_guids': array<str>
+            'project_families': dict<str, array<str>>
+            'updates': set<struct {
+                callset: str,
+                project_guid: str,
+                remap_pedigree_hash: int32
+            }>
+        New Global fields:
+            'project_guids': array<tuple (
```
hanars, line 42: It might be worth renaming this field to something like `project_sample_types` to make it clearer what it is

v03_pipeline/migrations/lookup/0003_add_sample_type.py
```diff
@@ -0,0 +1,93 @@
...
+                    continue
+
+                project_has_ht = True
+                key = (project_guid, sample_type.value)
+                project_guids[i] = key
+                project_families[key] = project_families.pop(project_guid)
+                break
+
+            # It is possible that there are projects in the lookup globals with no corresponding project ht.
+            # For those projects, set sample_type to missing and log project_guid, so we can delete them later.
```
hanars, line 80: This feels like a case that should be impossible. Would it be better for this to hard fail, and for us to first have to run any project deletions before we can run the migration?

bpblanken: I agree with this, hard fail makes sense and if there's something out of wack we can debug outside of this migration.

jklugherz: so I ran the migration without actually applying it with this check, here are projects that appear in project_guids but have no project tables:
MITO: R0566_tgg_ravenscroft_wes
SNV_INDEL 38: 52224_julia_test
SNV_INDEL 37: 52124_julia_test

They need to be removed from the lookup table.
- However, I don't want to run the delete project dag for 52224_julia_test since it has a family in project_families (['F051011_fam2']) and I don't want that family table to be deleted as well! 
- R0566_tgg_ravenscroft_wes I'm not sure about. It also has families in project_families but the corresponding family tables don't exist.

bpblanken: ah yes.  `R0566_tgg_ravenscroft_wes` is my fault.

bpblanken: We should run the `delete` families task... The family and project table didn't get copied over during the migration.

jklugherz: R0566_tgg_ravenscroft_wes is gone from MITO
52124_julia_test is gone from SNV 37
still waiting on 52224_julia_test to be deleted, got a notebook running now in a cluster

v03_pipeline/lib/migration/base_migration.py
```diff
@@ -12,5 +12,10 @@ class BaseMigration(ABC):
...
+        ht: hl.Table,
+        reference_genome: ReferenceGenome,
```
bpblanken, line 17: This one should also probably be `**kwargs` since it's a base.

bpblanken (approved):
small thing but otherwise good!

PR #875, opened 2024-08-16
some inheritance refactoring

bpblanken:


v03_pipeline/lib/tasks/write_variant_annotations_vcf.py
```diff
@@ -12,7 +11,7 @@
 
 
 @luigi.util.inherits(BaseLoadingRunParams)
-class WriteVariantAnnotationsVCF(BaseHailTableTask):
+class WriteVariantAnnotationsVCF(luigi.Task):
```
bpblanken, line 14: we've been over-using the `BaseHailTableTask` to access the shared params, but this is definitely not writing or updating a hail table.

PR #876, opened 2024-08-19
Local Pipeline `api` and `worker`

bpblanken:


v03_pipeline/api/app.py
```diff
@@ -1,17 +1,80 @@
...
+    except web.HTTPError:
+        logger.exception('HTTPError')
+        raise
+    except Exception as e:
+        logger.exception('Unhandled Exception')
+        error_reason = f'{e}: {traceback.format_exc()}'
+        raise web.HTTPInternalServerError(reason=error_reason) from e
+
+
+async def loading_pipeline_enqueue(request: web.Request) -> web.Response:
```
bpblanken, line 29: The idea here is for this process to accept requests and manage a queue of length one (just a file on the filesystem).  It did not seem feasible for this process itself to run the luigi tasks, since they are likely to take many hours.

v03_pipeline/bin/pipeline_worker.py
```diff
@@ -0,0 +1,89 @@
...
+from v03_pipeline.lib.tasks import (
+    UpdateCachedReferenceDatasetQueries,
+    UpdateVariantAnnotationsTableWithNewSamplesTask,
+    WriteProjectFamilyTablesTask,
+)
+
+logger = get_logger(__name__)
+
+
+def main():
```
bpblanken, line 24: The idea is for this process to run as a sidecar in the same pod, basically looping indefinitely for a request to show up, then blocking on doing the work, then releasing the request once it succeeds or finishes.

v03_pipeline/bin/pipeline_worker.py
```diff
@@ -0,0 +1,89 @@
...
+                        project_guid=lpr.projects_to_run[i],
+                        project_remap_path=project_remap_paths[i],
+                        project_pedigree_path=project_pedigree_paths[i],
+                        force=False,
+                        **task_kwargs,
+                    )
+                    for i in range(len(lpr.projects_to_run))
+                ],
+            ]
+            luigi.build(tasks)
```
bpblanken, line 78: There are two options for scheduling here:
- Run with `--local-scheduler`, which means this process will handle scheduling.  This is what we're doing now in dataproc and works fine.
- We stand up a second process in the pod that's running the luigi central scheduler (literally just the `luigid` command) that will expose the luigi UI.  

My preference is to at least try to get the centralized scheduler working since it appears lightweight and the UI is passable.  It was super easy to get working locally (I went down the rabbit hole for ~an hour or so last week when trying to run tasks inside of `aiohttp` and came up pretty quickly).  The main unknowns there are how to expose the  `luigid` endpoint in `helm` correctly, but I think this is at most a day of futzing.

v03_pipeline/lib/model/cached_reference_dataset_query.py
```diff
@@ -13,7 +13,7 @@
 )
 
 
-class CachedReferenceDatasetQuery(Enum):
+class CachedReferenceDatasetQuery(str, Enum):
```
bpblanken, line 16: `StrEnum` is available in python3.11 only.  It's necessary for the json serialization in the endpoint.

v03_pipeline/bin/pipeline_worker.py
```diff
@@ -0,0 +1,89 @@
...
+                    lpr.sample_type,
+                    project_guid,
+                )
+                for project_guid in lpr.projects_to_run
+            ]
+            task_kwargs = {
+                k: v for k, v in lpr.model_dump().items() if k != 'projects_to_run'
+            }
+            tasks = [
+                UpdateCachedReferenceDatasetQueries(
```
bpblanken, line 53: These are the three tasks necessary to get everything to run correctly.  They do not have any dependency structure between them, though they share upstream dependencies (like `UpdatedReferenceDataset` or `WriteImprotedCallset`).  

The blockers to condensing this to a single task are mostly around how we have split tasks apart in airflow.  We have a `for project_guid in projects_to_run` loop there so that each project task can be viewed in isolation.

v03_pipeline/bin/pipeline_worker.py
hanars: Agreed that if we can get this done in under about a week it would be worth it, having a UI exposed to users so they can see how pipeline runs are progressing would be a very valuable feature

hanars (approved):
Exciting stuff!

v03_pipeline/api/app.py
```diff
@@ -1,17 +1,80 @@
...
+        {'Successfully queued': lpr.model_dump()},
+        status=web_exceptions.HTTPAccepted.status_code,
+    )
 
 
 async def status(_: web.Request) -> web.Response:
     return web.json_response({'success': True})
 
 
 async def init_web_app():
```
jklugherz, line 68: Why use an async web server here as opposed to a normal synchronous one?

bpblanken: Good question!  This is basically a consistency choice.  This web server comes embedded in the hail docker image and is what the hail search service uses.  It is however, overkill.  Especially if it requires aiofiles to touch the filesystem 🤷

PR #881, opened 2024-08-29
Wrap import methods in validation

bpblanken:
This improves the quality of the validation errors thrown during our first import task, trying to disambiguate hail error thrown by issues with the VCF rather than Hail/Dataproc/bugs in our code.


v03_pipeline/lib/misc/io.py
```diff
@@ -19,6 +23,29 @@
 FEMALE = 'Female'
 
 
+def validated_hl_function(
```
bpblanken, line 26: Idea here is to wrap function with a call to "checkpoint", forcing evaluation.  If one of the assigned regexes matches, we throw a validation error with the paired message.

bpblanken: Additionally, I don't think any of the exceptions added in this PR can contribute to our expected "support for multiple validation errors"...  any of these generally preclude further hail operations.

hanars (approved):
I like this apporach!

PR #882, opened 2024-08-29
add stricter AF prefilter option

hanars:
This expands the AF pre-filter table to include another bucket of variants, with an AF between 0.001 and 0.01. This allows us to more strictly prefilter the entries table on searches (like the AIP search) with stricter AF criteria than locus criteria

I am not sure exactly what would have to be run in order to generate this new table, let me know if I need to write anything else to assist with that


bpblanken:
confirming that we'll want to do this manually.  Looks like our code doesn't handle the 'select' validation for these crdqs because the selects are not nested under a `gnomad_genomes`, which they maybe should be.

bpblanken (approved):

PR #883, opened 2024-08-29
VEP Config update

bpblanken:
Turns out passing `VEP_CONFIG_URI` is going to be very annoying if we're trying to support both reference genomes.  This fixes by:
- Removing the env var.
- Moving the configs to `seqr-reference-data` instead of `seqr-luigi` so that they're not in the "v03_pipeline" package for local users but will be downloaded with the rest of the VEP data.  This pattern becomes shared between our dataproc and on prem users.
- Hardcoding the file paths when calling VEP.


v03_pipeline/lib/vep.py
```diff
@@ -1,13 +1,10 @@
-import hail as hl
+from string import Template
 
-from v03_pipeline.lib.model import DatasetType, Env, ReferenceGenome
+import hail as hl
 
+from v03_pipeline.lib.model import DatasetType, ReferenceGenome
 
-def validate_vep_config_reference_genome(reference_genome) -> None:
```
bpblanken, line 6: No longer necessary since the reference genome is now embedded as part of the path.

hanars (approved):

PR #884, opened 2024-08-29
Raise a validation error if there are SNV_INDEL symbolic alleles

bpblanken:
…DEL symbolic alleles

### DIFF
--- a/v03_pipeline/lib/misc/validation.py
+++ b/v03_pipeline/lib/misc/validation.py
@@ -22,7 +22,8 @@ def validate_allele_type(
         ),
     )
     if ht.count() > 0:
-        msg = f'Alleles with invalid AlleleType are present in the callset: {ht.alleles.take(10)}'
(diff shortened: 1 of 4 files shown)


hanars (approved):

v03_pipeline/lib/misc/validation.py
```diff
@@ -13,14 +13,17 @@ class SeqrValidationError(Exception):
 
 def validate_allele_type(
     mt: hl.MatrixTable,
+    dataset_type: DatasetType,
 ) -> None:
     ht = mt.rows()
     ht = ht.filter(
-        hl.numeric_allele_type(ht.alleles[0], ht.alleles[1])
-        == hl.genetics.allele_type.AlleleType.UNKNOWN,
+        dataset_type.invalid_allele_types.contains(
```
bpblanken, line 20: I fixed this to be more precise in the error message... using a collected set rather than just the first 10 alleles.

PR #896, opened 2024-09-13
use clinvar allele id to map caid to 'unmappable' variants

jklugherz:


v03_pipeline/lib/tasks/write_new_variants_table.py
```diff
@@ -195,12 +195,30 @@ def create_table(self) -> hl.Table:
...
+                key='clinvar_allele_id',
+            )
+            for ar_ht_chunk, clinvar_allele_id_chunk in register_alleles_in_chunks(
                 new_variants_ht,
                 self.reference_genome,
             ):
                 ar_ht = ar_ht.union(ar_ht_chunk)
+                clinvar_allele_ht = clinvar_allele_ht.union(clinvar_allele_id_chunk)
             new_variants_ht = new_variants_ht.join(ar_ht, 'left')
+            new_variants_ht = new_variants_ht.key_by(new_variants_ht.clinvar.alleleId)
```
bpblanken, line 213: Is it possible to do this with a dictionary lookup rather than a `key_by`?  I'm extremely scared of hail key_by because it causes shuffles.

PR #899, opened 2024-09-19
Update pipeline to deliver a validation_errors file, with all validation exceptions.

bpblanken:
Updates the pipeline to run all validations in a loop, throwing an exception will all failures.  Also writes the failures in a json file.


v03_pipeline/lib/misc/validation.py
```diff
@@ -102,7 +106,7 @@ def validate_expected_contig_frequency(
     )
     if missing_contigs:
         msg = 'Missing the following expected contigs:{}'.format(
-            ', '.join(missing_contigs),
+            ', '.join(sorted(missing_contigs)),
```
bpblanken, line 109: Provides consistent output.

v03_pipeline/lib/tasks/write_imported_callset.py
```diff
@@ -62,50 +66,64 @@ def requires(self) -> list[luigi.Task]:
...
-            self.reference_genome,
-            self.dataset_type,
-        )
-        filters_path = None
-        if (
-            Env.EXPECT_WES_FILTERS
-            and not self.skip_expect_filters
-            and self.dataset_type.expect_filters(
-                self.sample_type,
+        if self.clone(WriteValidationErrorsForRunTask).complete():
```
bpblanken, line 69: The control flow is a little wonky... this task isn't really a dependency because it cannot run before we either import or validate the callset, but we do want to avoid re-running (we have retries enabled in airflow) if it it exists.

v03_pipeline/lib/tasks/write_metadata_for_run.py
```diff
@@ -27,9 +27,6 @@ def output(self) -> luigi.Target:
             ),
         )
 
-    def complete(self) -> bool:
```
bpblanken, line 30: this is the inherited "complete" method, so is unnecessary code.

v03_pipeline/lib/tasks/write_imported_callset.py
hanars: doesn't this imply that theres something wrong with our overall control flow? Its a bit unclear to me what the actual issue is, but it feels like if we split our some steps into their own tasks that would fix this issue more explicitly

v03_pipeline/lib/tasks/write_imported_callset.py
```diff
@@ -62,50 +66,64 @@ def requires(self) -> list[luigi.Task]:
...
+            raise SeqrValidationError(
+                self.clone(WriteValidationErrorsForRunTask).to_single_error_message(),
             )
-        ):
-            filters_path = valid_filters_path(
-                self.dataset_type,
-                self.sample_type,
+        try:
```
hanars, line 73: this feels like a very large try/except block given that AFAIK there is only one function call in this block thats capable of generating the exception thats being caught. I think it would be more readable to just wrap that one call in the try/except

bpblanken: I think there are four functions that can throw validation errors (`import_vcf`, `select_relevant_fields` (will throw if a field we're expecting is missing), `validate_imported_field_types`, and `split_multi_hts`).  Only one is explicit though, so a comment is worthwhile?  Or I can at least improve the function names to clarify that they're capable of throwing an exception (that feels a little `Java` to me though).

v03_pipeline/lib/tasks/write_imported_callset.py
bpblanken: I think, in some sense, yes.  This block is effectively preventing _airflow_ from triggering a retry due to a validation error.  We probably don't want that to be on the luigi side and should try to put it into `airflow`.... I think with potentially an additional task that checks for this file first.

bpblanken: I will remove the block from these tasks

v03_pipeline/lib/tasks/write_imported_callset.py
bpblanken: they're currently decorated functions:

```
@validated_hl_function(...)
def import_vcf()
```

renaming to 

```
@validated_hl_function(...)
def import_valid_vcf()
```

and having the function name be misleading.  Or doing:

```
import_valid_vcf = validated_hl_function(import_vcf)
```

seems weird to me.  

[Renaming](https://stackoverflow.com/questions/10874432/possible-to-change-function-name-in-definition) within the decorator feels inappropriate.

bpblanken: I added some helpful comments to these functions to make it more clear.

PR #902, opened 2024-09-19
add task to reload hail backend globals

jklugherz:


v03_pipeline/lib/tasks/trigger_hail_backend_reload.py
```diff
@@ -0,0 +1,36 @@
...
+)
+
+logger = get_logger(__name__)
+
+
+@luigi.util.inherits(BaseLoadingPipelineParams)
+class TriggerHailBackendReload(luigi.Task):
+    run_id = luigi.Parameter()
+
+    def output(self):
```
bpblanken, line 19: I bet we don't even need this, we can just re-run the task cause it's idempotent!

v03_pipeline/lib/tasks/trigger_hail_backend_reload.py
```diff
@@ -0,0 +1,36 @@
...
+                dataset_type=self.dataset_type,
+                run_id=self.run_id,
+            ),
+        )
+
+    def complete(self):
+        return GCSorLocalTarget(self.output().path).exists()
+
+    def run(self):
+        url = f'{Env.HAIL_BACKEND_SERVICE_HOSTNAME}:{Env.HAIL_BACKEND_SERVICE_PORT}/reload_globals'
```
bpblanken, line 32: 👍

v03_pipeline/bin/pipeline_worker.py
```diff
@@ -73,6 +76,14 @@ def main():
                     for i in range(len(lpr.projects_to_run))
                 ],
             ]
+            if Env.SHOULD_TRIGGER_HAIL_BACKEND_RELOAD:
+                tasks.append(
+                    TriggerHailBackendReload(
```
bpblanken, line 81: I think we need something that'll make sure this task runs after all the others.  Maybe `require` the WriteProjectFamilyTablesTask?

bpblanken: Or even better, the annotations table task, since the `/reload` route very specifically has to do with the globals on that table!

v03_pipeline/lib/tasks/trigger_hail_backend_reload.py
jklugherz: that tracks, I added a `done` attribute to the task to track completion

v03_pipeline/bin/pipeline_worker.py
jklugherz: Ok, I made `UpdateVariantAnnotationsTableWithNewSamplesTask` a requirement of the new task. 

I introduced another class `BaseProjectInfoParams` to track the project-related parameters instead of copy pasting them into the new task.

bpblanken: Yep that sounds like the right thing for sure!

v03_pipeline/lib/tasks/base/base_project_info_params.py
```diff
@@ -0,0 +1,11 @@
+import luigi
+import luigi.util
+
+from v03_pipeline.lib.tasks.base.base_loading_run_params import BaseLoadingRunParams
+
+
+@luigi.util.inherits(BaseLoadingRunParams)
+class BaseLoadingRunWithProjectInfoParams(luigi.Task):
```
jklugherz, line 8: @bpblanken how do you feel about this naming...

bpblanken: I think it's ok!  One sneaky thing, we've done work to make the pedigree paths a computable function of `project_guid` so this isn't going to be as necessary in the future.  The right thing for now though!

v03_pipeline/lib/tasks/trigger_hail_backend_reload.py
```diff
@@ -0,0 +1,31 @@
...
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.done = False
+
+    def requires(self):
+        return self.clone(UpdateVariantAnnotationsTableWithNewSamplesTask)
+
+    def run(self):
+        url = f'{Env.HAIL_BACKEND_SERVICE_HOSTNAME}:{Env.HAIL_BACKEND_SERVICE_PORT}/reload_globals'
+        res = requests.post(url, headers={'From': 'loading-pipelines'}, timeout=300)
```
bpblanken, line 26: This is a great!  Maybe `{'From': 'pipeline-runner'}` for consistency with the helm chart

v03_pipeline/lib/paths.py
```diff
@@ -351,3 +351,18 @@ def loading_pipeline_queue_path() -> str:
         'loading_pipeline_queue',
         'request.json',
     )
+
+
+def hail_backend_reload_success_for_run_path(
```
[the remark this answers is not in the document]
jklugherz: true!

v03_pipeline/lib/model/environment.py
```diff
@@ -54,4 +61,5 @@ class Env:
     PROJECT_ID: str | None = PROJECT_ID
     REFERENCE_DATASETS_DIR: str = REFERENCE_DATASETS_DIR
     SHOULD_REGISTER_ALLELES: bool = SHOULD_REGISTER_ALLELES
+    SHOULD_TRIGGER_HAIL_BACKEND_RELOAD: bool = SHOULD_TRIGGER_HAIL_BACKEND_RELOAD
```
jklugherz, line 64: to confirm, we don't have to set this env var because we're not using this functionality

bpblanken: Yes that is correct!  At least for now!

bpblanken (approved):

PR #904, opened 2024-09-20
Validation refactoring

bpblanken:
I was having unit test trouble on a larger branch, so split out the refactoring portions that will all us to run all validation functions together.


v03_pipeline/lib/misc/validation.py
```diff
@@ -11,9 +24,43 @@ class SeqrValidationError(Exception):
...
+def get_validation_dependencies(
+    dataset_type: DatasetType,
+    reference_genome: ReferenceGenome,
+    callset_path: str,
+    skip_validation: bool,
+    skip_check_sex_and_relatedness: bool,
+    **_: Any,
+) -> dict[str, hl.Table]:
+    deps = {}
+    if skip_validation or not dataset_type.can_run_validation:
```
hanars, line 36: this function is only called once and its in an `if not self.skip_validation and self.dataset_type.can_run_validation:` block so I think this check is redundant

v03_pipeline/lib/misc/validation.py
```diff
@@ -104,8 +154,17 @@ def _validate_field(
 
 def validate_imputed_sex_ploidy(
     mt: hl.MatrixTable,
-    sex_check_ht: hl.Table,
+    dataset_type: DatasetType,
+    skip_check_sex_and_relatedness: bool,
+    sex_check_ht: hl.Table | None = None,  # nb: sex_check_ht will be undefined if
+    **_: Any,
 ) -> None:
+    if not (
```
hanars, line 162: instead of writing the same conditional check here and in `get_validation_dependencies` you could just check here if `sex_check_ht` is None

v03_pipeline/lib/misc/validation.py
bpblanken: 👍 I would love to reduce the number of places we have this check 😬

hanars (approved):

PR #908, opened 2024-09-25
Fix missing sample types in delete dags

bpblanken:


jklugherz (approved):
thanks for identifying and fixing this bug!

bpblanken:
@jklugherz do you feel like this is actually the right approach?

PR #917, opened 2024-10-04
remove version prefix from paths for local users

bpblanken:


jklugherz:
why can't local users have the prefix?

bpblanken:
@jklugherz this is a great question and is maybe a bug... but we've set up everything on the seqr end to read these paths without a prefix.  We're also removing the prefix in the airflow rsync rn.   It was an inertia decision to go in this direction rather than try to change on the read side.

jklugherz (approved):

PR #920, opened 2024-10-07
add task to write success file for run

jklugherz:


v03_pipeline/bin/pipeline_worker.py
```diff
@@ -53,13 +54,16 @@ def main():
...
                 UpdateVariantAnnotationsTableWithNewSamplesTask(
-                    project_guids=lpr.projects_to_run,
-                    project_remap_paths=project_remap_paths,
-                    project_pedigree_paths=project_pedigree_paths,
-                    run_id=run_id,
-                    **task_kwargs,
+                    **loading_run_task_params,
                 ),
                 *[
                     WriteProjectFamilyTablesTask(
```
jklugherz, line 69: there's an argument for making this a luigi task with dynamic dependencies (it would make the task parameter inheritance cleaner, and I could reuse the task inside of the `WriteSuccessFileTask` requires) but that feels silly since this can be accomplished by a simple loop

bpblanken: 👍 Do we even need `WriteProjectFamilyTablesTask` here if it's in the requires of the `WriteSuccessFileTask`?

jklugherz: nope!!

v03_pipeline/lib/tasks/write_success_file.py
```diff
@@ -0,0 +1,37 @@
...
+                WriteProjectFamilyTablesTask,
+                project_guid=self.project_guids[i],
+                project_remap_path=self.project_remap_paths[i],
+                project_pedigree_path=self.project_pedigree_paths[i],
+            )
+            for i in range(len(self.project_guids))
+        ]
+
+    def run(self):
+        with hfs.open(self.output().path, mode='w') as f:
```
bpblanken, line 36: super nitpicky but you can just do:
```
with self.output().open('w') as f:
```

we don't need the hail filesystem for this!

PR #922, opened 2024-10-09
var seqr

bpblanken:
This... probably needs to go in today as I'm having trouble testing the helm side without the right directories here.  There shouldn't be any impact on production as the env vars are overridden for us anyways.


v03_pipeline/var/luigi_config/luigi-defaults.conf
```diff
@@ -1,9 +0,0 @@
-[core]
```
bpblanken, line 1: moving this to helm!

PR #927, opened 2024-10-17
Caid refactor

bpblanken:


v03_pipeline/lib/misc/allele_registry.py
```diff
@@ -106,7 +104,10 @@ def register_alleles(
 
 
 def build_url(base_url: str, reference_genome: ReferenceGenome) -> str:
-    login, password = get_ar_credentials_from_secret_manager()
+    login, password = (
+        Env.CLINGEN_ALLELE_REGISTRY_LOGIN,
```
jklugherz, line 108: are these just going to be configured in plain text in helm?

bpblanken: The `LOGIN` yes, the `PASSWORD` is going to be fetched from the kubernetes secret https://github.com/broadinstitute/seqr-helm/pull/71/files#diff-a43f3d84b733a2ff1c60f1dc01111db8fe8c957fa5c6d097b129147549977819R4.  (we might put the login in the secret too just for sanity... cause that's what we're doing now?)

The idea is to initially move the secret fetch to `airflow` (so it'll be passed in plain text for the next few weeks 🤦 ) and then eventually it'll go away.

bpblanken: The airflow side is here: https://github.com/broadinstitute/seqr-pipeline-airflow/pull/297

jklugherz (approved):

PR #930, opened 2024-10-21
add task to write relatedness check to tsv

jklugherz:


v03_pipeline/lib/tasks/write_remapped_and_subsetted_callset.py
```diff
@@ -69,6 +72,7 @@ def requires(self) -> list[luigi.Task]:
                 *requirements,
                 self.clone(WriteRelatednessCheckTableTask),
                 self.clone(WriteSexCheckTableTask),
+                self.clone(WriteRelatednessCheckTsvTask),
```
bpblanken, line 75: Any reason to not just delete `WriteRelatednessCheckTableTask`?

bpblanken:
Yep this looks great!  one small question but otherwise that's it.

v03_pipeline/lib/tasks/write_remapped_and_subsetted_callset.py
jklugherz: didn't think to but it is definitely the move

PR #932, opened 2024-10-24
Support gcs dirs in rsync

bpblanken:
Supports `REFERENCE_DATASETS_DIR` being set to a `gcs` dir by an external user!


jklugherz:
When would users run this script? Just once, and then they keep their own of reference data up to date?

bpblanken:
@jklugherz It should be run [in the pipeline-runner](https://github.com/broadinstitute/seqr-helm/blob/main/charts/pipeline-runner/values.yaml#L48) in it.  Idea is that a user may use a `REFERENCE_DATASETS_DIR` that is not "ours"

PR #933, opened 2024-10-24
Create dataproc cluster task

bpblanken:
- Adds a `CreateDataprocClusterTask`

#938


v03_pipeline/lib/tasks/dataproc/create_dataproc_cluster.py
```diff
@@ -0,0 +1,168 @@
...
+from v03_pipeline.lib.logger import get_logger
+from v03_pipeline.lib.model import Env, ReferenceGenome
+from v03_pipeline.lib.tasks.base.base_loading_pipeline_params import (
+    BaseLoadingPipelineParams,
+)
+
+CLUSTER_NAME_PREFIX = 'pipeline-runner'
+DEBIAN_IMAGE = '2.1.33-debian11'
+HAIL_VERSION = hl.version().split('-')[0]
+INSTANCE_TYPE = 'n1-highmem-8'
```
jklugherz, line 17: should this be configurable?

v03_pipeline/lib/tasks/dataproc/create_dataproc_cluster.py
```diff
@@ -0,0 +1,168 @@
...
+    # NB: The luigi.dataproc.contrib module was old and bad
+    # so we built our own shim.
+    run_id = luigi.Parameter()
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        # https://cloud.google.com/dataproc/docs/tutorials/python-library-example
+        self.client = dataproc.ClusterControllerClient(
+            client_options={
+                'api_endpoint': f'{Env.GCLOUD_REGION}-dataproc.googleapis.com:443'.format(
```
jklugherz, line 131: do you need the f string and .format here?

v03_pipeline/lib/tasks/dataproc/create_dataproc_cluster_test.py
```diff
@@ -0,0 +1,125 @@
...
+        worker.add(task)
+        worker.run()
+        self.assertTrue(task.complete())
+
+    def test_spinup_cluster_already_exists_failed(
+        self,
+        mock_cluster_controller: Mock,
+    ) -> None:
+        mock_client = mock_cluster_controller.return_value
+        mock_client.get_cluster.return_value = SimpleNamespace(
```
jklugherz, line 34: nice I just looked up `SimpleNamespace` and it seems useful! would a dictionary work here too?

v03_pipeline/lib/tasks/dataproc/create_dataproc_cluster.py
bpblanken: No I do not!  Good find!

Weird that the linter didn't catch it!

bpblanken: it's crazy syntax how it is actually 😵

v03_pipeline/lib/tasks/dataproc/create_dataproc_cluster_test.py
bpblanken: they're attributes on the returned object from the google cloud api... so dictionary "__getitem__" syntax doesn't work.  very annoying that there isn't a clearer way to do this in python

v03_pipeline/lib/tasks/dataproc/create_dataproc_cluster.py
bpblanken: Good question... I left it un-configurable because we don't configure it, and in theory it's duplicated "scaling" config with the number of pre-emptibles.

PR #935, opened 2024-10-25
add relatedness check file path to metadata.json

jklugherz:


bpblanken (approved):
we should probably try to remove the `s` from `callset`, but out of scope for now!

PR #936, opened 2024-10-28
Benb/use metadata as source of family table load

bpblanken:


v03_pipeline/lib/tasks/write_project_family_tables.py
```diff
@@ -26,27 +27,26 @@ def complete(self) -> bool:
...
+        return [
+            self.clone(
+                WriteRemappedAndSubsettedCallsetTask,
+            ),
+            self.clone(
+                UpdateProjectTableTask,
+            ),
+        ]
 
-        # Fetch family guids from pedigree
```
bpblanken, line 38: The specific failure was a family existing in the pedigree, failing a check and thus being excluded from the rest of the pipeline, but being present on the  project table from a different callset.

v03_pipeline/lib/tasks/write_project_family_tables.py
```diff
@@ -26,27 +27,26 @@ def complete(self) -> bool:
...
-        families_guids_in_pedigree = {
-            f.family_guid for f in parse_pedigree_ht_to_families(pedigree_ht)
-        }
-
-        # Intersect them
-        family_guids_to_load = (
-            family_guids_in_project_table & families_guids_in_pedigree
+    def run(self):
+        ht = hl.read_matrix_table(
+            remapped_and_subsetted_callset_path(
```
hanars, line 42: just sanity checking that this works for AnVIL loading where we don't have remaps (or usually subsets)

bpblanken: yeah the same file gets generated, we just run a "subset" that has all families in it.

hanars (approved):

PR #948, opened 2024-11-01
feat: Add ability to run tasks dataproc.

bpblanken:
Note that it is not actually enabled and the feature flag isn't created here.


v03_pipeline/lib/tasks/dataproc/base_run_job_on_dataproc.py
```diff
@@ -0,0 +1,101 @@
...
+            job = self.client.get_job(
+                request={
+                    'project_id': Env.GCLOUD_PROJECT,
+                    'region': Env.GCLOUD_REGION,
+                    'job_id': f'{self.task_name}-{self.run_id}',
+                },
+            )
+        except google.api_core.exceptions.NotFound:
+            return False
+        else:
```
jklugherz, line 56: is this `else` needed here? won't the following lines be executed anyway if no exception is thrown in the `try` block?

v03_pipeline/lib/tasks/dataproc/base_run_job_on_dataproc.py
```diff
@@ -0,0 +1,101 @@
...
+    def complete(self) -> bool:
+        if not self.dataset_type.requires_dataproc:
+            msg = f'{self.dataset_type} should not require a dataproc job'
+            raise RuntimeError(msg)
+        try:
+            job = self.client.get_job(
+                request={
+                    'project_id': Env.GCLOUD_PROJECT,
+                    'region': Env.GCLOUD_REGION,
+                    'job_id': f'{self.task_name}-{self.run_id}',
```
jklugherz, line 51: nit, but you could make job_id an instance attribute `self.job_id = f'{self.task_name}-{self.run_id}'` so that you do it in just 1 spot instead of 4

v03_pipeline/lib/tasks/dataproc/create_dataproc_cluster.py
```diff
@@ -166,11 +175,13 @@ def run(self):
                 'cluster': get_cluster_config(self.reference_genome, self.run_id),
             },
         )
-        while True:
+        wait_s = 0
+        while wait_s < TIMEOUT_S:
```
jklugherz, line 179: do we also want this waiting behavior in `BaseRunJobOnDataprocTask`?

bpblanken: I left it out because I couldn't think of a good timeout... there's going to be wide variability depending on who's using it and how much compute they as for.  Setting an extreme timout (like 48 hours) might be better though.

v03_pipeline/lib/tasks/dataproc/base_run_job_on_dataproc.py
bpblanken: 👍

v03_pipeline/lib/tasks/dataproc/base_run_job_on_dataproc.py
bpblanken: yeah... I was getting [a lint error](https://docs.astral.sh/ruff/rules/try-consider-else/) if I had the return within the `try`.  I don't think the rule properly accounts for the `return` in the `expect` block.  I can lift them out no problem though.

jklugherz (approved):

PR #950, opened 2024-11-04
Dependency reordering so that `ValidateCallsetTask` runs before updating the reference data.

bpblanken:


v03_pipeline/lib/tasks/reference_data/updated_cached_reference_dataset_query.py
```diff
@@ -71,6 +72,14 @@ def requires(self) -> luigi.Task:
                     ],
                 ),
             )
+        # Special nested import to avoid a circular dependency issue
+        # (ValidateCallset -> this file -> UpdatedReferenceDatasetCollection -> ValidateCallset)
+        # The specific CRDQ referenced in ValidateCallset will never reach
+        # this line due to it being a "query_raw_dataset"
+        from v03_pipeline.lib.tasks.reference_data.updated_reference_dataset_collection import (
```
bpblanken, line 79: Calling this out as weird/off.  We have a pseudo-circular dependency in the pipeline, in that `ValidateCallsetTask` depends on an updated validation hail table, which depends on `gnomad` versioning.  pulling the `import` down into the `requires()` fixes the issue and was the cleanest and easiest way forwards.  The _best_ way forward is to split the crdq task into two tasks, one which depends on the `combined` table and one that doesn't.  I left that as out of scope.

v03_pipeline/lib/tasks/reference_data/update_cached_reference_dataset_queries_test.py
```diff
@@ -21,99 +21,85 @@ class UpdateCachedReferenceDatasetQueriesTest(unittest.TestCase):
...
-                    crdq=CachedReferenceDatasetQuery.GNOMAD_QC,
-                ),
-                mock.call(
-                    reference_genome=ReferenceGenome.GRCh38,
-                    dataset_type=DatasetType.SNV_INDEL,
-                    crdq=CachedReferenceDatasetQuery.HIGH_AF_VARIANTS,
-                ),
-            ],
-        )
+        self.assertEqual(mock_crdq_task.call_count, 4)
```
hanars, line 63: it seems like a loss of test coverage to no longer be testing that the tasks are called with the correct arguments

v03_pipeline/lib/tasks/reference_data/updated_cached_reference_dataset_query.py
hanars: might be worth leaving a note in the comment saying that this could be resolved by splitting the crdq task?

PR #955, opened 2024-11-06
add apogee2 score

jklugherz:
- Update mitimpact score annotation to reflect apogee2_score (replacing apogee1 score)


v03_pipeline/lib/reference_data/config.py
```diff
@@ -503,9 +503,9 @@ def custom_mpc_select(ht):
     },
     'mitimpact': {
         '38': {
-            'version': '3.0.7',
-            'path': 'gs://seqr-reference-data/GRCh38/mitochondrial/MitImpact/MitImpact_db_3.0.7.ht',
-            'select': {'score': 'APOGEE_score'},
+            'version': '3.1.3',
+            'path': 'gs://seqr-reference-data/GRCh38/mitochondrial/MitImpact/MitImpact_db_3.1.3.ht',
```
bpblanken, line 507: did we have to change anything in an old import script to generate the ht?

jklugherz: not really, I followed the same procedure of reading the .txt file into a hail table, annotating locus and alleles, and the only difference was that I had to annotate both APOGEE1_score and APOGEE2_score as floats

v03_pipeline/lib/tasks/reference_data/updated_reference_dataset_collection_test.py
```diff
@@ -343,3 +373,70 @@ def test_update_task_with_existing_reference_dataset_collection_table(
...
+            project_remap_paths=[],
+            project_pedigree_paths=[],
+            skip_validation=True,
+            run_id='2',
+        )
+        worker.add(task)
+        worker.run()
+        self.assertTrue(task.complete())
+
+        # Expect additional mitimpact select field to be present in rdc
```
bpblanken, line 424: Did we really not have coverage for an additional field?  Or was it only in `compare_globals_tests`?

More coverage is good though!

v03_pipeline/var/test/reference_data/test_combined_mito_1.ht/README.txt
```diff
@@ -1,3 +1,3 @@
 This folder comprises a Hail (www.hail.is) native Table or MatrixTable.
-  Written with version 0.2.133-4c60fddb171a
-  Created at 2024/11/02 15:10:48
\ No newline at end of file
+  Written with version 0.2.130-bea04d9c79b5
+  Created at 2024/11/06 13:26:14
```
jklugherz, line 3: I altered this table so that it only has the mitimpact annotations with the old values, to be used in updated_reference_dataset_collection_test

v03_pipeline/var/test/reference_data/test_combined_mito_2.ht/README.txt
```diff
@@ -0,0 +1,3 @@
+This folder comprises a Hail (www.hail.is) native Table or MatrixTable.
+  Written with version 0.2.130-bea04d9c79b5
+  Created at 2024/11/06 12:09:16
```
jklugherz, line 3: I made a copy of the original test_combined_mito_1 table (has all mito annotations) and updated the path, version, and added apogee2_score

v03_pipeline/lib/tasks/reference_data/updated_reference_dataset_collection_test.py
jklugherz: true it's in `test_get_datasets_to_update_select_different` lol

v03_pipeline/lib/reference_data/config.py
bpblanken: cool... I guess we should update that script 🤷?  Better to have whatever we did codified even if it's rough.

v03_pipeline/lib/tasks/reference_data/updated_reference_dataset_collection_test.py
jklugherz: I removed this test

PR #957, opened 2024-11-08
reference dataset refactor pt 1

jklugherz:


v03_pipeline/lib/reference_data/reference_dataset.py
```diff
@@ -0,0 +1,115 @@
...
+            (ReferenceDataset.cadd, ReferenceGenome.GRCh37): 'v1.7',
+            (ReferenceDataset.cadd, ReferenceGenome.GRCh38): 'v1.7',
+            (ReferenceDataset.hgmd, ReferenceGenome.GRCh37): 'HGMD_Pro_2023',
+            (ReferenceDataset.hgmd, ReferenceGenome.GRCh38): 'HGMD_Pro_2023',
+        }[self, reference_genome]
+
+    def raw_dataset_path(self, reference_genome: ReferenceGenome):
+        return {
+            (ReferenceDataset.cadd, ReferenceGenome.GRCh37): {
+                'snv': 'https://krishna.gs.washington.edu/download/CADD/v1.7/GRCh37/whole_genome_SNVs.tsv.gz',
```
jklugherz, line 55: Should we be keeping a copy of the raw dataset in our bucket if it already is not there? I would do that once it wouldn't be part of the script. (We currently have the intermediate cadd dataset table in our bucket but not the raw one, and in the new flow we won't be keeping the intermediate version, just the parsed one).

v03_pipeline/lib/tasks/reference_data/base_updated_reference_dataset.py
```diff
@@ -0,0 +1,52 @@
...
+            key_type,
+            key=key_type.fields,
+            globals=hl.Struct(
+                version=hl.missing(hl.tstr),
+                enums=hl.Struct(),
+            ),
+        )
+
+    def update_table(self, ht):
+        ht_version = hl.eval(ht.globals).version
```
jklugherz, line 38: I'm using `BaseUpdateTask` but technically this flow can be different in that if the table doesn't exist, we create it. If the table exists but the version (and we don't even have to read the table we can just read the name) is different, also write it. We're no longer updating an existing table. Maybe this should inherit BaseWriteTask instead.

v03_pipeline/lib/tasks/reference_data/base_updated_reference_dataset.py
```diff
@@ -0,0 +1,52 @@
...
+
+        if ht_version != self.reference_dataset.version:
```
jklugherz, line 40: Also I think at this point it's sufficient to just check the versions instead of all of the globals and selects to determine if we need to reload. We really wouldn't change the path or selects if we weren't also changing the version.

v03_pipeline/lib/tasks/reference_data/base_updated_reference_dataset.py
```diff
@@ -0,0 +1,52 @@
...
+            ht = self.reference_dataset.load_parsed_dataset_func(
+                self.reference_dataset.raw_dataset_path,
+                self.reference_genome,
+            )
+            # selects logic goes here
+            # enum logic goes here
+            ht = ht.annotate_globals(
+                version=self.reference_dataset.version,
```
jklugherz, line 48: In line with https://github.com/broadinstitute/seqr-loading-pipelines/pull/957/files#r1835112508 I think we only need to annotate the version onto the table but that might mess up the globals checking when we join these onto new variants table.

v03_pipeline/lib/reference_data/cadd.py
```diff
@@ -0,0 +1,68 @@
...
+        'f1': 'pos',
+        'f2': 'ref',
+        'f3': 'alt',
+        'f4': 'RawScore',
+        'f5': 'PHRED',
+    }
+    types = {'f0': hl.tstr, 'f1': hl.tint, 'f4': hl.tfloat32, 'f5': hl.tfloat32}
+
+    with tempfile.NamedTemporaryFile(suffix='.tsv.gz', delete=False) as tmp_file:
+        urllib.request.urlretrieve(raw_dataset_path, tmp_file.name)  # noqa: S310
```
jklugherz, line 25: made some tweaks to the og script and I need to test it but the raw files are BIG

v03_pipeline/lib/tasks/reference_data/base_updated_reference_dataset.py
bpblanken: yeah, that's my initial thought.  `BaseUpdateTask` is intended for tables that are stateful and need to be updated, so I think `BaseWriteTask` might be better.

v03_pipeline/lib/reference_data/reference_dataset.py
```diff
@@ -0,0 +1,106 @@
...
+                'indel': 'https://krishna.gs.washington.edu/download/CADD/v1.7/GRCh38/gnomad.genomes.r4.0.indel.tsv.gz',
+            },
+            (ReferenceDataset.cadd, ReferenceGenome.GRCh38): {
+                'snv': 'https://krishna.gs.washington.edu/download/CADD/v1.7/GRCh38/whole_genome_SNVs.tsv.gz',
```
bpblanken, line 57: Are there any other cases where there's multiple files?  I not maybe just pass a list here and `union` upstream?

v03_pipeline/lib/reference_data/reference_dataset.py
jklugherz: talked offline, we should not be keeping mirrors of publicly available datasets

v03_pipeline/lib/reference_data/cadd.py
```diff
@@ -0,0 +1,68 @@
...
+
+
+def load_cadd_ht_from_raw_dataset(
+    raw_dataset_paths: dict[str, str],
+    reference_genome: ReferenceGenome,
+):
+    snvs_ht = import_cadd_table(raw_dataset_paths['snv'], reference_genome)
+    indel_ht = import_cadd_table(raw_dataset_paths['indel'], reference_genome)
+    ht = snvs_ht.union(indel_ht)
+    return ht.naive_coalesce(10000)
```
bpblanken, line 68: Should use `write` in `io.py` here to help with partitioning.

v03_pipeline/lib/tasks/reference_data/base_updated_reference_dataset.py
jklugherz: we are in agreement on this

v03_pipeline/lib/reference_data/reference_dataset.py
jklugherz: I _think_ this is the only case with multiple files

v03_pipeline/lib/reference_data/cadd.py
bpblanken: ooh, actually this isn't writing, so maybe just removing the repartitioning?

v03_pipeline/lib/reference_data/reference_dataset.py
```diff
@@ -0,0 +1,106 @@
...
+            ]
+        return reference_datasets
+
+    @property
+    def access_control(self) -> AccessControl:
+        if self == ReferenceDataset.hgmd:
+            return AccessControl.PRIVATE
+        return AccessControl.PUBLIC
+
+    def version(self, reference_genome: ReferenceGenome):
```
bpblanken, line 42: as discussed, I guess make both of these `v0` or `0` or `0.0.0`?

v03_pipeline/lib/tasks/reference_data/base_updated_reference_dataset.py
```diff
@@ -0,0 +1,41 @@
...
+            ht_version = hl.eval(ht.globals).version
+
+            if ht_version == self.reference_dataset.version:
+                return ht
+
+        ht = self.reference_dataset.load_parsed_dataset_func(
+            self.reference_dataset.raw_dataset_path,
+            self.reference_genome,
+        )
+        # selects logic goes here
```
jklugherz, line 36: I will leave selects and enums out of this PR, those can be worked on in parallel and is already a separate ticket: https://github.com/broadinstitute/seqr/issues/4473

bpblanken: Is it possible to just build the `selects` into the `load_cadd_ht_from_raw_dataset`?

v03_pipeline/lib/paths.py
```diff
@@ -336,7 +336,8 @@ def valid_reference_dataset_path(
             reference_genome,
         ),
         'reference_datasets',
-        f'{reference_dataset.name}_{reference_dataset.version}.ht',
+        f'{reference_dataset.name}',
```
bpblanken, line 339: 👍

v03_pipeline/lib/paths.py
```diff
@@ -57,6 +58,27 @@ def _v03_reference_data_prefix(
...
+) -> str:
+    root = (
+        Env.PRIVATE_REFERENCE_DATASETS_DIR
+        if access_control == AccessControl.PRIVATE
+        else Env.REFERENCE_DATASETS_DIR
+    )
+    if Env.INCLUDE_PIPELINE_VERSION_IN_PREFIX:
+        return os.path.join(
+            root,
+            PipelineVersion.V03.value,
```
bpblanken, line 73: Worth making this v3.1?

v03_pipeline/lib/reference_data/cadd.py
```diff
@@ -0,0 +1,68 @@
...
+            reference_genome=hl.get_reference(ReferenceGenome.GRCh38),
+        )
+        alleles = hl.array([cadd_ht.ref, cadd_ht.alt])
+        cadd_ht = cadd_ht.transmute(locus=locus, alleles=alleles)
+
+        cadd_union_ht = cadd_ht.head(0)
+        contigs = reference_genome.standard_contigs.union(
+            reference_genome.optional_contigs,
+        )
+        for contig_subset in (contigs[:9], contigs[9:]):
```
bpblanken, line 52: this block looks... suspect.  Can we remove it?

jklugherz: yeah my guess is they wanted to split the logic into two chunks but I'll try removing it

jklugherz:
closing, see new branch reference-data-refactor

PR #958, opened 2024-11-12
reference data refactor - for comments

jklugherz:


v03_pipeline/lib/reference_datasets/config.py
```diff
@@ -0,0 +1,34 @@
+from v03_pipeline.lib.model import ReferenceGenome
+from v03_pipeline.lib.reference_data.cadd import load_cadd_ht_from_raw_dataset
+from v03_pipeline.lib.reference_data_OLD.hgmd import download_and_import_hgmd_vcf
+
+CONFIG = {
```
hanars, line 5: I was envisioning this as part of the enum class, so it would share types and constants more clearly, but I could be convinced that it is better to keep it in a different file
```
class ReferenceDataset(str, Enum):
    cadd = 'cadd'
    hgmd = 'hgmd'

    LOAD_DATASET_FUNC = 'load_parsed_dataset_func'
    VERSION = 'version'
    CONFIG = {
        cadd: {
            LOAD_DATASET_FUNC: load_cadd_ht_from_raw_dataset,
            ReferenceGenome.GRCh37: {
                VERSION: '0.0',
            },
        },
    }
```

v03_pipeline/lib/reference_datasets/config.py
```diff
@@ -0,0 +1,34 @@
...
+    'cadd': {
+        'load_parsed_dataset_func': load_cadd_ht_from_raw_dataset,
+        ReferenceGenome.GRCh37: {
+            'version': '0.0',
```
hanars, line 9: is there a reason we are starting at version 0 instead of version 1?

v03_pipeline/lib/reference_datasets/reference_dataset.py
```diff
@@ -0,0 +1,52 @@
...
+        return AccessControl.PUBLIC
+
+    def version(self, reference_genome: ReferenceGenome) -> str:
+        return self._config[reference_genome]['version']
+
+    def raw_dataset_path(self, reference_genome: ReferenceGenome) -> str | list[str]:
+        return self._config[reference_genome]['raw_dataset_path']
+
+    def load_parsed_dataset_func(self) -> callable:
+        return self._config['load_parsed_dataset_func']
```
bpblanken, line 52: what about just making these consistently named, so `cadd.py` defines `get_ht(...)` and we don't have a function in the config?

v03_pipeline/lib/tasks/reference_data/base_updated_reference_dataset.py
```diff
@@ -0,0 +1,42 @@
...
+    def output(self):
+        return GCSorLocalTarget(
+            valid_reference_dataset_path(
+                self.reference_genome,
+                self.reference_dataset,
+            ),
+        )
+
+    def create_table(self):
+        if does_file_exist(self.output().path):
```
bpblanken, line 25: I think we don't need this anymore with `luigi` doing the `output` checking right?

v03_pipeline/lib/reference_datasets/reference_dataset.py
```diff
@@ -0,0 +1,52 @@
...
+        reference_datasets = {
+            (ReferenceGenome.GRCh37, DatasetType.SNV_INDEL): [
+                ReferenceDataset.cadd,
+                ReferenceDataset.hgmd,
+            ],
+            (ReferenceGenome.GRCh38, DatasetType.SNV_INDEL): [
+                ReferenceDataset.cadd,
+                ReferenceDataset.hgmd,
+            ],
+        }.get((reference_genome, dataset_type), [])
```
hanars, line 30: I wonder if it would be clearer to add a `DATASET_TYPES`  to the config and then just iterate through here to get them. That way we do not need to explicitly configure them on/ off in a separate place from where they are configured
```
CONFIG = {
     cadd: {
            ReferenceGenome.GRCh37: {
                VERSION: '0.0',
                DATASET_TYPES: DatasetType.SNV_INDEL,
            },
}
...
def for_reference_genome_dataset_type(cls, reference_genome: ReferenceGenome, dataset_type: DatasetType) -> list['ReferenceDataset']:
    reference_datasets = [
        dataset for dataset, config in CONFIG.items()
        if dataset_type in config.get(reference_genome, {}).get(DATASET_TYPES, [])
    ] 
    ...
```

v03_pipeline/lib/reference_datasets/reference_dataset.py
jklugherz: we still need to make the function an attribute of the enum so that the task can call it, are you suggesting I go back to the dictionary access pattern?

v03_pipeline/lib/reference_datasets/config.py
jklugherz: oh yeah I like it in the same file

v03_pipeline/lib/reference_datasets/config.py
jklugherz: no I have no strong opinion on this.. 1.0 sounds more trustworthy though so let's do that

v03_pipeline/lib/reference_datasets/reference_dataset.py
bpblanken: I was thinking about trying to do something similar to [what we did with the migrations](https://github.com/broadinstitute/seqr-loading-pipelines/blob/main/v03_pipeline/lib/migration/misc.py#L14), so that we could just return `cadd.get_ht` directly from this function.  But maybe that's too much!

hanars: We could also try something like `importlib` to dynamically read in the correct file? Something like
```
module = imortlib.import_module(f'v03_pipeline.lib.reference_datasets.{self}')
return module. get_ht
```

bpblanken: yeah, something like that is what I was thinking.  It's definitely doable and saves a config option.

v03_pipeline/lib/reference_datasets/reference_dataset.py
```diff
@@ -0,0 +1,93 @@
...
+ENUM_SELECT = 'enum_select'
+
+
+class ReferenceDataset(str, Enum):
+    cadd = 'cadd'
+    hgmd = 'hgmd'
+
+    CONFIG = MappingProxyType(
+        {
+            'cadd': {
```
hanars, line 23: I think you can actually directly use `cadd` here as the variable not as the string

v03_pipeline/lib/reference_datasets/reference_dataset.py
```diff
@@ -0,0 +1,93 @@
...
+                        'https://krishna.gs.washington.edu/download/CADD/v1.7/GRCh38/whole_genome_SNVs.tsv.gz',
+                        'https://krishna.gs.washington.edu/download/CADD/v1.7/GRCh37/gnomad.genomes-exomes.r4.0.indel.tsv.gz',
+                    ],
+                },
+            },
+            'hgmd': {
+                LOAD_DATASET_FUNC: download_and_import_hgmd_vcf,
+                ReferenceGenome.GRCh37: {
+                    VERSION: '1.0',
+                    RAW_DATASET_PATH: 'gs://seqr-reference-data/v3.1/GRCh37/reference_datasets/hgmd-v2020.1.vcf.bgz',
```
hanars, line 44: HGMD can never be in the public bucket. This needs to be removed immediately

v03_pipeline/lib/reference_datasets/reference_dataset.py
```diff
@@ -0,0 +1,93 @@
...
+    hgmd = 'hgmd'
+
+    CONFIG = MappingProxyType(
+        {
+            'cadd': {
+                LOAD_DATASET_FUNC: load_cadd_ht_from_raw_dataset,
+                ReferenceGenome.GRCh37: {
+                    VERSION: '1.0',
+                    RAW_DATASET_PATH: [
+                        'https://krishna.gs.washington.edu/download/CADD/v1.7/GRCh37/whole_genome_SNVs.tsv.gz',
```
hanars, line 28: do we have any concerns about updating from CADD 1.6 to 1.7?

v03_pipeline/lib/reference_datasets/reference_dataset.py
```diff
@@ -1,16 +1,57 @@
...
+                        'https://krishna.gs.washington.edu/download/CADD/v1.7/GRCh37/gnomad.genomes-exomes.r4.0.indel.tsv.gz',
+                    ],
+                },
+            },
+            'hgmd': {
+                LOAD_DATASET_FUNC: download_and_import_hgmd_vcf,
+                ReferenceGenome.GRCh37: {
+                    VERSION: '1.0',
+                    RAW_DATASET_PATH: 'gs://seqr-reference-data/v3.1/GRCh37/reference_datasets/hgmd-v2020.1.vcf.bgz',
+                    ENUM_SELECT: HGMD_ENUM_SELECT,
```
bpblanken, line 45: if you import the `hgmd` module directly we can reference this as `hgmd.ENUM_SELECT`~

v03_pipeline/lib/reference_datasets/reference_dataset.py
bpblanken: I don't think this is actually there, it's just a typo!

v03_pipeline/lib/reference_datasets/reference_dataset.py
hanars: Lynn says we are good!

v03_pipeline/lib/reference_datasets/reference_dataset.py
bpblanken: I think we can't reference the enum value until it's created.  It's probably better to lift `CONFIG` off the enum for memory reasons any ways, placing it below `ReferenceDataset` in the module so we can use the enum value.

v03_pipeline/lib/reference_datasets/reference_dataset.py
hanars: 😌

hanars: also I do not think we want to use an HGMD version from 2020, the current config is using 2023

jklugherz: I think copilot autocompleted this path and I didn't check. It's not actually in there

v03_pipeline/lib/reference_datasets/reference_dataset.py
jklugherz: good question, currently checking out the release notes: https://cadd.bihealth.org/static/ReleaseNotes_CADD_v1.7.pdf

v03_pipeline/lib/reference_datasets/reference_dataset.py
```diff
@@ -0,0 +1,93 @@
...
+                },
+                ReferenceGenome.GRCh38: {
+                    VERSION: '1.0',
+                    RAW_DATASET_PATH: 'gs://seqr-reference-data/v3.1/GRCh38/reference_datasets/hgmd-v2020.1.vcf.bgz',
+                    ENUM_SELECT: HGMD_ENUM_SELECT,
```
hanars, line 50: is this actually used anywhere? If not, it might be worth pulling HGMD out of this PR to limit the scope and then add it back in its wown PR that actually implements the enum select functionality

hanars: also, looking at the current config `enum_select` is identical across reference genomes every place it is used so I doo not think we should be nesting it under the build

jklugherz: noted, and I'll leave hgmd out

v03_pipeline/lib/reference_datasets/reference_dataset.py
jklugherz: alright I tried the importlib approach, the only modification I made is that instead of returning a callable I just return the ht

v03_pipeline/lib/reference_datasets/reference_dataset.py
jklugherz: I implemented it this way! I like that it keeps the config for a reference dataset in one place.

v03_pipeline/lib/reference_datasets/reference_dataset.py
jklugherz: CONFIG is now no longer an attribute of the enum and uses the enums as keys. 

@bpblanken did you leave a comment somewhere about namedtuples?

v03_pipeline/lib/reference_datasets/reference_dataset.py
jklugherz: sounds good. I still need to test/run the new cadd script with the 1.7 data sources.

v03_pipeline/lib/reference_datasets/reference_dataset.py
```diff
@@ -0,0 +1,78 @@
...
+        return module.get_ht(raw_dataset_path, reference_genome)
+
+
+CONFIG = {
+    ReferenceDataset.cadd: {
+        DATASET_TYPES: [DatasetType.SNV_INDEL],
+        ReferenceGenome.GRCh37: {
+            VERSION: '1.0',
+            RAW_DATASET_PATH: [
+                'https://krishna.gs.washington.edu/download/CADD/v1.7/GRCh37/whole_genome_SNVs.tsv.gz',
```
bpblanken, line 66: I think 37 and 38 are jumbled here.

v03_pipeline/lib/reference_datasets/reference_dataset.py
```diff
@@ -0,0 +1,78 @@
...
+    ) -> hl.Table:
+        module = importlib.import_module(
+            f'v03_pipeline.lib.reference_datasets.{self.name}',
+        )
+        return module.get_ht(raw_dataset_path, reference_genome)
+
+
+CONFIG = {
+    ReferenceDataset.cadd: {
+        DATASET_TYPES: [DatasetType.SNV_INDEL],
```
bpblanken, line 62: If this is python and not `json`, this should be a `frozenset`.

v03_pipeline/lib/reference_datasets/reference_dataset.py
```diff
@@ -0,0 +1,78 @@
...
+
+    def version(self, reference_genome: ReferenceGenome) -> str:
+        return CONFIG[self][reference_genome][VERSION]
+
+    def raw_dataset_path(self, reference_genome: ReferenceGenome) -> str | list[str]:
+        return CONFIG[self][reference_genome][RAW_DATASET_PATH]
+
+    def get_ht(
+        self,
+        raw_dataset_path: str | list[str],
```
bpblanken, line 51: Why not call `self.raw_dataset_path` inside the function?

jklugherz: so true

v03_pipeline/lib/reference_datasets/cadd.py
```diff
@@ -0,0 +1,58 @@
...
+
+def get_ht(
+    raw_dataset_paths: list[str],
+    reference_genome: ReferenceGenome,
+) -> hl.Table:
+    snv_path, indel_path = raw_dataset_paths
+    snvs_ht = import_cadd_table(snv_path, reference_genome)
+    indel_ht = import_cadd_table(indel_path, reference_genome)
+    ht = snvs_ht.union(indel_ht)
+    return ht.select('PHRED')
```
hanars, line 58: it would probably speed this up to move the select into `import_cadd_table` so you have dropped the other columns before you do the join

hanars (approved):

v03_pipeline/lib/reference_datasets/reference_dataset.py
bpblanken: ah, cause `self.raw_dataset_path` isn't a property and requires an arg.  I fixed it on my branch :/

PR #959, opened 2024-11-12
hgmd, gnomad_exomes, gnomad_genomes, and gnomad_qc 

jklugherz:


v03_pipeline/lib/reference_datasets/gnomad_qc.py
```diff
@@ -0,0 +1,10 @@
+import hail as hl
+
+
+def get_ht(raw_dataset_path: str, *_) -> hl.Table:
+    if raw_dataset_path.endswith('mt'):
+        ht = hl.read_matrix_table(raw_dataset_path).rows()
+    else:
+        ht = hl.read_table(raw_dataset_path)
```
bpblanken, line 8: This line shouldn't run right?  `get_ht` will only be called if the table doesn't exist?

v03_pipeline/lib/reference_datasets/reference_dataset.py
```diff
@@ -47,8 +50,14 @@ def raw_dataset_path(self, reference_genome: ReferenceGenome) -> str | list[str]
         return CONFIG[self][reference_genome][RAW_DATASET_PATH]
 
     def get_ht(self, reference_genome: ReferenceGenome) -> hl.Table:
+        file_name = (
```
bpblanken, line 53: ooof, this is annoying... I wasn't able to come up with something better in the few min I've been thinking about this though.

v03_pipeline/lib/reference_datasets/gnomad.py
```diff
@@ -0,0 +1,61 @@
...
+    return ht.select(**selects)
+
+
+def get_ht(
+    raw_dataset_path: str,
+    reference_genome: ReferenceGenome,
+) -> hl.Table:
+    ht = hl.read_table(raw_dataset_path)
+
+    if reference_genome == ReferenceGenome.GRCh37:
```
bpblanken, line 56: something like:

```
if reference_genome == ReferenceGenome.GRCh37:
  return process_gnomad_v2_ht(ht)
return process_gnomad_v4_ht(ht)
```

might be cleaner!

v03_pipeline/lib/reference_datasets/gnomad_qc.py
jklugherz: In the original config the 37 source is a matrix table and 38 is a hail table

```
    'gnomad_qc': {
        '37': {
            'version': 'v2',
            'custom_import': import_matrix_table,
            # Note: copied from 'gs://gnomad/sample_qc/mt/gnomad.joint.high_callrate_common_biallelic_snps.pruned.mt'
            'source_path': 'gs://seqr-reference-data/gnomad_qc/GRCh37/gnomad.joint.high_callrate_common_biallelic_snps.pruned.mt',
        },
        '38': {
            'version': '4.0',
            'path': 'gs://gcp-public-data--gnomad/release/4.0/pca/gnomad.v4.0.pca_loadings.ht',
        },
```

v03_pipeline/lib/reference_datasets/gnomad.py
```diff
@@ -0,0 +1,61 @@
...
+    selects['FAF_AF'] = hl.float32(ht.faf[ht.globals.faf_index_dict['adj']].faf95)
+    selects['Hemi'] = hl.if_else(
+        ht.locus.in_autosome_or_par(),
+        0,
+        ht.freq[ht.globals.freq_index_dict['XY_adj']].AC,
+    )
+    return ht.select(**selects)
+
+
+def get_ht(
```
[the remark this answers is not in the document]
bpblanken: ah, `*_` is to help with unused variable linter errors.  my thought was we should make `get_ht` on the `ReferenceDataset` class pass through all `kwargs` to these individual `get_ht` funcs, and within them only use a `ReferenceGenome` or `DatasetType` if they're needed.

v03_pipeline/lib/reference_datasets/gnomad_qc.py
bpblanken: ah, ok!

v03_pipeline/lib/reference_datasets/gnomad.py
```diff
@@ -0,0 +1,61 @@
...
+        ht.freq[ht.globals.freq_index_dict['gnomad_male']].AC,
+    )
+    return ht.select(**selects)
+
+
+def process_gnomad_v4_ht(ht: hl.Table) -> hl.Table:
+    selects = {}
+    global_idx = hl.eval(ht.globals.freq_index_dict['adj'])
+    selects['AF'] = hl.float32(ht.freq[global_idx].AF)
+    selects['AN'] = ht.freq[global_idx].AN
```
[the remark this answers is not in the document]
jklugherz: yup they are in both tables

v03_pipeline/lib/reference_datasets/gnomad.py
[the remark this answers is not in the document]
jklugherz: I'm still not quite sure what the solution would look like here, can you add a code snippet?

v03_pipeline/lib/reference_datasets/reference_dataset.py
```diff
@@ -138,5 +148,55 @@ def requires(self) -> ReferenceDataset:
...
+            VERSION: '1.0',
+            RAW_DATASET_PATH: 'gs://gcp-public-data--gnomad/release/2.1.1/ht/genomes/gnomad.genomes.r2.1.1.sites.ht',
+        },
+        ReferenceGenome.GRCh38: {
+            DATASET_TYPES: frozenset([DatasetType.SNV_INDEL]),
+            VERSION: '1.0',
+            RAW_DATASET_PATH: 'gs://gcp-public-data--gnomad/release/4.1/ht/genomes/gnomad.genomes.v4.1.sites.ht',
+        },
+    },
+    ReferenceDataset.gnomad_qc: {
```
hanars, line 189: This looks identical to the `gnomad_genomes` config so it might be good to do what we did for `clinvar_path` below and do 
```
CONFIG[ReferenceDatasetQuery.gnomad_qc] = CONFIG[ReferenceDataset.gnomad_genomes]
```

v03_pipeline/lib/reference_datasets/gnomad.py
```diff
@@ -0,0 +1,56 @@
+import hail as hl
+
+from v03_pipeline.lib.model import ReferenceGenome
+
+
+def select_fields_v2(ht: hl.Table) -> hl.Table:
+    selects = {}
+    global_idx = hl.eval(ht.globals.freq_index_dict['gnomad'])
+    selects['AF'] = hl.float32(ht.freq[global_idx].AF)
```
hanars, line 9: unless you are conditionally modifying what select to run its more readable to directly call the select fields than assign to a dict:
```
ht = ht.select(
    AF=hl.float32(ht.freq[global_idx].AF),
    AN= ht.freq[global_idx].AN,
    ...
```

v03_pipeline/lib/reference_datasets/gnomad.py
hanars: The v2 and v4 code actually seems mostly identical so I think it would be better to run a single shared function for the shared behavior and only conditionally do the few things that are different. The difference seems to mostly be what the key to use in the `freq_index_dict` global and a couple places where you annotate from a different field between them, so I feel like conditional helpers just in those places would be better

v03_pipeline/lib/reference_datasets/hgmd.py
```diff
@@ -0,0 +1,34 @@
...
+        contig_recoding=reference_genome.contig_recoding(),
+    )
+    ht = mt.rows()
+    ht = ht.select(
+        **{
+            'accession': ht.rsid,
+            'class': ht.info.CLASS,
+        },
+    )
+    return ht.transmute(**get_enum_select_fields(ht, ENUMS))
```
hanars, line 34: just a note that my open dbnsfp PR moves this to be called automatically as part of the shared code so we should make sure that whichever gets merged first we correctly update the other

v03_pipeline/lib/reference_datasets/hgmd.py
```diff
@@ -0,0 +1,34 @@
...
+        skip_invalid_loci=True,
+        contig_recoding=reference_genome.contig_recoding(),
+    )
+    ht = mt.rows()
+    ht = ht.select(
+        **{
+            'accession': ht.rsid,
+            'class': ht.info.CLASS,
+        },
+    )
```
hanars, line 33: Its much more readable to use keyword args than a dictionary if the key names are not variable
```suggestion
    ht = ht.select(
        accession=ht.rsid,
        class=ht.info.CLASS,
    )
```

v03_pipeline/lib/reference_datasets/reference_dataset.py
hanars: what if we make this explicitly configured in the CONFIG with like a MODULE_NAME key, so then here we would just have 
```
module_file_name = CONFIG[self].get(MODULE_NAME, self.name)
```

v03_pipeline/lib/reference_datasets/hgmd.py
jklugherz: `class` is python keyword so that's why I used the dictionary syntax

hanars: Ah that makes sense. I think it might still be more readable for accession to be a kwarg but less important

v03_pipeline/lib/reference_datasets/gnomad_qc.py
jklugherz: yup that was a copy/paste typo, and I updated the gnomad_qc script to check `reference_genome` instead of the path suffix, which I agree is a better solution

v03_pipeline/lib/reference_datasets/reference_dataset.py
jklugherz: good idea, I did this with `gnomad_exomes` and `gnomad_genomes`

v03_pipeline/lib/reference_datasets/gnomad.py
bpblanken: yeah, I think we should move away from selecting from a dictionary of selects!

v03_pipeline/lib/reference_datasets/reference_dataset.py
```diff
@@ -138,5 +143,45 @@ def requires(self) -> ReferenceDataset:
...
+        },
+        ReferenceGenome.GRCh38: {
+            DATASET_TYPES: frozenset([DatasetType.SNV_INDEL]),
+            VERSION: '1.0',
+            RAW_DATASET_PATH: 'gs://gcp-public-data--gnomad/release/4.0/pca/gnomad.v4.0.pca_loadings.ht',
+        },
+    },
 }
 CONFIG[ReferenceDatasetQuery.clinvar_path] = CONFIG[ReferenceDataset.clinvar]
+CONFIG[ReferenceDataset.gnomad_genomes] = CONFIG[ReferenceDataset.gnomad_exomes]
```
bpblanken, line 187: hrm, this feels wrong to me.  shouldn't there be different paths and versions for `genomes` and `exomes`?

`genomes/gnomad.genomes.v4.1.sites.ht` vs `exomes/gnomad.exomes.v4.1.sites.ht`?  and they might not be updated synchronously.

jklugherz: yeahhhh this was incorrect

v03_pipeline/lib/reference_datasets/reference_dataset.py
jklugherz: `gnomad_qc` config being identical to gnomad_genomes was a typo when this comment was made so I'll resolve

v03_pipeline/lib/reference_datasets/hgmd.py
jklugherz: I ported your enum changes into my PR so hopefully we won't need to do that unless

v03_pipeline/lib/reference_datasets/gnomad.py
jklugherz: I took out the dictionary selects and added reference genome conditional helpers, lmk if you think it needs any more tweaks for readability

v03_pipeline/lib/reference_datasets/gnomad.py
```diff
@@ -0,0 +1,58 @@
...
+    return 'gnomad' if reference_genome == ReferenceGenome.GRCh37 else 'adj'
+
+
+def af_popmax_expression(
+    ht: hl.Table,
+    reference_genome: ReferenceGenome,
+) -> hl.Expression:
+    if reference_genome == ReferenceGenome.GRCh37:
+        return ht.popmax[ht.globals.popmax_index_dict['gnomad']].AF
+    return ht.grpmax['gnomad'].AF if hasattr(ht.grpmax, 'gnomad') else ht.grpmax.AF
```
hanars, line 16: `hasattr(ht.grpmax, 'gnomad')` seems like something that should either always be true or false for a given file so shouldn't we just use whichever one is true for the latest release?

v03_pipeline/lib/reference_datasets/hgmd.py
```diff
@@ -0,0 +1,30 @@
...
+    'R',
+]
+
+
+def get_ht(raw_dataset_path: str, reference_genome: ReferenceGenome) -> hl.Table:
+    mt = hl.import_vcf(
+        raw_dataset_path,
+        reference_genome=reference_genome.value,
+        force=True,
+        min_partitions=100,
```
hanars, line 20: I haven't been setting an explicit min_partitions and have just need relying on hail to pick a reasonable number. Is there a way we should go about picking a partition number?

v03_pipeline/lib/reference_datasets/hgmd.py
```diff
@@ -0,0 +1,30 @@
+import hail as hl
+
+from v03_pipeline.lib.model import ReferenceGenome
+
+HGMD_CLASSES = [
```
hanars, line 5: If these are only ever used in the config its probably better just to define them there

v03_pipeline/lib/reference_datasets/reference_dataset.py
```diff
@@ -146,5 +161,56 @@ def requires(self) -> ReferenceDataset:
...
+            VERSION: '1.0',
+            RAW_DATASET_PATH: 'gs://gcp-public-data--gnomad/release/2.1.1/ht/genomes/gnomad.genomes.r2.1.1.sites.ht',
+        },
+        ReferenceGenome.GRCh38: {
+            DATASET_TYPES: frozenset([DatasetType.SNV_INDEL]),
+            VERSION: '1.0',
+            RAW_DATASET_PATH: 'gs://gcp-public-data--gnomad/release/4.1/ht/genomes/gnomad.genomes.v4.1.sites.ht',
+        },
+    },
+    ReferenceDataset.gnomad_qc: {
```
hanars, line 203: Shouldn't this be a `ReferenceDatasetQuery` not a `ReferenceDataset`?

jklugherz: I don't think so because its dependency is itself: https://github.com/broadinstitute/seqr-loading-pipelines/blob/6206bdcf20f5ea4519872e964a6713c1dd308466/v03_pipeline/lib/model/cached_reference_dataset_query.py#L31

v03_pipeline/lib/reference_datasets/hgmd.py
jklugherz: Good question, optimal partitioning remains somewhat of a mystery to me, and in this case since we're using `force=True` everything will end up on one partition (I could be wrong) so maybe the move here is to repartition after import.

It looks like there is an optimal partition size of 128 mb, maybe we should start using this more: https://github.com/broadinstitute/seqr-loading-pipelines/blob/main/v03_pipeline/lib/misc/io.py#L72

v03_pipeline/lib/reference_datasets/gnomad.py
jklugherz: I could use`hl.coalesce` here instead

v03_pipeline/lib/reference_datasets/hgmd.py
bpblanken: I think hail will default to the number of partitions in the sharded vcf, which is 1 here (because it's not sharded).  The partitioning will only matter through the import step, as we'll repartition when we call `write` to [write the file](https://github.com/broadinstitute/seqr-loading-pipelines/blob/main/v03_pipeline/lib/misc/io.py#L293).  

But yes, `force=True` forces serial execution, so the `min_partitions` is useless:

```
In [7]: mt = hl.import_vcf('gs://seqr-reference-data-private/GRCh37/HGMD/HGMD_Pro_2023.1_hg19.vcf.gz', force=True, min_partitions=100)
2024-11-14 17:47:43.245 Hail: WARN: You are trying to read gs://seqr-reference-data-private/GRCh37/HGMD/HGMD_Pro_2023.1_hg19.vcf.gz with *ONE* core of parallelism. This will be very slow. If this file is block-gzipped (bgzip-ed), use force_bgz=True instead.

In [8]: mt.n_partitions()
Out[8]: 1
```

v03_pipeline/lib/reference_datasets/gnomad.py
hanars: Its more that `hasattr` should always be the same for every row - for a given row I value may be missing or not but structs always need to have the same columns defined

v03_pipeline/lib/reference_datasets/hgmd.py
hanars: From the hail docs:
```
force ([bool](https://docs.python.org/3.9/library/functions.html#bool)) – If True, load gzipped files serially on one core. This should be used only when absolutely necessary, as processing time will be increased due to lack of parallelism.
```
Is it absolutely necessary here?

hanars: > It looks like there is an optimal partition size of 128 mb, maybe we should start using this more: https://github.com/broadinstitute/seqr-loading-pipelines/blob/main/v03_pipeline/lib/misc/io.py#L72

Is there any reason not to run this on all the tables we generate before we write out the final file?

bpblanken: ^ it's running already on every call to `create_table`.

v03_pipeline/lib/reference_datasets/reference_dataset.py
bpblanken: Yes that's correct that it's still special and would be included with the code as is.  I also think we can fix when we write the code that utilizes `for_reference_genome_dataset_type`.

v03_pipeline/lib/reference_datasets/gnomad.py
jklugherz: Is it okay to run `hasattr(ht.grpmax, 'gnomad')` on every row though? should I run `hasattr(ht.grpmax, 'gnomad')` before the `ht.select` instead?

v03_pipeline/lib/reference_datasets/hgmd.py
jklugherz: I removed `min_partitions`

v03_pipeline/lib/reference_datasets/gnomad.py
hanars: I think my issue is more that theres one table we use and its hardcoded so we should not be running conditional code as part of this loading process, the loading should be deterministic.  I think you should run that command locally, determine the structure of the table we are using, and then have this function use the fields that exist and not maintain code for fields that do not

jklugherz: So 38 genomes has the nested `gnomad` struct inside `grpmax` and 38 exomes does not, and that is why this conditional is here. It seems to be the only difference between loading for gnomad exomes and genomes. 

I made new `gnomad_genomes` and `gnomad_exomes` modules that call a shared `get_ht` from a `gnomad_utils` module and provide a `af_popmax_expression` function as an argument. I also got rid of the `MODULE_NAME` config attribute.

hanars (approved):

PR #960, opened 2024-11-12
Benb/clinvar refactor

bpblanken:


v03_pipeline/lib/reference_datasets/misc.py
```diff
@@ -0,0 +1,31 @@
...
+    ):
+        return field.map(
+            lambda x: lookup[x],
+        )
+    return lookup[field]
+
+
+def filter_contigs(ht, reference_genome: ReferenceGenome, dataset_type: DatasetType):
+    if dataset_type == DatasetType.MITO:
+        ht = ht.filter(ht.locus.contig == reference_genome.mito_contig)
```
jklugherz, line 28: is it better to return ht here since contigs will already be filtered to mito?

bpblanken: Yes!  Good suggestion!

v03_pipeline/lib/reference_datasets/misc.py
```diff
@@ -0,0 +1,31 @@
...
+            # hard fail if a mapped key is present and has an unexpected value
+            # but propagate missing values.
+            [(hl.missing(hl.tstr), hl.missing(hl.tint32))],
+        ),
+    )
+    if (
+        isinstance(field.dtype, hl.tarray | hl.tset)
+        and field.dtype.element_type == hl.tstr
+    ):
+        return field.map(
```
jklugherz, line 20: Why did we originally have this notation `[f'{field_name}_ids']` in the enum map? Do we still need it?

bpblanken: Ah.  I removed a lot of the logic that was building the `selects` and `enum_selects` dictionaries in favor of just selecting fields manually.  This means we'll need to name the fields ourselves... which I thought was actually clearer and more maintainable than the existing "magic", especially for `clinvar`, where we were doing both a select and then an enum select. 

For example, [assertion_ids](https://github.com/broadinstitute/seqr-loading-pipelines/pull/960/files#diff-caa8fea0ec85be082e2702ee31130dc4a8af587722e66be3ace335dd075198d0R153).

jklugherz (approved):

v03_pipeline/var/test/reference_data/clinvar.vcf.gz
jklugherz, line 1: Do we want to also start a new `reference_datasets` directory for test fixtures?

v03_pipeline/lib/reference_datasets/clinvar.py
```diff
@@ -0,0 +1,189 @@
...
+        force_bgz=True,
+    ).rows()
+    ht = filter_contigs(ht, reference_genome, dataset_type)
+    submitters_ht = get_submission_summary_ht()
+    ht = ht.annotate(
+        submitters=submitters_ht[ht.rsid].Submitters,
+        conditions=submitters_ht[ht.rsid].Conditions,
+    )
+    ht = select_fields(ht)
+    return ht.annotate_globals(
```
hanars, line 183: theres a  very similar call to `annotate_globals` in the `create_table` method of the update task. I feel like because we have such a consistent schema for how we want globals to be annotated it would be better to have this logic shared/ generalized in some way

v03_pipeline/lib/model/definitions.py
```diff
@@ -63,6 +63,13 @@ def optional_contigs(self) -> set[str]:
             },
         }[self]
 
+    @property
+    def mito_contig(self) -> str:
+        return {
+            ReferenceGenome.GRCh37: 'MT',
+            ReferenceGenome.GRCh38: 'chrM',
+        }
```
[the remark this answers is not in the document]
bpblanken: yes!

v03_pipeline/lib/reference_datasets/clinvar.py
jklugherz: one related question: should the update task be doing any globals annotation or should that happen only in the `get_ht` scripts?

bpblanken: yeah, I also had this question and thought.  I didn't love the enum magic we were doing to try to fit `clinvar` into the config structure last time around, so felt it was better to avoid putting `enums` in the config and just `annotate` manually in the `get_ht` functions.  It's not a bad idea to _also_ put the `enums` in the config so that we can share the `globals` logic downstream.

hanars: agreed that putting them in the config and sharing the logic is worth doing even if we don't use it in the `get_ht` function

v03_pipeline/lib/reference_datasets/clinvar.py
```diff
@@ -0,0 +1,189 @@
...
+        shutil.copyfileobj(r.raw, tmp_file)
+    ht = hl.import_vcf(
+        tmp_file.name,
+        reference_genome=reference_genome.value,
+        drop_samples=True,
+        skip_invalid_loci=True,
+        contig_recoding=reference_genome.contig_recoding(include_mt=True),
+        force_bgz=True,
+    ).rows()
+    ht = filter_contigs(ht, reference_genome, dataset_type)
```
hanars, line 176: I thought we agreed to maintain one table per dataset type and do the filtering at join time

bpblanken: Yes, that's correct! I did this wrong.

bpblanken: I removed `dataset_type` from this function and we'll manage the filtering to `mito` downstream when we annotate the annotations table, which was my original thought and got ahead of myself here.

v03_pipeline/lib/reference_datasets/clinvar.py
```diff
@@ -0,0 +1,187 @@
...
+        shutil.copyfileobj(r.raw, tmp_file)
+    ht = hl.import_vcf(
+        tmp_file.name,
+        reference_genome=reference_genome.value,
+        drop_samples=True,
+        skip_invalid_loci=True,
+        contig_recoding=reference_genome.contig_recoding(include_mt=True),
+        force_bgz=True,
+    ).rows()
+    ht = filter_contigs(ht, reference_genome)
```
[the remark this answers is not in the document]
bpblanken: 👍

v03_pipeline/lib/reference_datasets/clinvar.py
hanars: other than this I think all my comments have been resolved!

hanars (changes requested):

v03_pipeline/lib/reference_datasets/clinvar.py
bpblanken: yep!  It's in progress.

v03_pipeline/lib/reference_datasets/misc.py
hanars: except if you just changed the line for annotating the assertions in `select_fields` from `assertions=hl.if_else(` to `assertion=hl.if_else(` wouldn't the existing enum functionality properly name everything? Since seqr itself does "magic" parsing of the enums into fields, it feels like using shared functionality that assigns  field names to enums is much safer than every function needing to get this right

bpblanken: it wasn't the `id` or `ids` suffix that was the issue, it was the two pass select... one to determine the field names and a second to actually do the transmute.  That didn't make sense to me if we're not configuring enums, but if we _are_, the calculus probably changes.  I can add it back.

hanars: I think we are configuring them right? The current `select_fields` func does a select followed by a transmute so I feel like adding it back makes sense

bpblanken: Cool, I added this back.  I'm still a little skeptical if the configuration is the right approach here in general (just because we have very few enum fields and rarely edit them), but this brings us to parity and is ok!

hanars (approved):

PR #964, opened 2024-11-14
dbnsfp reference dataset

hanars:
Adds DBNSFP. I moved the transmute for enums into the shared code as I think we would always want to run it if there are enums. I also move enums up out of the reference genome nesting because in our current config the enums are identical across reference builys which means we have a bunch of copy-paste for them and I feel like its easier to keep things synced up if we only define them once, but I can revert that if folks feel like its better as is


v03_pipeline/lib/reference_datasets/reference_dataset.py
```diff
@@ -92,36 +102,32 @@ def requires(self) -> ReferenceDataset:
...
             VERSION: '1.0',
-            RAW_DATASET_PATH: [
-                'https://krishna.gs.washington.edu/download/CADD/v1.7/GRCh38/whole_genome_SNVs.tsv.gz',
-                'https://krishna.gs.washington.edu/download/CADD/v1.7/GRCh38/gnomad.genomes.r4.0.indel.tsv.gz',
-            ],
+            RAW_DATASET_PATH: 'https://dbnsfp.s3.amazonaws.com/dbNSFP4.7a.zip',
         },
     },
     ReferenceDataset.clinvar: {
+        ENUMS: clinvar.ENUMS,
```
bpblanken, line 121: 👍 this looks ok to me!  I know at some point in the past we've had different enums for the different reference genomes but we can handle that later on if need me.

v03_pipeline/lib/reference_datasets/reference_dataset.py
```diff
@@ -48,9 +51,13 @@ def version(self, reference_genome: ReferenceGenome) -> str:
             )
         return version
 
-    def enums(self, reference_genome) -> hl.Struct:
-        if ENUMS in CONFIG[self][reference_genome]:
-            return hl.Struct(**CONFIG[self][reference_genome][ENUMS])
+    @property
+    def enums(self) -> dict:
```
bpblanken, line 55: type on this can be `dict | None`

v03_pipeline/lib/reference_datasets/reference_dataset.py
```diff
@@ -48,9 +51,13 @@ def version(self, reference_genome: ReferenceGenome) -> str:
...
+        return CONFIG[self].get(ENUMS)
+
+    def enum_globals(self) -> hl.Struct:
```
bpblanken, line 58: can be an `@property`

v03_pipeline/lib/reference_datasets/reference_dataset.py
bpblanken: I think yes, that's a great idea.  There isn't a good reason to have them in the data module now if they are only referenced here.

v03_pipeline/lib/reference_datasets/dbnsfp.py
```diff
@@ -0,0 +1,83 @@
...
+def predictor_parse(field: hl.StringExpression) -> hl.StringExpression:
+    return field.split(';').find(lambda p: p != '.')
+
+
+# adapted from download_and_create_reference_datasets/v02/hail_scripts/write_dbnsfp_ht.py
+def get_ht(raw_dataset_path: str, reference_genome: ReferenceGenome) -> hl.Table:
+    types = TYPES[reference_genome]
+    rename = RENAME[reference_genome]
+
+    with tempfile.TemporaryDirectory() as temp_dir:
```
bpblanken, line 64: I think there's a subtle and annoying bug here.  This directory will get deleted when the function returns a lazy handle to the hail table.  But, when we go to actually write the table in `UpdatedReferenceDataset`, the underlying `*.gz` files will have disappeared and hail will throw an exception.  

We have a few alternatives here:
1) leave the temp files in place and pollute disk (which is what we are now doing for clinvar/mitimpact).  This is generally fine but might cause issues for large datasets that might be >100GB.
2) checkpoint to `seqr-scratch-temp` inside this function and clean up disk immediately.

I think moving towards 2) (or brainstorming something else) is probably better!

hanars: I'm just going to share what you did for mitimpact

bpblanken: 👍

v03_pipeline/lib/reference_datasets/misc.py
```diff
@@ -31,3 +38,31 @@ def filter_contigs(ht, reference_genome: ReferenceGenome):
...
     )
+
+
+def key_by_locus_alleles(ht: hl.Table, reference_genome: ReferenceGenome) -> hl.Table:
+    chrom = (
+        hl.format('chr%s', ht.chrom)
+        if reference_genome == ReferenceGenome.GRCh38
+        else ht.chrom
+    )
+    locus = hl.locus(chrom, ht.pos, reference_genome.value)
```
bpblanken, line 49: can these three lines be condensed into one `ht = ht.transmute`?

v03_pipeline/lib/reference_datasets/misc.py
```diff
@@ -31,3 +38,31 @@ def filter_contigs(ht, reference_genome: ReferenceGenome):
...
+    alleles = hl.array([ht.ref, ht.alt])
+    ht = ht.transmute(locus=locus, alleles=alleles)
+    return ht.key_by('locus', 'alleles')
+
+
+@contextmanager
+def download_zip_file(url, suffix='.zip'):
+    extracted_filename = url.removesuffix('.zip').split('/')[-1]
```
bpblanken, line 57: actually, this line was sort-of special cased to `mitimpact`... the file that's extracted from the archive is just `MitImpact_db_3.1.3.txt`.  I think that isn't true for `dbnsfp`?

hanars: oh no you are right

PR #966, opened 2024-11-14
gnomad_mito, gnomad_non_coding_constraint, local_constraint_mito, screen

jklugherz:


v03_pipeline/lib/reference_datasets/gnomad_mito.py
```diff
@@ -0,0 +1,18 @@
...
+    selects['AC_het'] = hl.int32(ht.AC_het)
+    selects['AF_hom'] = ht.AF_hom
+    selects['AF_het'] = ht.AF_het
+    selects['max_hl'] = ht.max_hl
+    return ht.select(**selects)
+
+
+def get_ht(raw_dataset_path: str, *_) -> hl.Table:
+    ht = hl.read_table(raw_dataset_path)
+    ht = select_fields(ht)
```
bpblanken, line 17: same comment as elsewhere, but I think just 

```
ht = ht.select(
 AN=hl.int32(ht.AN)
...
)
return ht.select_globals()
``` 

is simpler!

v03_pipeline/lib/reference_datasets/reference_dataset.py
```diff
@@ -209,5 +213,47 @@ def requires(self) -> ReferenceDataset:
...
+            DATASET_TYPES: frozenset([DatasetType.MITO]),
+            VERSION: '1.0',
+            RAW_DATASET_PATH: 'gs://gcp-public-data--gnomad/release/3.1/ht/genomes/gnomad.genomes.v3.1.sites.chrM.ht',
+        },
+    },
+    ReferenceDataset.gnomad_non_coding_constraint: {
+        ReferenceGenome.GRCh38: {
+            DATASET_TYPES: frozenset([DatasetType.SNV_INDEL]),
+            VERSION: '1.0',
+            RAW_DATASET_PATH: 'gs://seqr-reference-data/GRCh38/gnomad_nc_constraint/gnomad_non-coding_constraint_z_scores.ht',
```
hanars, line 227: Wherever possible we should not be hosting reference datasets. If this is not publicly available we should add an explicit comment to the config explaining where we got the data that we put in the public bucket and why we are hosting the data ourselves

v03_pipeline/lib/reference_datasets/reference_dataset.py
```diff
@@ -209,5 +213,47 @@ def requires(self) -> ReferenceDataset:
...
+                'dELS',
+                'pELS',
+                'DNase-only',
+                'low-DNase',
+            ],
+        },
+        ReferenceGenome.GRCh38: {
+            DATASET_TYPES: frozenset([DatasetType.SNV_INDEL]),
+            VERSION: '1.0',
+            RAW_DATASET_PATH: 'gs://seqr-reference-data/GRCh38/ccREs/GRCh38-ccREs.ht',
```
hanars, line 246: Same as above

v03_pipeline/lib/reference_datasets/reference_dataset.py
```diff
@@ -209,5 +213,47 @@ def requires(self) -> ReferenceDataset:
...
+        },
+    },
+    ReferenceDataset.local_constraint_mito: {
+        ReferenceGenome.GRCh38: {
+            DATASET_TYPES: frozenset([DatasetType.MITO]),
+            VERSION: '1.0',
+            # Originally sourced from https://www.biorxiv.org/content/10.1101/2022.12.16.520778v2.supplementary-material
+            # Supplementary Table 7.
+            RAW_DATASET_PATH: 'gs://seqr-reference-data/GRCh38/mitochondrial/local_constraint.tsv',
```
hanars, line 255: Same as above. In this case I think we had downloaded it because we did not want to deal with zip archives which is how this table is available in biorxiv but now that we wrote helpers for downloading tables from zip files we should use the original host instead: 
https://www.biorxiv.org/content/biorxiv/early/2023/01/27/2022.12.16.520778/DC3/embed/media-3.zip?download=true

v03_pipeline/lib/reference_datasets/reference_dataset.py
jklugherz: ah there's a script that I missed to create the ht from the .bed file, which I need to incorporate into get_ht  for `screen`

v03_pipeline/lib/reference_datasets/reference_dataset.py
hanars: So looking at the README in that file,  `constraint_z_genome_1kb.raw.download.txt.gz` seems to be everything, `constraint_z_genome_1kb.qc.download.txt.gz` seems to be everything that passed QC and constraint_z_enh_gene_roadmaplinks.all.download.txt.gz is z scores of enhancers for specific genes. The enhancer one is definitely wrong, unfortunately we did not document what we used when we copied that file so aI am also not sure which is better.
 According to the README theres ~2,500,000 raw scores and around ~2,000,000 passing QC scores. Maybe check the file we currently have in GCP and see if it has a number similar to either of those and if ti does use that and if not just use the raw scores

jklugherz: we have 2,042,629 rows in our hosted reference dataset, so I'll use the raw scores

jklugherz: true, I thought having fewer scores would be worse but I think it's a closer match. I used the passing scores file.

v03_pipeline/lib/reference_datasets/reference_dataset.py
jklugherz: screen is updated!

v03_pipeline/lib/reference_datasets/reference_dataset.py
jklugherz: this is my last todo on this PR

v03_pipeline/lib/reference_datasets/misc.py
```diff
@@ -78,8 +77,29 @@ def download_zip_file(url, suffix='.zip'):
     with tempfile.NamedTemporaryFile(
         suffix=suffix,
     ) as tmp_file, requests.get(url, stream=True, timeout=10) as r:
-        shutil.copyfileobj(r.raw, tmp_file)
+        tmp_file.write(r.content)
```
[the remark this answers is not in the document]
bpblanken: hrm, this is weird.  investigating!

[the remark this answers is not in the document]
bpblanken: Ok! I figured out what's going on here.  The `.zip` file we added here is sneakily `gzip`'ed ('Content-Encoding': 'gzip') in addition to being a `zip` archive:

```
In [84]: r = requests.get('https://www.biorxiv.org/content/biorxiv/early/2023/01/27/2022.12.16.520778/DC3/embed/media-3.zip', stream=True, timeout=10)
In [85]: len(r.raw.read(decode_content=True))
Out[85]: 332443
In [86]: r = requests.get('https://www.biorxiv.org/content/biorxiv/early/2023/01/27/2022.12.16.520778/DC3/embed/media-3.zip', stream=True, timeout=10)
In [87]: len(r.raw.read())
Out[87]: 319854
```

The `shuttle.copyfileobj` source does not decode the content, so we're left with a double compressed tmp file.  This approach is ok.  An alternative might be to just roll our own `copyfileobj` to keep the chunked approach.  Here's the existing method:  

```
def copyfileobj(fsrc, fdst, length=16*1024):
    """copy data from file-like object fsrc to file-like object fdst"""
    while 1:
        buf = fsrc.read(length)
        if not buf:
            break
        fdst.write(buf)
```

we could just add `decode_content=True` to `read` and it would work.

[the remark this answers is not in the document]
jklugherz: Thanks for looking into this. It's probably better if we keep the chunking approach, so I added the new `copyfileobj` implementation.

bpblanken (approved):

PR #967, opened 2024-11-14
Splice AI reference dataset

hanars:


v03_pipeline/lib/reference_datasets/reference_dataset.py
```diff
@@ -132,6 +134,35 @@ def requires(self) -> ReferenceDataset:
...
+        ENUMS: {
+            'splice_consequence': [
+                'Acceptor gain',
+                'Acceptor loss',
+                'Donor gain',
+                'Donor loss',
+                'No consequence',
+            ],
+        },
+        SKIP_ENUM_ANNOTATION: True,
```
hanars, line 147: In the original script for writing the hail table we were actually mapping the index of an array to one of the the above strings. It seemed very silly to run that enum and then immediately transmute it back to the index, so I instead am just directly annotating the enumerated field and skipping the transmute step. 
I do still think its better behavior to have the enum transmute run by default for everything with enums, as I think this edge case will be rare

v03_pipeline/lib/reference_datasets/reference_dataset.py
```diff
@@ -132,6 +134,35 @@ def requires(self) -> ReferenceDataset:
...
+            RAW_DATASET_PATH: [
+                'gs://seqr-reference-data/GRCh37/spliceai/spliceai_scores.masked.snv.hg19.vcf.gz',
+                'gs://seqr-reference-data/GRCh37/spliceai/spliceai_scores.masked.indel.hg19.vcf.gz',
+            ],
+        },
+        ReferenceGenome.GRCh38: {
+            DATASET_TYPES: frozenset([DatasetType.SNV_INDEL]),
+            VERSION: '1.0',
+            # NB: SpliceAI data is only available to download for authenticated Illumina users, so we will host the data
+            RAW_DATASET_PATH: [
```
bpblanken, line 160: I don't see these paths existing....


```
bblanken@wmbbe-67c seqr-loading-pipelines % gsutil ls gs://seqr-reference-data/GRCh37/spliceai/
gs://seqr-reference-data/GRCh37/spliceai/
gs://seqr-reference-data/GRCh37/spliceai/exome_spliceai_scores.vcf.gz
gs://seqr-reference-data/GRCh37/spliceai/whole_genome_filtered_spliceai_scores.vcf.gz
gs://seqr-reference-data/GRCh37/spliceai/spliceai_scores.ht/
bblanken@wmbbe-67c seqr-loading-pipelines % gsutil ls gs://seqr-reference-data/GRCh38/spliceai/
gs://seqr-reference-data/GRCh38/spliceai/
gs://seqr-reference-data/GRCh38/spliceai/exome_spliceai_scores.liftover.vcf.gz
gs://seqr-reference-data/GRCh38/spliceai/spliceai_scores.ht/
```

I also don't see the paths referenced in the original file!

hanars: I asked Be about the originals and he told me it would be better to dowload them from Illumina directly rather than having someone try and go unearth the old ones

v03_pipeline/lib/reference_datasets/reference_dataset.py
```diff
@@ -73,7 +74,7 @@ def get_ht(
         )
         path = self.raw_dataset_path(reference_genome)
         ht = module.get_ht(path, reference_genome)
-        if self.enums:
+        if self.enums and not CONFIG[self].get(SKIP_ENUM_ANNOTATION):
```
bpblanken, line 77: I don't love the added config option for this; I think it's confusing that we use it here and then three lines below call `annotate_globals` without referencing the option.  

What if, instead, we modify `get_enum_select_fields` to not hard fail if the field to enum map isn't present on the table?

hanars: ooh I like that!

bpblanken (approved):

PR #968, opened 2024-11-14
mitomap

jklugherz:


v03_pipeline/lib/reference_datasets/reference_dataset.py
```diff
@@ -209,5 +210,13 @@ def requires(self) -> ReferenceDataset:
             RAW_DATASET_PATH: 'gs://gcp-public-data--gnomad/release/4.0/pca/gnomad.v4.0.pca_loadings.ht',
         },
     },
+    ReferenceDataset.mitomap: {
+        ReferenceGenome.GRCh38: {
+            DATASET_TYPES: frozenset([DatasetType.MITO]),
+            VERSION: '1.0',
+            # Downloaded via https://www.mitomap.org/foswiki/bin/view/MITOMAP/ConfirmedMutations
+            RAW_DATASET_PATH: 'gs://seqr-reference-data/GRCh38/mitochondrial/MITOMAP/mitomap_confirmed_mutations_nov_2024.csv',
```
jklugherz, line 218: I just downloaded this, the last version we have was from 2022. Looks like there's about 30 more pathogenic variants.

hanars: yay newer data!

v03_pipeline/lib/reference_datasets/mitomap.py
```diff
@@ -0,0 +1,29 @@
...
+
+
+# adapted from download_and_create_reference_datasets/v02/mito/write_mito_mitomap_ht.py
+def get_ht(raw_dataset_path: str, reference_genome: ReferenceGenome) -> hl.Table:
+    ht = hl.import_table(
+        raw_dataset_path,
+        delimiter=',',
+        types={'"Position"': hl.tstr, '"Allele"': hl.tstr},
+    )
+    # NB: all strings in the downloaded CSV file are wrapped in double quotes
```
hanars, line 13: import_table takes a `quote` arguments - does using that resolve this issue?

v03_pipeline/lib/reference_datasets/mitomap.py
```diff
@@ -0,0 +1,29 @@
...
+
+from v03_pipeline.lib.model import ReferenceGenome
+
+
+# adapted from download_and_create_reference_datasets/v02/mito/write_mito_mitomap_ht.py
+def get_ht(raw_dataset_path: str, reference_genome: ReferenceGenome) -> hl.Table:
+    ht = hl.import_table(
+        raw_dataset_path,
+        delimiter=',',
+        types={'"Position"': hl.tstr, '"Allele"': hl.tstr},
```
hanars, line 11: by default hail imputes everything as a string, so the only reason to explicitly pass types for imputation is if you want it to not be a string (although for position it does feel weird that you can't just import it directly as an int)

v03_pipeline/lib/reference_datasets/mitomap.py
```diff
@@ -0,0 +1,29 @@
...
+        },
+    )
+    ht = ht.select(
+        locus=hl.locus(
+            'chrM',
+            hl.parse_int32(hl.parse_json(ht.Position, dtype='str')),
+            reference_genome=reference_genome.value,
+        ),
+        alleles=ht.Allele.first_match_in('m.[0-9]+([ATGC]+)>([ATGC]+)'),
+        pathogenic=True,
```
hanars, line 27: why do we assign a value thats always True?

v03_pipeline/lib/reference_datasets/mitomap.py
jklugherz: yes! thanks

v03_pipeline/lib/reference_datasets/reference_dataset.py
hanars: Why are we downloading this and storing a copy in seqr-reference-data instead of directly using the external download?

jklugherz: The site has a button to download the csv but I found no way to programmatically download via an api, I could be wrong though https://www.mitomap.org/foswiki/bin/view/MITOMAP/ConfirmedMutations

bpblanken: Ah!  It looks like the site generates temporary links when you click the "CSV" button.  For examples: blob:https://www.mitomap.org/0f70e9e7-b27b-4db6-9026-014578cdac39... These appear to expire quickly so yes, hosting our own is probably the right idea here.

hanars (approved):

PR #973, opened 2024-11-16
Fix bugs related to nested fields/split_multi

bpblanken:


v03_pipeline/lib/paths.py
```diff
@@ -328,7 +331,7 @@ def valid_reference_dataset_collection_path(
 
 def valid_reference_dataset_path(
     reference_genome: ReferenceGenome,
-    reference_dataset: ReferenceDataset,
+    reference_dataset: 'ReferenceDataset',
```
bpblanken, line 334: There's an annoying circular dependency that gets triggered because of the type import.  Restructuring felt out of scope for this.

v03_pipeline/lib/reference_datasets/exac.py
```diff
@@ -1,10 +1,11 @@
 import hail as hl
 
+from v03_pipeline.lib.misc.nested_field import parse_nested_field
 from v03_pipeline.lib.model import ReferenceGenome
 from v03_pipeline.lib.reference_datasets.misc import vcf_to_ht
 
-RENAME = {
-    'AF_POPMAX': 'info.AF_POPMAX',
+SELECT = {
+    'AF_POPMAX': 'info.POPMAX',
```
bpblanken, line 8: This field change comes from us using the `vcf` rather than the `ht`.  Test caught it!

v03_pipeline/lib/reference_datasets/exac.py
```diff
@@ -1,10 +1,11 @@
 import hail as hl
 
+from v03_pipeline.lib.misc.nested_field import parse_nested_field
```
hanars, line 3: I don't see this file/function anywhere?

bpblanken: Here it is! 

https://github.com/broadinstitute/seqr-loading-pipelines/blob/main/v03_pipeline/lib/misc/nested_field.py

it doesn't need to be moved into the `reference_datasets` folder because it's used when importing normal callsets too.

hanars: 🤦

hanars (approved):

v03_pipeline/lib/paths.py
jklugherz: when I first created the `ReferenceDataset` enum it was challenging to figure out how to avoid this

PR #974, opened 2024-11-18
Add test for `UpdatedReferenceDataset` and `UpdatedReferenceDatasetQuery`

bpblanken:


v03_pipeline/lib/tasks/reference_data/updated_reference_dataset_query_test.py
```diff
@@ -0,0 +1,92 @@
...
+                globals=hl.Struct(version='2021-01-01'),
+            ),
+            valid_reference_dataset_path(
+                ReferenceGenome.GRCh38,
+                ReferenceDataset.clinvar,
+            ),
+        )
+
+    @responses.activate
+    def test_updated_query_and_dependency(
```
bpblanken, line 51: This is a single test which covers both tasks, which was super straightforward!

jklugherz (approved):
👍

PR #978, opened 2024-11-20
update annotations with updated reference datasets refactor

jklugherz:
- updates variant annotations task to use new reference dataset sources
- updates tests including so many test fixtures to represent new reference data tables


v03_pipeline/lib/reference_datasets/reference_dataset.py
```diff
@@ -23,20 +25,31 @@ def for_reference_genome_dataset_type(
...
             if dataset_type in config.get(reference_genome, {}).get(DATASET_TYPES)
         ]
         if not Env.ACCESS_PRIVATE_REFERENCE_DATASETS:
-            return [
+            reference_datasets = [
                 dataset
                 for dataset in reference_datasets
                 if dataset.access_control == AccessControl.PUBLIC
             ]
+        if not include_queries:
```
bpblanken, line 41: I think we want this to not just exclude `queries`, but also `gnomad_qc` and `gnomad_coding_and_noncoding` right?  So maybe a config flag?

v03_pipeline/lib/tasks/reference_data/update_variant_annotations_table_with_updated_reference_dataset_new_test.py
```diff
@@ -0,0 +1,109 @@
...
+)
+from v03_pipeline.lib.test.mock_complete_task import MockCompleteTask
+from v03_pipeline.lib.test.mocked_dataroot_testcase import MockedDatarootTestCase
+
+TEST_SNV_INDEL_VCF = 'v03_pipeline/var/test/callsets/1kg_30variants.vcf'
+
+
+class UpdateVATWithUpdatedRDC(MockedDatarootTestCase):
+    def setUp(self) -> None:
+        super().setUp()
```
bpblanken, line 29: yeah, I don't know if we want a new file for this... we should just be replacing the existing three `valid_reference_dataset_collection_path` with calls to new `valid_refernece_dataset_path`.

v03_pipeline/lib/reference_datasets/reference_dataset.py
```diff
@@ -349,7 +362,12 @@ def get_ht(
     },
 }
 CONFIG[ReferenceDatasetQuery.clinvar_path_variants] = CONFIG[ReferenceDataset.clinvar]
+CONFIG[ReferenceDatasetQuery.clinvar_path_variants].EXCLUDE_FROM_ANNOTATIONS = True
```
bpblanken, line 365: What if we added `ReferenceDatasetQuery.clinvar_path_variants` to the config above and did:

```
  ReferenceDatasetQuery.clinvar_path_variants: {
     EXCLUDE_FROM_ANNOTATIONS: True,
     **CONFIG[ReferenceDataset.clinvar],
  }
```

I _think_ that works? (it works in js at least!)

v03_pipeline/lib/reference_datasets/reference_dataset.py
```diff
@@ -23,11 +27,13 @@ def for_reference_genome_dataset_type(
         cls,
         reference_genome: ReferenceGenome,
         dataset_type: DatasetType,
-    ) -> list['ReferenceDataset']:
+        include_all: bool = True
```
bpblanken: We have a linter rule disabled for this...but we should try to follow it if possible: 
```
    "FBT",   # flake-boolean-trap... disallows boolean args to functions... fixing this code will require refactors.
```

The idea is to define a separate function, so maybe `for_annotation(cls, reference_genome...)` that would disambiguate between the `include_all=True` or `include_all=False` approaches.

v03_pipeline/lib/annotations/snv_indel.py
```diff
@@ -80,9 +80,9 @@ def gnomad_non_coding_constraint(
         z_score=(
             interval_ht.index(ht.locus, all_matches=True)
             .filter(
-                lambda x: hl.is_defined(x.gnomad_non_coding_constraint['z_score']),
+                lambda x: hl.is_defined(x['z_score']),
```
bpblanken, line 83: These no longer take in the `interval_ht` right?  So the dependency itself change 🤔

jklugherz: The `interval_ht` is the `screen ht` and `ht` is the annotations table. I could rename the `interval_ht` arg to be representative of the dataset

v03_pipeline/lib/reference_datasets/reference_dataset.py
jklugherz: Ya I was thinking that separate functions would be easier to maintain

v03_pipeline/lib/annotations/snv_indel.py
bpblanken: gotcha, yeah.  I think making the arg name representative of the dataset is worthwhile now that the ht itself is dataset specific!

v03_pipeline/lib/tasks/reference_data/update_variant_annotations_table_with_updated_reference_dataset_test.py
```diff
@@ -974,152 +537,144 @@ def test_update_vat_with_updated_rdc_mito_38(
...
+                        clinvar=hl.Struct(
+                            alleleId=None,
+                            conflictingPathogenicities=None,
+                            goldStars=None,
+                            pathogenicity_id=None,
+                            assertion_ids=None,
+                            submitters=None,
+                            conditions=None,
+                        ),
+                        dbnsfp=hl.Struct(
```
jklugherz, line 642: `dbnsfp_mito` has fewer fields than the new mito `dbnsfp` annotation since we used to treat it as a separate dataset, so we'll need to decide how to handle this (keep the extra fields or add a config somewhere to filter down to the fields we want for the mito dataset type)

PR #982, opened 2024-11-20
Updated `gnomad_coding_and_noncoding` test table.

bpblanken:


v03_pipeline/lib/misc/validation.py
```diff
@@ -24,36 +18,6 @@ class SeqrValidationError(Exception):
     pass
 
 
-def get_validation_dependencies(
```
bpblanken, line 27: I opted to move this rather than just change `cached_reference_dataset_query_path` -> `valid_reference_dataset_path` because its location caught me by surprise.  This function was also the cause of the circular dependency issue we were seeing, so I guess it's more "correct" to be in the task rather than here.

PR #985, opened 2024-11-22
Make rdq dataset type specific

bpblanken:


v03_pipeline/lib/paths.py
```diff
@@ -338,22 +338,22 @@ def valid_reference_dataset_path(
             reference_dataset.access_control,
             reference_genome,
         ),
-        'reference_datasets',
```
bpblanken, line 341: removed this constant since it's no longer necessary to disambiguate between query types.

v03_pipeline/lib/reference_datasets/misc.py
```diff
@@ -45,6 +47,16 @@ def get_enum_select_fields(
     return enum_select_fields
 
 
+def mito_contig_filter(
+    reference_genome: ReferenceGenome,
+    dataset_type: DatasetType,
+    ht: hl.Table,
+) -> Callable[[DatasetType, ReferenceGenome, hl.Table], hl.Expression]:
```
jklugherz, line 54: any reason to return a callable instead of calling `hl.filter` inside of the function? (I'm also wondering if I should do this for the custom select)

bpblanken: I don't think there is!  We're doing that with `filter_contigs`, so we should be consistent with that (or make `filter_contigs` consistent with this, but I think changing it makes it just ever so slightly clearer)!

jklugherz (approved):

PR #986, opened 2024-11-22
annotate new variants with updated rds

jklugherz:


bpblanken:
this looked good to me!

PR #989, opened 2024-11-25
Benb/other cleanup

bpblanken:


v03_pipeline/lib/reference_datasets/misc.py
```diff
@@ -57,6 +57,12 @@ def filter_mito_contigs(
 
 
 def filter_contigs(ht, reference_genome: ReferenceGenome):
+    if hasattr(ht, 'interval'):
```
bpblanken, line 60: I made a small change to the logic here.  Instead of skipping this we filter within the interval.  I believe this is the same as what we have in production right now.

v03_pipeline/lib/annotations/fields_test.py
```diff
@@ -11,10 +11,6 @@
 from v03_pipeline.lib.paths import valid_reference_dataset_path
 from v03_pipeline.lib.reference_datasets.reference_dataset import ReferenceDataset
 from v03_pipeline.lib.test.mocked_dataroot_testcase import MockedDatarootTestCase
-from v03_pipeline.lib.test.mocked_reference_dataset_test_case import (
```
bpblanken, line 14: `mocked_reference_datasets_test_case` no longer has these individually referenced.

v03_pipeline/lib/reference_datasets/reference_dataset.py
```diff
@@ -15,71 +15,54 @@
...
 
 class BaseReferenceDataset:
     @classmethod
-    def _for_reference_genome_dataset_type(
+    def for_reference_genome_dataset_type(
         cls,
         reference_genome: ReferenceGenome,
         dataset_type: DatasetType,
-    ) -> list[Union['ReferenceDataset', 'ReferenceDatasetQuery']]:
+    ) -> set[Union['ReferenceDataset', 'ReferenceDatasetQuery']]:
```
bpblanken, line 33: I changed these to be sets instead of lists for easier downstream set union and symmetric difference

v03_pipeline/lib/tasks/reference_data/update_variant_annotations_table_with_updated_reference_dataset.py
```diff
@@ -19,60 +18,50 @@
...
         annotation_ht_versions = dict(
             hl.eval(hl.read_table(self.output().path).globals.versions),
         )
-
-        # Find datasets with mismatched versions
-        for dataset in reference_datasets:
-            if annotation_ht_versions.get(dataset.name) != dataset.version(
-                self.reference_genome,
+        self._datasets_to_update = (
+            reference_dataset_names ^ annotation_ht_versions.keys()
```
jklugherz, line 45: nice this is a cleaner way to to find datasets to drop

v03_pipeline/lib/tasks/reference_data/update_variant_annotations_table_with_updated_reference_dataset_test.py
```diff
@@ -412,21 +376,10 @@ def test_update_vat_mito_38(
     @patch(
         'v03_pipeline.lib.tasks.base.base_update_variant_annotations_table.BaseUpdateVariantAnnotationsTableTask.initialize_table',
     )
-    @patch(
-        'v03_pipeline.lib.tasks.base.base_update_variant_annotations_table.UpdatedReferenceDatasetTask',
-    )
-    @patch(
-        'v03_pipeline.lib.tasks.base.base_update_variant_annotations_table.UpdatedReferenceDatasetQueryTask',
-    )
```
jklugherz, line 420: were these mocks just not needed?

v03_pipeline/lib/test/mocked_reference_datasets_testcase.py
```diff
@@ -0,0 +1,40 @@
...
+        super().setUp()
+        for reference_genome in ReferenceGenome:
+            with mock_clinvar_urls(reference_genome):
+                path = os.path.join(
+                    REFERENCE_DATASETS_PATH,
+                    reference_genome.value,
+                )
+                # Use listdir, allowing for missing datasets
+                # in the tests.
+                for dataset_name in os.listdir(
```
jklugherz, line 27: nice this is also much cleaner!

v03_pipeline/lib/tasks/reference_data/update_variant_annotations_table_with_updated_reference_dataset_test.py
bpblanken: Sorta!  The `UpdatedReferenceDataset` are no-ops if every reference dataset is on disk, which they now are (there were maybe 2-3 missing).  The `UpdatedReferenceDatasetQuery` now just run!  It's a slight performance detriment to the tests though.

jklugherz (approved):

PR #994, opened 2024-12-02
Resolve bugs and increment versions.  

bpblanken:
A smattering of bugs I found during QA.


jklugherz (approved):

v03_pipeline/lib/reference_datasets/exac.py
```diff
@@ -18,5 +17,12 @@
 def get_ht(path: str, reference_genome: ReferenceGenome) -> hl.Table:
     ht = vcf_to_ht(path, reference_genome, split_multi=True)
     return ht.select(
-        **{k: parse_nested_field(ht, v) for k, v in SELECT.items()},
+        AF_POPMAX=hl.or_missing(
```
bpblanken, line 20: `info.POPMAX` is a string representing population, which isn't the right field.

v03_pipeline/lib/reference_datasets/splice_ai.py
```diff
@@ -49,5 +49,6 @@ def get_ht(
         ),
     ).drop('delta_scores')
     return ht.group_by(*ht.key).aggregate(
+        delta_score=hl.agg.max(ht.delta_scores),
```
hanars, line 52: so the aggregation is now happening twice in an inconsistent way, which I do not think is correct. Essentially, its parsing the VCF into a list of scores per site, then aggregating those scores, then at the end its doing a group by here and re-aggregating. I think this is unliklely to be producing the correct result. I think we need to do a quick investigation into places where we do have multiple sites after the VCF is initially read in an see if there really are different `ht.info.SpliceAI` values for the exact same site multiple times in the vcf. If not and theres just weird duplication a call to `distinct` would be better than a `group_by`, and if there are then we need to be explicitly deciding which of the duplicate entries to use

Also this shouldn't work because `delta_scores` is dropped on the line above this

bpblanken: I'm looking into this more thoroughly.  This was a solution but not a good one.

bpblanken: Looks like the duplicates (that I could find) are related to gene/transcripts.  I think the easiest path forwards is to filter to duplicate rows and remove those with a `info.SpliceAI[1]` that contains a `.`.  There are instances of transcript scores that do not have corresponding gene rows.  
````
+---------------+------------+------+-----------+----------+------------------------------------------------+
| locus         | alleles    | rsid |      qual | filters  | info.SpliceAI                                  |
+---------------+------------+------+-----------+----------+------------------------------------------------+
| locus<GRCh37> | array<str> | str  |   float64 | set<str> | array<str>                                     |
+---------------+------------+------+-----------+----------+------------------------------------------------+
| 1:861264      | ["C","A"]  | NA   | -1.00e+01 | NA       | ["A|AL645608.1|0.00|0.00|0.00|0.00|2|27|12|1"] |
| 1:861264      | ["C","A"]  | NA   | -1.00e+01 | NA       | ["A|SAMD11|0.02|0.01|0.00|0.00|14|38|14|38"]   |
+---------------+------------+------+-----------+----------+------------------------------------------------+
```

v03_pipeline/lib/misc/nested_field.py
```diff
@@ -12,4 +12,7 @@ def parse_nested_field(t: hl.MatrixTable | hl.Table, fields: str):
             ]
         else:
             expression = expression[field]
+    # Parse float64s into float32s to save space!
+    if expression.dtype == hl.tfloat64:
```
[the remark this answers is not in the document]
bpblanken: ah, this isn't actually new code.  It's something that was dropped during the refactor... it only impacts a couple of fields across all reference datasets though because we are much more manual in how we're selecting fields now and were mostly thorough with correctly specifying `float32`.

hanars (approved):

hanars (approved):

PR #997, opened 2024-12-10
Add service account credentialing

bpblanken:
service account token-ing is required for terra api requests...this gets us credentialed both locally and in dataproc.

### DIFF
--- a/v03_pipeline/lib/misc/gcp.py
+++ b/v03_pipeline/lib/misc/gcp.py
@@ -1,0 +1,36 @@
+import datetime
+
+import google.auth
+import google.auth.transport.requests
+import google.oauth2.credentials
+import pytz
+
+SERVICE_ACCOUNT_CREDENTIALS = None
+SOCIAL_AUTH_GOOGLE_OAUTH2_SCOPE = [
+    'https://www.googleapis.com/auth/userinfo.profile',
+    'https://www.googleapis.com/auth/userinfo.email',
+    'openid',
+]
+ONE_MINUTE_S = 60
+
+
+def get_service_account_credentials() -> google.oauth2.credentials.Credentials:
+    global SERVICE_ACCOUNT_CREDENTIALS
+    if not SERVICE_ACCOUNT_CREDENTIALS:
+        SERVICE_ACCOUNT_CREDENTIALS, _ = google.auth.default(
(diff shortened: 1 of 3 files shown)


v03_pipeline/lib/tasks/dataproc/create_dataproc_cluster.py
```diff
@@ -35,6 +38,8 @@ def get_cluster_config(reference_genome: ReferenceGenome, run_id: str):
                     'REFERENCE_GENOME': reference_genome.value,
                     'PIPELINE_RUNNER_APP_VERSION': Env.PIPELINE_RUNNER_APP_VERSION,
                 },
+                'service_account': service_account_credentials.service_account_email,
```
bpblanken, line 41: Note we will also call `get_service_account_credentials()` downstream when we actually make the API request.  This ensures we properly scope and credential the cluster at spinup time.

v03_pipeline/lib/misc/gcp.py
```diff
@@ -0,0 +1,36 @@
...
+SOCIAL_AUTH_GOOGLE_OAUTH2_SCOPE = [
+    'https://www.googleapis.com/auth/userinfo.profile',
+    'https://www.googleapis.com/auth/userinfo.email',
+    'openid',
+]
+ONE_MINUTE_S = 60
+
+
+def get_service_account_credentials() -> google.oauth2.credentials.Credentials:
+    global SERVICE_ACCOUNT_CREDENTIALS
```
jklugherz, line 18: Why are you using the `global` keyword here?

jklugherz: I think I figured it out. Is it because we don't want to reset `SERVICE_ACCOUNT_CREDENTIALS` every time `get_service_account_credentials()` is called?

bpblanken: I'm changing a variable with `global` scope.  More info: https://stackoverflow.com/questions/4693120/why-isnt-the-global-keyword-needed-to-access-a-global-variable

bpblanken: yeah, we just refresh the token.

jklugherz (approved):

PR #998, opened 2024-12-11
Add helper functions for querying `Terra Data Repository`

bpblanken:
Adds helper functions for querying `tdr` and the underlying `big query` stores.  Does not manage the persistence of these into the pipeline!


v03_pipeline/lib/misc/allele_registry.py
```diff
@@ -96,13 +96,7 @@ def register_alleles(
...
     with hfs.open(formatted_vcf_file_name, 'r') as vcf_in:
         data = vcf_in.read()
-        s = requests.Session()
-        retries = Retry(
-            total=5,
-            backoff_factor=1,
-            status_forcelist=[500, 502, 503, 504],
-        )
-        s.mount('https://', HTTPAdapter(max_retries=retries))
+        s = requests_retry_session()
```
jklugherz, line 99: 👍

v03_pipeline/lib/misc/terra_data_repository.py
```diff
@@ -0,0 +1,82 @@
...
+    for item in items:
+        if not any(x['cloudResource'] == BIGQUERY_RESOURCE for x in item['storage']):
+            # Hard failure on purpose to prompt manual investigation.
+            msg = 'Datasets without bigquery sources are unsupported'
+            raise ValueError(msg)
+    return [x['id'] for x in items]
+
+
+def gen_bq_table_names() -> Generator[str]:
+    with ThreadPoolExecutor(max_workers=5) as executor:
```
jklugherz, line 58: How did you chose 5 `max_workers`?

v03_pipeline/lib/misc/terra_data_repository.py
```diff
@@ -0,0 +1,82 @@
...
+    client = bigquery.Client()
+    return client.query_and_wait(
+        f"""
+        SELECT {','.join(BIGQUERY_METRICS)}
+        FROM `{table_name}.sample`
+    """,  # noqa: S608
+    )
+
+
+def _get_dataset_ids() -> list[str]:
```
jklugherz, line 46: Are we calling this on every pipeline run? It fetches all of the datasets in the TDR, is there a way to filter it first by dataset name perhaps?

bpblanken: Yes! This should be called every pipeline run, but it's only a single API request.  The plan was to use the result of this request + a persisted list of dataset ids that we've seen before to create a "new dataset ids" list that would be passed into `gen_bq_sample_metrics`.

v03_pipeline/lib/misc/terra_data_repository.py
bpblanken: I believe ai was responsible for this choice 🍭.

PR #999, opened 2024-12-13
Build Sex Check Table from TDR Metrics

bpblanken:


v03_pipeline/lib/tasks/write_tdr_metrics_file.py
```diff
@@ -0,0 +1,35 @@
...
+from v03_pipeline.lib.paths import tdr_metrics_path
+from v03_pipeline.lib.tasks.base.base_loading_pipeline_params import (
+    BaseLoadingPipelineParams,
+)
+from v03_pipeline.lib.tasks.files import GCSorLocalTarget
+
+
+@luigi.util.inherits(BaseLoadingPipelineParams)
+class WriteTDRMetricsFileTask(luigi.Task):
+    bq_table_name = luigi.Parameter()
```
bpblanken, line 19: The idea is to write the "metrics" files out as tsvs, one per "TDR dataset".  We can then import them into the pipeline's existing sex check structure.  Later on, we can tweak that to be less reliant on hail and more performant.

v03_pipeline/lib/tasks/base/base_loading_run_params.py
```diff
@@ -34,6 +34,10 @@ class BaseLoadingRunParams(luigi.Task):
         default=False,
         parsing=luigi.BoolParameter.EXPLICIT_PARSING,
     )
+    skip_expect_tdr_metrics = luigi.BoolParameter(
+        default=False,
+        parsing=luigi.BoolParameter.EXPLICIT_PARSING,
+    )
```
matren395, line 40: what's the logic here? will it be an option to not use TDR metrics directly ?

matren395, line 40: what's the logic here? will it be possible to not use TDR metrics and pass in a TSV directly ?

matren395 (commented):
this will import/read every TDR dataset for every run, right ? how tested is the performance ?

bpblanken:
```
this will import/read every TDR dataset for every run, right ? how tested is the performance ?
```

it shouldn't no... There will be one batch of api requests to fetch the dataset names, but we don't actually make the bq query unless the metrics file isn't persisted.

v03_pipeline/lib/tasks/base/base_loading_run_params.py
bpblanken: yes, we actually can pass in a tsv directly.  The tsvs are stored, one per TDR "dataset" here: https://github.com/broadinstitute/seqr-loading-pipelines/pull/999/files#diff-583758b4dea21ca2ae6729f0255825672e2de91bdc96b90aab88a8657f2f3cf5R115

bpblanken: We'll be able to modify them if need be, though we really shouldn't.

bpblanken:
Yep!  Performance has been fine locally.

PR #1000, opened 2024-12-13
feat: Handle parsing empty predicted sex into Unknown

bpblanken:


v03_pipeline/lib/model/definitions.py
```diff
@@ -18,12 +18,12 @@ class Sex(str, Enum):
     XYY = 'XYY'
 
     @property
-    def imputed_sex_value(self):
+    def imputed_sex_values(self) -> list[str]:
```
bpblanken, line 21: :todo this should be renamed to predicted_sex value but it was too messy to handle on this pr.

matren395 (approved):
lgtm!

PR #1003, opened 2024-12-17
bugfix: exclude samples from relationship checking that are not present in the expected loadable samples

bpblanken:
This fixes a regression caused by [this pr](https://github.com/broadinstitute/seqr-loading-pipelines/pull/921/files), which modified the relationship checking behavior for parents that are identified in the pedigree but excluded from loading.  

Originally we excluded those parental ids from the parsed relationships, which did not work in cases where the parental id helped identify sibling pairs.  

This pr preserves the behavior of excluded parents (test case is [here](https://github.com/broadinstitute/seqr-loading-pipelines/blob/main/v03_pipeline/lib/misc/pedigree_test.py#L466)), but also allows us to exclude relationships where the sample is not included as loadable in the pedigree (test case is [here](https://github.com/broadinstitute/seqr-loading-pipelines/pull/1003/files#diff-380bf733d18b80ac2529be9bfe12cfaf563ed28776503a13e1bbab7ceea8aa9bR197-R221)).

### DIFF
--- a/v03_pipeline/lib/misc/family_loading_failures.py
+++ b/v03_pipeline/lib/misc/family_loading_failures.py
@@ -16,7 +16,7 @@ def passes_relatedness_check(
     relatedness_check_lookup: dict[tuple[str, str], list],
     sample_id: str,
     other_id: str,
-    relation: Relation,
+    relations: list[Relation],
 ) -> tuple[bool, str | None]:
     # No relationship to check, return true
     if other_id is None:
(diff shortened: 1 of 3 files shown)


v03_pipeline/lib/misc/family_loading_failures.py
```diff
@@ -16,93 +16,64 @@ def passes_relatedness_check(
...
-            aunt_nephew_id,
-            Relation.AUNT_NEPHEW,
-        )
-        if not success:
-            failure_reasons.append(reason)
+        for other_id in relationship_set:
+            # Handle case where relation is identified in the
+            # pedigree as a "dummy" but is not included in
+            # the list of samples to load.
+            if other_id not in family.samples:
```
bpblanken, line 67: code duplication handled with better for-looping!

v03_pipeline/lib/misc/pedigree.py
```diff
@@ -8,17 +8,17 @@
 
 
 class Relation(Enum):
-    PARENT = 'parent'
-    GRANDPARENT = 'grandparent'
+    PARENT_CHILD = 'parent_child'
```
bpblanken, line 11: I renamed these for clarity to indicate the bi-directionality.

matren395 (commented):
really quick - can you explain the Test cases and what the tests are doing before I approve ?

bpblanken:
@matren395 

Yes!

So the root issue at play here is parents that are identified in the pedigree but that are not actually real samples present in the callset.  

We have one test [here](https://github.com/broadinstitute/seqr-loading-pipelines/blob/main/v03_pipeline/lib/misc/pedigree_test.py#L466) that ensures we do actually parse those samples as parents.  A previous iteration of this code _did not_... which is why that test exists.

The newly added test [here](https://github.com/broadinstitute/seqr-loading-pipelines/pull/1003/files#diff-380bf733d18b80ac2529be9bfe12cfaf563ed28776503a13e1bbab7ceea8aa9bR201-R210) checks that the mother ('sample_2') that is parsed from the pedigree does not actually fail the family even if the parent/child relationship is missing from the relatedness check table.

matren395:
okay! now that the tests make sense, lgtm then

matren395 (approved):
^^lgtm

PR #1006, opened 2025-01-06
feat: Add rsync task between hail search data/reference data stored in gcs and on local disk.  

bpblanken:
Note: does not actually run the task... the end of pipeline needs a bit of refactor which I will do downstream.


jklugherz (approved):
nice! this is very clear and well tested. I also like how you split the work into two PRs.

PR #1009, opened 2025-01-08
refactor + chore: shuffle end of pipeline steps to enable dataproc for local users

bpblanken:
Makes sure that the "rsync" runs after the "RunPipeline" step but before the "WriteSuccessFile"


v03_pipeline/lib/tasks/run_pipeline.py
```diff
@@ -0,0 +1,32 @@
...
+    UpdateVariantAnnotationsTableWithNewSamplesTask,
+)
+from v03_pipeline.lib.tasks.write_metadata_for_run import WriteMetadataForRunTask
+from v03_pipeline.lib.tasks.write_project_family_tables import (
+    WriteProjectFamilyTablesTask,
+)
+
+
+@luigi.util.inherits(BaseLoadingRunParams)
+class RunPipelineTask(luigi.WrapperTask):
```
bpblanken, line 17: luigi.WrapperTask is new but [sane](https://luigi.readthedocs.io/en/latest/luigi_patterns.html#triggering-many-tasks)

jklugherz: nice to know that this exists

jklugherz (approved):

PR #1012, opened 2025-01-10
feat: move liftover file into repo to remove dependency on gcs

bpblanken:
- Moves the liftover file into the repo (and the deployed package) for the v03 pipeline.
- Moves the liftover file to being locally downloaded with the `download_reference_data.sh` script for the v02 pipeline.
- Moves the v02 validation tables into the local download.
- Does not touch the v03 validation tables as they should be handled already.


docker/bin/load_data.sh
```diff
@@ -26,6 +26,11 @@ python3 -m seqr_loading SeqrMTToESTask --local-scheduler \
     --reference-ht-path "/seqr-reference-data/${FULL_BUILD_VERSION}/combined_reference_data_grch${BUILD_VERSION}.ht" \
     --clinvar-ht-path "/seqr-reference-data/${FULL_BUILD_VERSION}/clinvar.${FULL_BUILD_VERSION}.ht" \
     --vep-config-json-path "/vep_configs/vep-${FULL_BUILD_VERSION}-loftee.json" \
+    --grch38-to-grch37-ref-chain "/seqr-reference-data/grch38_to_grch37.over.chain.gz" \
+    --GlobalConfig-validation-37-noncoding-ht "/seqr-reference-data/GRCh37/common_noncoding_variants.grch37.ht" \
```
bpblanken, line 30: All 4 tables are getting passed as args here, even though only two are relevant to the load.

This should be [fine, given that we only hl.read_table the relevant tables](https://github.com/broadinstitute/seqr-loading-pipelines/blob/main/luigi_pipeline/lib/hail_tasks.py#L107-L112)

v03_pipeline/lib/model/constants.py
```diff
@@ -10,3 +12,11 @@
     'R0815_gregor_training_project_',
     'R0816_gregor_training_project_',
 }
+GRCH37_TO_GRCH38_LIFTOVER_REF_PATH = importlib.resources.path(
+    'v03_pipeline.var.liftover',
+    'grch37_to_grch38.over.chain.gz',
+).as_posix()
+GRCH38_TO_GRCH37_LIFTOVER_REF_PATH = importlib.resources.path(
```
bpblanken, line 19: On dataproc we don't have clean access to file paths that are stored in the repo, so we access the path as if it is a python package.

hanars: On dataproc shouldn't we be using the resources directly in gs?

bpblanken: oh yes...

this is equivalent to:

GRCH38_TO_GRCH37_LIFTOVER_REF_PATH = (
  'gs://hail-common/references/grch37_to_grch38.over.chain.gz'
  if FeatureFlag.RUN_PIPELINE_ON_DATAPROC else  'v03_pipeline/var/liftover/grch38_to_grch37.over.chain.gz'
)

I'm happy to swap to that if we think it's more clear.

hanars: yeah thats much more clear to me

PR #1019, opened 2025-01-15
bugfix bad error body check

bpblanken:


v03_pipeline/lib/tasks/write_validation_errors_for_run.py
```diff
@@ -50,20 +50,11 @@ def wrapper(self: luigi.Task):
         try:
             return f(self)
         except SeqrValidationError as e:
-            if isinstance(
-                e.args[1],
```
bpblanken, line 54: `e.args[1]` doesn't work if `args` is len == 1.  

I did the todo too.

hanars (approved):

PR #1029, opened 2025-01-31
bugfix: unsorted pedigree

bpblanken:


v03_pipeline/lib/misc/pedigree_test.py
```diff
@@ -490,3 +491,20 @@ def test_subsetted_pedigree_with_removed_parent(self) -> None:
...
+            iter(
+                [
+                    family
+                    for family in parsed_pedigree
+                    if family.family_guid == 'family_2_1'
+                ],
+            ),
+        )
+        self.assertTrue(
+            family.samples.keys() == {'RGP_164_1', 'RGP_164_2', 'RGP_164_4'},
```
matren395, line 509: do we use real samples for test samples elsewhere?

matren395 (commented):
other than my one comment, if the tests pass all fine then lgtm

v03_pipeline/lib/misc/pedigree_test.py
bpblanken: yeah, there's no real data or meaning associated with them, the strings are just the same.  

worth renaming though just for clarity.

PR #1030, opened 2025-02-05
feat: add support for over-writing genotypes for males on x non par

bpblanken:
Implementation of: 

Screenshot 2025-02-05 at 5 28 50 PM


v03_pipeline/lib/model/feature_flag.py
```diff
@@ -24,5 +27,6 @@ class FeatureFlag:
     EXPECT_TDR_METRICS: bool = EXPECT_TDR_METRICS
     EXPECT_WES_FILTERS: bool = EXPECT_WES_FILTERS
     INCLUDE_PIPELINE_VERSION_IN_PREFIX: bool = INCLUDE_PIPELINE_VERSION_IN_PREFIX
+    OVERWRITE_SV_MALE_NON_PAR_CALLS: bool = OVERWRITE_SV_MALE_NON_PAR_CALLS
```
bpblanken, line 30: My understanding of this logic is it will only apply for internal SV callsets.

v03_pipeline/lib/misc/sv.py
```diff
@@ -0,0 +1,51 @@
+import hail as hl
+
+from v03_pipeline.lib.annotations import sv
+from v03_pipeline.lib.misc.pedigree import Family
+from v03_pipeline.lib.model import ReferenceGenome, Sex
+
+
+def overwrite_male_non_par_calls(
```
bpblanken, line 8: @matren395 tag as resident actual biologist of this code.

matren395: not clear of the context, but - male X non-par would be regions on ChrX that are (biologically) haploid . reasonably confident these are relevant for a few diseases. It's here where all that cool sex-linked-trait-stuff happens

v03_pipeline/lib/misc/sv.py
```diff
@@ -0,0 +1,51 @@
...
+        [
+            i
+            for i in hl.get_reference(ReferenceGenome.GRCh38).par
+            if i.start.contig == ReferenceGenome.GRCh38.x_contig
+        ],
+    )
+    non_par_interval = hl.interval(
+        par_intervals[0].end,
+        par_intervals[1].start,
+    )
```
matren395, line 28: yep! PARs are at the start and the end of ChrX

v03_pipeline/lib/misc/sv.py
```diff
@@ -0,0 +1,51 @@
...
+                    hl.interval(
+                        mt.start_locus,
+                        mt.end_locus,
+                    ),
+                )
+            ),
+            hl.Call([1, 1], phased=False),
+            mt.GT,
+        ),
+    )
```
matren395, line 50: how would this handle SVs overlapping the start of a non-PAR regions, and do we want to be handling them this way ?

matren395: and what do the GTs look like elsewise? are they left haploid ?

matren395 (commented):
👁️ 👁️

matren395:
I guess the broader question is what _is_ that code ur implementing in the first place

v03_pipeline/lib/misc/sv.py
bpblanken: I believe `overlaps` handles cases where the end of the SV is > the start of the non-Par region.

v03_pipeline/lib/model/feature_flag.py
hanars: I don't know why we wouldn't apply this to all SVs in seqr, presumably we are doing this because of desired seqr search behavior. I don't think its rreasonable for us to write search queries that are able to handle branching logic for determining what a genotype for a sample is based on whether its internal or external data

v03_pipeline/lib/misc/sv.py
hanars: I think theres more than one par interval though?

bpblanken: I think there's two par intervals at the ends of the chromosome, but only one par region [end par1, start par2]

matren395: yep, there's two intervals - one at the start and one at the end of ChrX. If you take the space between the end of the first and the start of the second (what I believe this is doing) , you get the non-PAR portion of ChrX. 

all of my knowledge is from https://en.wikipedia.org/wiki/Pseudoautosomal_region

v03_pipeline/lib/misc/sv.py
```diff
@@ -0,0 +1,52 @@
...
+                male_sample_ids.contains(mt.s)
+                & non_par_interval.overlaps(
+                    hl.interval(
+                        mt.start_locus,
+                        mt.end_locus,
+                    ),
+                )
+                & mt.GT.is_het()
+            ),
+            hl.Call([1], phased=False),
```
bpblanken, line 48: I tweaked this to 1) only change het diploid calls and 2) replace with a haploid call.

v03_pipeline/lib/tasks/write_imported_callset.py
```diff
@@ -121,6 +122,10 @@ def create_table(self) -> hl.MatrixTable:
         if self.dataset_type.has_multi_allelic_variants:
             # NB: throws SeqrValidationError
             mt = split_multi_hts(mt, self.skip_validation)
+
+        if self.dataset_type.overwrite_male_non_par_calls:
```
bpblanken, line 126: this is no longer feature flagged and will run for all `SV`s.

v03_pipeline/lib/misc/sv.py
```diff
@@ -0,0 +1,52 @@
...
+    # fields are dropped by the time we format variants.
+    mt = mt.annotate_rows(
+        start_locus=sv.start_locus(mt),
+        end_locus=sv.end_locus(mt),
+    )
+    mt = mt.annotate_entries(
+        GT=hl.if_else(
+            (
+                male_sample_ids.contains(mt.s)
+                & non_par_interval.overlaps(
```
hanars, line 40: I don't think this is correct - if a single SV call overlaps both non-par and par regions I think we want to keep the diploid call because its real for the par region. I think we only want to change it if the SV is 100% overlapped by non-par region

bpblanken: hrm,  I think the current behavior is to "obtain those calls overlapping non-PAR regions and update the GT to 1/1 in males."

I think your logic to keep as diploid makes sense though, if we're ok with the behavior change.

hanars: I'm not like super confident in my knowledge of this data/biology, so it might be better to maintain parity

bpblanken: it might be worth asking lynn + alba on this.  I think we can get a fast answer.

PR #1031, opened 2025-02-06
gnomad svs

bpblanken:
Updates the gnomad sv field names.


hanars:
I don't think it is a good idea for the previously loaded SVs in seqr to be annotated with one version of gnomad and the new samples to be annotated with a different version. Are we planning to go back and update the annotations for everything we currently have?

bpblanken:
I don't think we have the capability to go back and update any of the old variants.  I know we, at some point, discussed how the gnomad annotations work and whether or not we should own them as part of our pipeline, and I believe the answer was to leave them as part of the WDL.  That leaves us in this situation though.

hanars:
This is a blocker for us accepting the data from the Talkowski team, as having a random mix of gnomad versions within a seqr project is an unacceptable end user experience. They either need to deliver us a callset that is annotated with the SAME annotations as past data, specifically using the same version of gnomad, or they need to deliver us an updated version of the previous callset with updated gnomad annotations

PR #1034, opened 2025-02-06
add sample qc task and filter_flags

jklugherz:
adds sample_qc luigi task with a single metric - filter_flags - that outputs a json file, makes it a dependency of the write callset task, and adds the sample qc json to the metadata json.


v03_pipeline/lib/methods/sample_qc.py
```diff
@@ -0,0 +1,31 @@
+import hail as hl
+from gnomad.sample_qc.pipeline import filter_rows_for_qc
+
+
+def call_sample_qc(
+    mt: hl.MatrixTable,
+):
+    mt = mt.annotate_entries(
```
jklugherz, line 8: this code might not be necessary, not sure what purpose it served in the original [script](https://github.com/broadinstitute/tgg_methods/blob/master/tgg/seqr_scripts/seqr_sample_qc.py#L349)

v03_pipeline/lib/paths.py
```diff
@@ -198,6 +198,22 @@ def relatedness_check_tsv_path(
     )
 
 
+def sample_qc_tsv_path(
```
bpblanken, line 201: might we be able to just output these metrics when we write the `metadata.json` for the run?

bpblanken: So just add a new blob to the json there.

v03_pipeline/lib/tasks/write_remapped_and_subsetted_callset.py
```diff
@@ -82,6 +83,20 @@ def requires(self) -> list[luigi.Task]:
                 self.clone(WriteRelatednessCheckTsvTask),
                 self.clone(WriteSexCheckTableTask),
             ]
+        # There is no functional dependency between `sample_qc` and this task,
```
bpblanken, line 86: 👍 this is right as is!  but there is a functional dependency between the tdr metrics and _some_ of the sample qc data, which we haven't gotten yet.

v03_pipeline/lib/paths.py
bpblanken: (the old-style seqr upload isn't a thing we're going to support any more, we'll need to re-write the ingestion of the new json!)

hanars: agreed that we want the results of the QC in the `metadata.json`

jklugherz: Do we want to do the sample qc work as a separate task at the same time as s+r checks and either
- a. write the output to a file or
- b. write the output to the callset mt globals?
We could also do it inside of the `WriteMetadataForRunTask` but that feels wrong b/c if qc fails somehow it should fail earlier.

bpblanken: Yeah, let's do it at the same time as the s+r checks.  That's a reasonable place to put it, as it's for sure downstream of the tdr metrics.  

The bit inside the `WriteMetadataForRunTask` should likely just be reading the output of the upstream task and writing it to the json.  We won't actually fail any families or samples due to the metrics we're computing and collecting.

jklugherz: yup so I kept the task that outputs the sample qc results to a tsv file and then modified `WriteMetadataForRunTask` to add it to the json

v03_pipeline/lib/methods/sample_qc.py
```diff
@@ -0,0 +1,86 @@
...
+    )
+    return mt.drop(
+        'contamination_rate',
+        'percent_bases_at_20x',
+        'mean_coverage',
+        'filtered_callrate',
+    )
+
+
+def sample_qc_tsv_to_dict(tsv_file_path: str) -> dict:
```
bpblanken, line 77: Is there a strong reason to even have the tsv format?  Is it possible to just use json all the way through?

jklugherz: changed it to json!

v03_pipeline/lib/tasks/write_sample_qc_json_test.py
```diff
@@ -0,0 +1,69 @@
...
+from v03_pipeline.lib.test.mock_complete_task import MockCompleteTask
+from v03_pipeline.lib.test.mocked_dataroot_testcase import MockedDatarootTestCase
+
+TEST_VCF = 'v03_pipeline/var/test/callsets/1kg_30variants.vcf'
+TEST_TDR_METRICS_FILE = 'v03_pipeline/var/test/tdr_metrics.tsv'
+TEST_RUN_ID = 'manual__2024-04-03'
+
+
+class WriteSampleQCJsonTaskTest(MockedDatarootTestCase):
+    @patch('v03_pipeline.lib.tasks.write_sample_qc_json.WriteTDRMetricsFilesTask')
```
jklugherz, line 18: I was not able to figure out the bigquery function mock contamination that was happening when this test was run with `WriteSexCheckTableTaskTest`, so I just mocked the entire `WriteTDRMetricsFilesTask` 🤷

v03_pipeline/lib/methods/sample_qc.py
```diff
@@ -0,0 +1,69 @@
...
+WES_COVERAGE_LOW_THRESHOLD = 85
+WGS_CALLRATE_LOW_THRESHOLD = 30
+
+
+def call_sample_qc(
+    mt: hl.MatrixTable,
+    tdr_metrics_ht: hl.Table,
+    sample_type: SampleType,
+):
+    mt = mt.annotate_cols(sample_type=sample_type)
```
jklugherz, line 17: noting to myself that this is unnecessary because sample_type is already a top level field in the metadata!

v03_pipeline/lib/tasks/write_sample_qc_json.py
```diff
@@ -0,0 +1,60 @@
...
+        callset_mt = call_sample_qc(
+            callset_mt,
+            tdr_metrics_ht,
+            self.sample_type,
+        )
+        ht = callset_mt.cols()
+        sample_qc_dict = defaultdict(dict)
+        for row in ht.flatten().collect():
+            r = dict(row)
+            sample_id = r.pop('s')
```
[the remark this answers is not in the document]
jklugherz: idt this is possible `RuntimeError: dictionary changed size during iteration`

v03_pipeline/lib/methods/sample_qc.py
```diff
@@ -0,0 +1,69 @@
...
+    if sample_type == SampleType.WES:
+        flags['coverage'] = mt.percent_bases_at_20x < WES_COVERAGE_LOW_THRESHOLD
+    else:
+        flags['coverage'] = mt.mean_coverage < WGS_CALLRATE_LOW_THRESHOLD
+
+    return mt.annotate_cols(
+        filter_flags=hl.array(
+            [hl.or_missing(filter_cond, name) for name, filter_cond in flags.items()],
+        ).filter(hl.is_defined),
+    ).drop(
```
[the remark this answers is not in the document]
jklugherz: just realized these shouldn't actually be dropped because seqr needs the value of the metric used to calculate the flag.

bpblanken (approved):
lgtm!  thank you!

v03_pipeline/lib/tasks/write_sample_qc_json.py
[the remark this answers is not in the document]
bpblanken: 👍

PR #1041, opened 2025-02-14
feat: add gnomad_svs as reference dataset.

bpblanken:


hanars (approved):
looks good! To clarify for myself, this will update gnomad for all new calls, and there will need to be a manual step to update old calls, correct?

PR #1042, opened 2025-02-14
tdr updates for sample qc

jklugherz:
tiny PR. updates the existing test that uses the TDR so that the mock biguery rows look most like the actual response.


jklugherz:
> @matren395 can you also look at these?

The TDR metrics are used [here](https://github.com/broadinstitute/seqr-loading-pipelines/pull/1043/files#diff-4cb2bb53c1427d1e20f14173c921a92094c2a951044174fcecada7b23da247d2R51)

PR #1049, opened 2025-02-23
bugfix/improvement: handle duplicate seqr internal truth vid

bpblanken:


v03_pipeline/lib/misc/sv.py
```diff
@@ -1,9 +1,112 @@
...
+    )
+
+    # Then, collect into memory the necessary existing variants & the new variants
+    annotations_ht = annotations_ht.select('end_locus')
+    existing_variants = {
+        v.variant_id: v
+        for v in (
+            annotations_ht.filter(
+                duplicate_internal_variant_ids.contains(annotations_ht.variant_id),
+            ).collect()
```
hanars, line 59: it feel like you are doing a lot of this outside of hail when it would be better to do it inside of hail. I feel like you could create a ht of the duplicate rows of the new callset, inner join with the relevant columns from the existing annotations table and then do this computation inside a single hail aggregation

bpblanken: Yes, the choice to do it outside hail was intentional, based on the assumption that the number of duplicate variants would be small in perpetuity, would fit in local memory, and would be easier to manage in native python than in hail.  This is loosely based on the opinion (and not fact) that hail isn’t as useful for SVs as it is for SNV_INDELs and that we’re more maintainable and performant without the shuffles hail would induce.

v03_pipeline/lib/misc/callsets.py
```diff
@@ -74,4 +79,26 @@ def get_additional_row_fields(
...
+            and hl.eval(
+                hl.len(
+                    hl.read_table(
+                        variant_annotations_table_path(
+                            reference_genome,
+                            dataset_type,
+                        ),
+                    ).globals.updates,
+                )
+                > 0,
```
jklugherz, line 100: why is this >0 check here?

PR #1047, opened 2025-02-18
feat: compute running SV allele frequencies 

bpblanken:

PR #1050, opened 2025-02-26
Dev

jklugherz:


v03_pipeline/migrations/annotations/0004_add_gnomad_svs.py
```diff
@@ -0,0 +1,52 @@
...
+                'info.GNOMAD_V4.1_TRUTH_VID': mapping_ht[ht.key].info[
+                    'GNOMAD_V4.1_TRUTH_VID'
+                ],
+            },
+        )
+        gnomad_svs_ht = ReferenceDataset.gnomad_svs.get_ht(ReferenceGenome.GRCh38)
+        ht = ht.annotate(gnomad_svs=sv.gnomad_svs(ht, gnomad_svs_ht))
+        ht = ht.drop('info.GNOMAD_V4.1_TRUTH_VID')
+        return ht.annotate_globals(
+            versions=ht.globals.versions.annotate(gnomad_svs='1.0'),
```
hanars, line 50: wouldn't this already be the version- shouldn't we bump it?

v03_pipeline/migrations/annotations/0004_add_gnomad_svs.py
```diff
@@ -0,0 +1,52 @@
...
+            enums=ht.globals.enums.annotate(gnomad_svs=hl.Struct()),
```
hanars, line 51: this feels unnecessary, as this would already be in the existing table globals and does not seem to be changing

v03_pipeline/migrations/annotations/0004_add_gnomad_svs.py
jklugherz: you're right, the globals annotation happens when we call `ReferenceDataset.gnomad.get_ht`

v03_pipeline/migrations/annotations/0004_add_gnomad_svs.py
jklugherz: https://github.com/broadinstitute/seqr-loading-pipelines/pull/1050/files/c80c04bbaaa818d10f5888d866b9c72d85cca931#r1978158117 answers this

hanars (approved):

v03_pipeline/migrations/annotations/0004_add_gnomad_svs.py
bpblanken: oooh, there's some confusing bits here.  

`ReferenceDataset.gnomad_svs.get_ht` sets the globals on the `gnomad_svs` reference dataset (`gs://seqr-reference-data/v3.1/GRCh38/gnomad_svs/1.0.ht` if persisted).  This code was setting the globals on the SV annotations table as they're not currently there.  It's not overwhelmingly important, as [we exclude gnomad_svs](https://github.com/broadinstitute/seqr-loading-pipelines/pull/1047/files#diff-7e90c9a90492a35bcb85099e7b3b7746f84262caf2777e7655fb1d89c62fa561R41) from consideration when computing the reference datasets to update, and it would also be annotated by the normal process because I kept separate `for_reference_genome_dataset_type_annotations` (which includes `gnomad_svs`) and `for_reference_genome_dataset_type_annotations_updates` methods (which does not).   I felt the migration should include the correct globals on annotations table for completeness though!

jklugherz: right I got jumbled, this is an `annotation` table migration. 

> `This code was setting the globals on the SV annotations table as they're not currently there`
This is good reason to keep the globals annotation in this migration. @hanars, I'm going to ad the globals `versions` and `enums` annotations back and re-request review.

bpblanken: yeah, it would be awesome if these had tests, but not entirely sure it would have caught this 😬.

PR #1052, opened 2025-03-04
bug fix: don't drop sv locus/alleles fields so validation can run

jklugherz:
While qc'ing gnomad_svs, I noticed that `ReferenceDataset.gnomad_svs.get_ht(ReferenceGenome.GRCh38)` would error out during the allele validation step because `alleles` does not exist on the `ht` - it's dropped [here](https://github.com/broadinstitute/seqr-loading-pipelines/blob/main/v03_pipeline/lib/reference_datasets/gnomad_svs.py#L18). 

opted to remove this line and perform locus and allele validation on the gnomad_svs reference dataset.


v03_pipeline/lib/misc/validation.py
```diff
@@ -31,6 +31,9 @@ def validate_allele_type(
     **_: Any,
 ) -> None:
     ht = t.rows() if isinstance(t, hl.MatrixTable) else t
+    if not hasattr(ht, 'alleles'):
```
bpblanken, line 34: I think we should still validate alleles for these.  There's actually a well defined schema for SV symbolic alleles and might as well check that we're complying.  

Given we also need `locus` for [this block](https://github.com/broadinstitute/seqr-loading-pipelines/blame/main/v03_pipeline/lib/reference_datasets/misc.py#L86-L90) to work correctly, what would be the lift on removing the line that drops `locus`, `alleles` and just leaving it on the table?  We'll need logic to drop them when we [do this](https://github.com/broadinstitute/seqr-loading-pipelines/blob/main/v03_pipeline/lib/annotations/sv.py#L183) though.

The reason for dropping was mostly around consistency... they don't _need_ to be there because we don't consider them to be part of the key.

bpblanken: Oof, `validate_allele_type` has a hardcoded `DatasetType.SNV_INDEL`.  I think we want to fix that... I'll do it real quick!

jklugherz: > removing the line that drops locus, alleles and just leaving it on the table

I'm doing this

bpblanken (approved):

PR #1053, opened 2025-03-05
bug fix: handle set of dataset types during allele type validation

bpblanken:


v03_pipeline/lib/reference_datasets/reference_dataset.py
```diff
@@ -143,10 +149,8 @@ def get_ht(
         if enum_selects:
             ht = ht.transmute(**enum_selects)
         ht = filter_contigs(ht, reference_genome)
-        # Reference Datasets are DatasetType agnostic, but these
-        # methods (in theory) support SV/GCNV.  SNV_INDEL
-        # is passed as a proxy for non-SV/GCNV.
-        validate_allele_type(ht, DatasetType.SNV_INDEL)
+        for dataset_type in self.dataset_types(reference_genome):
```
jklugherz, line 152: I was also thinking that this fn `get_ht` is called in `UpdatedReferenceDatasetTask` which has a `dataset_type` we could pass to this function.. but then that violates `Reference Datasets are DatasetType agnostic` so I think what you did here is better.

bpblanken: Yeah, I took two stabs at this and this was the cleanest.  Reference datasets support multiple dataset types, so we should validate them all.

v03_pipeline/lib/reference_datasets/reference_dataset.py
```diff
@@ -143,10 +149,8 @@ def get_ht(
...
+            validate_allele_type(ht, dataset_type)
         validate_no_duplicate_variants(ht, reference_genome, DatasetType.SNV_INDEL)
```
jklugherz, line 154: can we put this line inside of the loop and pass `dataset_type` here too

bpblanken: Yes good call!

PR #1057, opened 2025-03-13
move ploidy checks from callset integrity validation to family/sample validation

jklugherz:


v03_pipeline/lib/misc/family_loading_failures.py
```diff
@@ -174,3 +174,44 @@ def get_families_failed_sex_check(
...
+                ).contains(sex_check_ht[mt.s].predicted_sex)
+            )
+        ),
+    )
+    discrepant_samples = mt.aggregate_cols(
+        hl.agg.filter(mt.discrepant, hl.agg.collect_as_set(mt.s)),
+    )
+    failed_families = {}
+    for family in families:
+        discrepant_loadable_samples = set(family.samples.keys()) & discrepant_samples
```
jklugherz, line 211: only real change here is to use the passed `families` to identify failures we care about

v03_pipeline/lib/misc/family_loading_failures_test.py
```diff
@@ -250,3 +252,189 @@ def test_get_families_failed_sex_check(self):
...
+        )
+        self.assertCountEqual(
+            failed_families.values(),
+            [
+                "Found samples with misaligned ploidy with their provided imputed sex (first 10, if applicable) : ['HG00731_1', 'HG00732_1', 'NA20889_1', 'NA20899_1']",
+            ],
+        )
+
+        # Invalid X chromosome case, but only discrepant family samples are reported
+        families = {
```
jklugherz, line 424: relevant change to the tests is the addition of this case

v03_pipeline/lib/tasks/write_remapped_and_subsetted_callset_test.py
```diff
@@ -180,6 +178,24 @@ def test_write_remapped_and_subsetted_callset_task_failed_sex_check_family(
                                 'samples': ['NA20885_1'],
                             },
                         },
+                        ploidy_check={
```
jklugherz, line 181: These failures are new, but I thought it safer to incorporate them into the test case rather than trying to adjust the underlying callset or sex check ht, because those are used in other tests.

v03_pipeline/lib/misc/family_loading_failures.py
bpblanken: 👍

v03_pipeline/lib/tasks/write_remapped_and_subsetted_callset_test.py
```diff
@@ -337,6 +353,7 @@ def test_write_remapped_and_subsetted_callset_task_all_families_failed(
                                 ],
                             },
                         },
+                        'ploidy_check': {},
```
bpblanken, line 356: Do we need to change the code that reads this json to account for the ploidy failures?

v03_pipeline/lib/tasks/write_remapped_and_subsetted_callset.py
```diff
@@ -134,12 +136,21 @@ def create_table(self) -> hl.MatrixTable:
                 sex_check_ht,
                 remap_lookup,
             )
+            families_failed_imputed_sex_ploidy = get_families_failed_imputed_sex_ploidy(
```
bpblanken, line 139: I think we should run this _before_ sex check, and also exclude the failures from the sex check!

jklugherz: I made this change

v03_pipeline/lib/tasks/write_remapped_and_subsetted_callset_test.py
```diff
@@ -221,120 +237,82 @@ def test_write_remapped_and_subsetted_callset_task_all_families_failed(
...
+                                'samples': ['NA20885_1'],
                                 'reasons': [
                                     'Sample NA20885_1 has pedigree sex M but imputed sex F',
                                 ],
                             },
+                        },
+                        'ploidy_check': {
+                            'cde_1': {
+                                'samples': ['NA20881_1'],
+                                'reasons': "Found samples with misaligned ploidy with their provided imputed sex (first 10, if applicable) : ['NA20881_1']",
```
bpblanken, line 303: the `(first 10, if applicable)` looks weird to me if these are only going to be one sample.  Do we need to change the string error message?

jklugherz: This message used to be part of a `SeqrValidationError` message, so maybe that's why it was kept short. We can take out the `(first 10, if applicable)` and just include all of the failed samples in this message.

bpblanken: Would it be better to format the `reasons` as a list like the [sex failures](https://github.com/broadinstitute/seqr-loading-pipelines/pull/1057/commits/b583a26656a2a70720f9f5a4e5a1a72201cfc186#diff-81bda30e9edc96dbaf6b9533d4db17f8afa17df35de4b37c3c05e1856e6779dbL173-L175).  So one string reason for each sample, rather than a list of samples embedded in a single "reason".

hanars: In the check sample job in seqr, we parse the structured json from the `failed_family_samples` and then use that information to report a human readable error but also to update the status for family models in seqr, update airtable etc. We have explicit parsing in this part of the code that takes that error dictionary and makes the error list we report in slack clear
For validation failures, we just take the `error_messages` and pass them through to the end user with no additional processing. So I think we need to think about formatting this message very differently than for sex checks

bpblanken: This block is under `failed_family_samples` now, and won't be considered a `validation` failure.  We decided that this check should  no longer be considered a callset-wide validation, but run as a check over the families to load.  We have a new key ("ploidy_check") for these though, which _maybe_ needs an update to the check new samples job (reading the code though, it seems fine as is!).

jklugherz: Yep I also took a look at the check samples job code and the `ploidy_check` will be incorporated into the `failed_family_samples` slack message, requiring no code change.

bpblanken (approved):

Issue #1056, opened 2025-03-11
DRAGEN/WES/S&R Ploidy Placeholder.


bpblanken:
- Exomes are being called with a different (?) variant caller version, so sex isn't imputed.
- No imputed sex means all calls will be diploid.
- Our plan is to use the ol' reliable hl.impute_sex on these.

lynnpais:
Done

PR #1059, opened 2025-03-19
Re-enable `hl.impute_sex` for WES samples.

bpblanken:
Fixes https://github.com/broadinstitute/seqr-loading-pipelines/issues/1056


v03_pipeline/lib/model/dataset_type.py
```diff
@@ -188,9 +188,12 @@ def expect_filters(
     def expect_tdr_metrics(
         self,
         reference_genome: ReferenceGenome,
+        sample_type: SampleType,
     ) -> bool:
         return (
-            self == DatasetType.SNV_INDEL and reference_genome == ReferenceGenome.GRCh38
+            self == DatasetType.SNV_INDEL
+            and reference_genome == ReferenceGenome.GRCh38
+            and sample_type == SampleType.WGS
```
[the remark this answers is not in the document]
bpblanken: I think this is actually wrong as is.  WES projects do have metrics.  I'm fixing.

v03_pipeline/lib/tasks/write_imported_callset.py
```diff
@@ -60,17 +59,6 @@ def requires(self) -> list[luigi.Task]:
...
-        if (
-            FeatureFlag.EXPECT_TDR_METRICS
-            and not self.skip_expect_tdr_metrics
-            and self.dataset_type.expect_tdr_metrics(
-                self.reference_genome,
-            )
-        ):
-            requirements = [
-                *requirements,
-                self.clone(WriteTDRMetricsFilesTask),
```
bpblanken, line 72: This dependency additionally lives under `WriteSexCheckTableTask` and  `WriteSampleQCJsonTask`, so not necessary here.

v03_pipeline/lib/methods/sex_check.py
```diff
@@ -0,0 +1,49 @@
...
+        mt.GT,
+        male_threshold=XY_FSTAT_THRESHOLD,
+        female_threshold=XX_FSTAT_THRESHOLD,
+        aaf_threshold=AAF_THRESHOLD,
+    )
+    ht = mt.annotate_cols(**impute_sex_ht[mt.col_key]).cols()
+    ht = ht.select(
+        predicted_sex=(
+            hl.case()
+            .when(hl.is_missing(ht.is_female), Sex.UNKNOWN.value)
```
bpblanken, line 38: Note that we don't handle aneuploidy here, but a sample identified as aneuploidy by the algorithm _should/might_ be labeled as UNKNOWN, excluding it from the family failures downstream.

jklugherz (approved):
goooood stuff

v03_pipeline/lib/methods/sex_check.py
```diff
@@ -0,0 +1,49 @@
...
+    mt = mt.filter_rows(
+        (hl.len(mt.alleles) == BIALLELIC) & hl.is_snp(mt.alleles[0], mt.alleles[1]),
+    )
+
+    # Filter to PASS variants only (variants with empty or missing filter set)
+    mt = mt.filter_rows(
+        hl.is_missing(mt.filters) | (mt.filters.length() == 0),
+        keep=True,
+    )
+    impute_sex_ht = hl.impute_sex(
```
matren395, line 28: we encountered a horror case in the BGE callset, where a sample with sufficiently horrible coverage was 1) imputed as WGS, and as such was 2) given haploid reads on sex chromosomes. 

again, a horror case, but I would like to prepare for this. some step filtering to only diploid samples would be sufficient .

v03_pipeline/lib/methods/sex_check.py
```diff
@@ -0,0 +1,49 @@
...
+)
+
+
+def compute_sex_check_ht(mt: hl.MatrixTable) -> hl.Table:
+    # Filter to SNVs and biallelics
+    # NB: We should already have filtered biallelics, but just in case.
+    mt = mt.filter_rows(
+        (hl.len(mt.alleles) == BIALLELIC) & hl.is_snp(mt.alleles[0], mt.alleles[1]),
+    )
+
```
matren395, line 22: this will NOT be perfect i don't think

```suggestion
mt = mt.filter_cols(
     hl.agg.all(
          mt.GT.is_diploid() | hl.is_missing(mt.GT)
          )
     )
)

```

matren395 (commented):
some edge case i want to cover

v03_pipeline/lib/methods/sex_check.py
bpblanken: ^ wouldn't this exclude samples that were not diploid?  wouldn't we want the callset to fail if there were inadvertent haploid calls?

v03_pipeline/lib/methods/sex_check.py
bpblanken: ah, I read your comments line-wise rather than time-wise.  code looks right to me.

v03_pipeline/lib/methods/sex_check.py
matren395: this would! which would just have excluded that one bad sample. 

Ben does this run on all samples in a callset, or after filtering to samples in the AirTable ?

PR #1062, opened 2025-03-21
flip ordering of struct

bpblanken:


v03_pipeline/lib/annotations/sv.py
```diff
@@ -204,9 +204,9 @@ def _safe_gt_stats_fetch(ht: hl.Table, field: str):
         row.gt_stats.AC[1] - (row.gt_stats.homozygote_count[1] * 2)
     ) + _safe_gt_stats_fetch(ht, 'Het')
     return hl.struct(
+        AF=hl.float32(AC / AN),
```
bpblanken, line 207: current table has this field first, tests didn't catch this because they build the initial table from this function :/ .

hanars (approved):

PR #1064, opened 2025-03-24
migration gnomad sv n_hom

bpblanken:


v03_pipeline/migrations/annotations/0005_modify_gnomad_sv_hom.py
```diff
@@ -0,0 +1,31 @@
+import hail as hl
+
+from v03_pipeline.lib.migration.base_migration import BaseMigration
+from v03_pipeline.lib.model import DatasetType, ReferenceGenome
+
+GNOMAD_SVS_HT = 'gs://seqr-reference-data/v3.1/GRCh38/gnomad_svs/1.1.ht'
```
bpblanken, line 6: Made this manually!

hanars (approved):

v03_pipeline/migrations/annotations/0005_modify_gnomad_sv_hom.py
bpblanken: for context, if I ran the pipeline normally, the 1.1 ht would be generated.  However, we have special logic that excludes gnomad_svs from the generic "update annotations table with updated reference data" flow because the key on the reference dataset is the gnomad_sv id rather than the seqr internal structural variant id.

PR #1069, opened 2025-04-02
bugfix:fix incorrect union-ing method of matrix tables.

bpblanken:
this went uncovered in the tests because we don't have a test over multiple SV projects inside a single callset.  Added that test!

### DIFF
--- a/v03_pipeline/lib/misc/callsets_test.py
+++ b/v03_pipeline/lib/misc/callsets_test.py
@@ -1,4 +1,5 @@
 import unittest
+
 import hail as hl
 
 from v03_pipeline.lib.misc.callsets import union_callset_mts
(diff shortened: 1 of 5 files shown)


v03_pipeline/lib/misc/callsets_test.py
```diff
@@ -0,0 +1,87 @@
...
+                    id=1,
+                    row_field=2,
+                    s='sample_1',
+                    GT=hl.Call(alleles=[0, 0], phased=False),
+                ),
+                hl.Struct(id=1, row_field=2, s='sample_2', GT=None),
+                hl.Struct(id=2, row_field=None, s='sample_1', GT=None),
+                hl.Struct(
+                    id=2,
+                    row_field=None,
```
matren395, line 82: super clarifying that the expected behavior here is to have _no_ row information for any row_fields past the first matrix table ? 

I'd also want to note how it would perform on more than two projects , if you think the juice would be worth the squeeze on that

bpblanken: Yes, that is correct.  The row field that is only present on the "right" matrix table ("second_row_field" in the test I made), is dropped.  `drop_right_row_fields` is configurable, and I left it as the default option of True.  

I did consider adding a validation line to ensure that every mt has the same row fields, but thought it wasn't explicitly necessary in this case.  I could be wrong though.

PR #1075, opened 2025-04-09
Remap from ped file

hanars:


v03_pipeline/lib/misc/io.py
```diff
@@ -283,31 +283,22 @@ def import_tdr_qc_metrics(file_path: str) -> hl.Table:
...
+    optional_selects = {'remap_id': ht.VCF_ID} if has_remap else {}
     return ht.select(
         sex=ht.Sex,
         family_guid=ht.Family_GUID,
         s=ht.Individual_ID,
         maternal_s=ht.Maternal_ID,
         paternal_s=ht.Paternal_ID,
-    )
+        **optional_selects,
+    ), has_remap
```
bpblanken, line 297: could we put this bool on the globals?  or, better yet, just check if `remap_id` exists as an attribute on the hail table object?

bpblanken:
One small nit but this is definitely the cleanest and least invasive approach of now!

PR #1076, opened 2025-04-10
feat: add tasks for writing parquet tables.  

bpblanken:
1) adds methods to de-enumerate our enums back to strings, but preserve the annotations table structure.  
2) adds methods to support camelCasing of structs.
3) adds exports of `new_variants`, `new_entries`, `new_clinvar_variants` and `new_transcripts` parquets.


v03_pipeline/lib/tasks/exports/misc.py
```diff
@@ -0,0 +1,305 @@
...
+
+def unmap_formatting_annotation_enums(
+    ht: hl.Table,
+    reference_genome: ReferenceGenome,
+    dataset_type: DatasetType,
+) -> hl.Table:
+    formatting_annotation_names = {
+        fa.__name__ for fa in dataset_type.formatting_annotation_fns(reference_genome)
+    }
+    if 'sorted_motif_feature_consequences' in formatting_annotation_names:
```
hanars, line 208: Is there a reason why the generic part of the unmapping function you wrote for `unmap_reference_dataset_annotation_enums` wouldn't work here?

v03_pipeline/lib/tasks/exports/write_new_clinvar_variants_parquet.py
```diff
@@ -0,0 +1,58 @@
...
+)
+from v03_pipeline.lib.tasks.files import GCSorLocalTarget
+from v03_pipeline.lib.tasks.update_new_variants_with_caids import (
+    UpdateNewVariantsWithCAIDsTask,
+)
+from v03_pipeline.lib.tasks.write_new_variants_table import WriteNewVariantsTableTask
+
+
+@luigi.util.inherits(BaseLoadingRunParams)
+class WriteNewClinvarVariantsParquetTask(BaseWriteParquetTask):
```
hanars, line 24: just confirming that, per our discussion, this is not the approach we are going to be using for MVP. Okay to leave in for now and take out later when the real approach is implemented

v03_pipeline/lib/tasks/exports/write_new_entries_parquet.py
```diff
@@ -0,0 +1,145 @@
...
+
+    def create_table(self) -> None:
+        unioned_ht = None
+        for project_guid, remapped_and_subsetted_callset_task in zip(
+            self.project_guids,
+            self.input()[REMAPPED_AND_SUBSETTED_CALLSET_TASKS],
+            strict=True,
+        ):
+            mt = hl.read_matrix_table(remapped_and_subsetted_callset_task.path)
+            ht = compute_callset_family_entries_ht(
```
hanars, line 84: isn't this already called in the main loading task we run? Why not make this task dependent on that task and reuse this table?

v03_pipeline/lib/tasks/exports/write_new_entries_parquet.py
```diff
@@ -0,0 +1,145 @@
...
+                        ),
+                    ),
+                ),
+            )
+            annotations_ht = hl.read_table(
+                self.input()[ANNOTATIONS_TABLE_TASK].path,
+            )
+            ht = ht.join(annotations_ht)
+
+            if self.input().get(HIGH_AF_VARIANTS_TABLE_TASK):
```
hanars, line 108: I feel like we will want this column for GRCh37 as well, given that in the clickhosue backend this column is designed to reduce the search space in a proband-only family, which will be the same order of magnitude across genome builds. But we can test that out and build it if necessary when we get to that part of the implementation

v03_pipeline/lib/tasks/exports/misc.py
bpblanken: yeah, the logic behind these being manual was we enumerated them manually, both in the annotations and in the globals.  Building this function was effectively just reversing an existing function.  I thought at the time that it was easier to just be explicit.

v03_pipeline/lib/tasks/exports/write_new_entries_parquet.py
bpblanken: I made the decision to use the subsetted tables as the source rather than the project tables because we didn't have a clean existing way to select families out of the project table (the existing method removes families, and our migration plan will use the entire project tables with no need to subset).  I thought it made more sense to use this for now as I don't actually think the compute difference will be noticeable, and it will more easily allow us to deprecate the project table.

hanars: My only concern is that in addition to supporting this task for new projects/families, we will also need to be able to programmatically migrate all data from the existing project tables. If we write a task here that uses that project table it would be very easy to reuse it to do the migration. However, if you think that using the subsetting is easier for the loading and would rather write and support a totally independent task for the migration thats fine too

bpblanken: yeah, the plan was to write a totally independent task for the migration, as the two would be functionally different.

hanars (approved):

PR #1083, opened 2025-04-17
bugfix: filter out null clinvar conditions.

bpblanken:
Noticed this while building schema for clinvar... this field requires Nulls, which we don't want.


v03_pipeline/lib/reference_datasets/clinvar.py
```diff
@@ -154,10 +154,13 @@ def select_fields(ht):
         conflictingPathogenicities=parsed_and_mapped_clnsigconf(ht),
         goldStars=CLINVAR_GOLD_STARS_LOOKUP.get(hl.delimit(ht.info.CLNREVSTAT)),
         submitters=ht.submitters,
-        # assumes the format 'MedGen#:condition', e.g.'C0023264:Leigh syndrome'
-        conditions=hl.map(
-            lambda p: p.split(r':')[1],
-            ht.conditions,
+        # assumes the format 'MedGen#:condition;MedGen#:condition', e.g.'C0023264:Leigh syndrome'
+        conditions=hl.filter(
```
bpblanken, line 158: Note this changes the one-to-one alignment between submitters and conditions, but I don't think we're doing anything that pairing.  It's already one-submitter-to-many-conditions in practice, we were just pulling the first one.

hanars (approved):

PR #1086, opened 2025-04-22
Benb/add key to pipeline

bpblanken:
Resolves https://github.com/broadinstitute/seqr-private/issues/1633


v03_pipeline/lib/tasks/write_new_variants_table.py
```diff
@@ -192,6 +192,12 @@ def create_table(self) -> hl.Table:
                 },
             )
             new_variants_ht = new_variants_ht.join(reference_dataset_ht, 'left')
+
+        # Add serial integer index
+        new_variants_ht = new_variants_ht.add_index(name='key_')
+        new_variants_ht = new_variants_ht.transmute(
+            key_=new_variants_ht.key_ + annotations_ht.index_globals().max_key_ + 1,
```
bpblanken, line 199: Idea here is to use a serially increasing counter to ensure that the assigned keys are always increasing.  Though we might not support deleting variants in the future, we currently do have flows that will result in variants being removed!

hanars (approved):

PR #1087, opened 2025-04-23
Misc pipeline fixups.

bpblanken:
Addresses both WES imputed sex & filtering sample qc to loadable families.


v03_pipeline/lib/tasks/write_metadata_for_run.py
```diff
@@ -84,13 +84,20 @@ def run(self) -> None:
                 self.reference_genome,
             )
         ):
+            loadable_samples = {
+                sample
+                for family_samples in metadata_json['family_samples'].values()
+                for sample in family_samples
```
hanars, line 90: when we did the seqr job we configured it to also include families included in failed family samples so we should include those here as well

hanars (approved):

PR #1091, opened 2025-04-28
chore: add a filter to the tests

bpblanken:
Adding this was useful when testing the parquet export.


v03_pipeline/lib/misc/family_entries.py
```diff
@@ -1,6 +1,30 @@
 import hail as hl
 
-from v03_pipeline.lib.model import DatasetType
+from v03_pipeline.lib.model import DatasetType, ReferenceGenome
+
+
+def initialize_project_table(
```
bpblanken, line 6: this function was present twice and made sense to share the logic.

PR #1093, opened 2025-04-30
feat: clickhouse loading service shim 

bpblanken:
This is a first pass to get a running service with connections to clickhouse and gcs working!  

Idea is to try to run this as a sidecar to the clickhouse pod.


v03_pipeline/lib/misc/clickhouse.py
```diff
@@ -0,0 +1,12 @@
+from clickhouse_driver import Client
+
+from v03_pipeline.lib.model.environment import Env
+
+
+def get_clickhouse_client() -> Client:
+    return Client(
+        host=Env.CLICKHOUSE_SERVICE_HOSTNAME,
+        port=Env.CLICKHOUSE_SERVICE_PORT,
+        user=Env.CLICKHOUSE_USER,
```
bpblanken, line 10: the seqr app is using `CLICKHOUSE_USERNAME`, but I think `CLICKHOUSE_USER` is the value expected by the service if used.

v03_pipeline/bin/clickhouse_loader.py
```diff
@@ -0,0 +1,57 @@
...
+    drop_staging_tables(client)
+    sys.exit(0)
+
+
+signal.signal(signal.SIGINT, signal_handler)
+signal.signal(signal.SIGTERM, signal_handler)
+
+
+@retry(tries=3, delay=5)
+def drop_staging_tables(client):
```
bpblanken, line 29: retry and signal handler ideas are inherited from the clickhouse loading reference script.

PR #1094, opened 2025-05-02
Benb/add support for direct insert tables

bpblanken:


v03_pipeline/var/test-clickhouse.xml
```diff
@@ -0,0 +1,53 @@
+<clickhouse>
```
bpblanken, line 1: This is the config for running a local instance of ClickHouse while running the unit tests, with support for loading arbitrary file paths.  It'll need documentation at some point.

v03_pipeline/lib/misc/clickhouse.py
```diff
@@ -1,6 +1,126 @@
...
+    BaseReferenceDataset,
+    ReferenceDataset,
+)
+
+logger = get_logger(__name__)
+
+GOOGLE_XML_API_PATH = 'https://storage.googleapis.com/'
+
+
+class ClickhouseTable(StrEnum):
```
bpblanken, line 25: `entries` will be added on a subsequent PR, the flow will be entirely different and much more complex.

pyproject.toml
```diff
@@ -69,6 +69,9 @@ inline-quotes = "single"
 '*migration*' = [
     'N999', # allow invalid module names
 ]
+'*clickhouse*' = [
+    'S608' # unsafe sql
```
bpblanken, line 73: clickhouse_driver doesn't support parameterization for table names, so those are handled with f-strings.

hanars (approved):

PR #1096, opened 2025-05-06
feat: alphabetize nested field

bpblanken:


v03_pipeline/lib/tasks/exports/misc_test.py
```diff
@@ -520,7 +539,25 @@ def test_camelcase_array_structexpression_fields(self) -> None:
                             existingInframeOorfs=None,
                             existingOutofframeOorfs=None,
                             existingUorfs=None,
-                            fiveutrAnnotation=None,
+                            fiveutrAnnotation=hl.Struct(
```
bpblanken, line 542: both of these tests are "before" the sort step, so they are not testing the functionality.  There's a unit test of the function in this file below, and also test of the actual export when writing transcripts.

hanars (approved):

PR #1099, opened 2025-05-13
refactor: improve code sharing for clickhouse migration tasks

bpblanken:


v03_pipeline/lib/model/dataset_type.py
```diff
@@ -101,6 +101,24 @@ def entries_fields(
...
+        self,
+    ) -> Callable[[hl.StructExpression], hl.StructExpression]:
+        return {
+            DatasetType.SNV_INDEL: lambda fe: hl.Struct(
+                sampleId=fe.s,
+                gt=hl.case()
+                .when(fe.GT.is_hom_ref(), 0)
+                .when(fe.GT.is_het(), 1)
+                .when(fe.GT.is_hom_var(), 2)
+                .default(hl.missing(hl.tint32)),
```
hanars, line 115: I realize this is a copy-paste, but you can just use `gt=fe.GT.n_alt_alleles()` here

hanars (approved):

v03_pipeline/lib/model/dataset_type.py
bpblanken: yes that's better!

PR #1115, opened 2025-06-11
feat: refactor transcript export to exclude MITO/GCNV/SV

bpblanken:


v03_pipeline/lib/tasks/exports/fields.py
```diff
@@ -256,11 +254,11 @@ def get_populations_export_fields(ht: hl.Table, dataset_type: DatasetType):
         },
         DatasetType.GCNV: lambda ht: {
             'seqrPop': hl.Struct(
-                af=ht.gt_stats.AF,
                 ac=ht.gt_stats.AC,
+                af=ht.gt_stats.AF,
```
bpblanken, line 258: these fields were not correctly ordered.

PR #1120, opened 2025-06-26
null out alternate contigs in lifted over coords

bpblanken:


v03_pipeline/lib/tasks/exports/fields.py
```diff
@@ -321,6 +321,12 @@ def get_variant_id_fields(
 def get_lifted_over_position_fields(ht: hl.Table, dataset_type: DatasetType):
     if dataset_type == DatasetType.MITO:
         return {'liftedOverPos': ht.rg37_locus.position}
+    # Null out alterate contigs.
+    if ht.rg37_locus.contig not in ReferenceGenome.GRCh37.standard_contigs:
```
bpblanken, line 325: An alternative here would be to add an `ALT` value to the enum in ClickHouse, but I don't think that these have much value?

hanars (approved):

PR #1129, opened 2025-07-16
chore: remove clinvar from clickhouse export

bpblanken:


v03_pipeline/lib/tasks/exports/misc.py
```diff
@@ -87,6 +87,15 @@ def export_parquet_filterable_transcripts_fields(
     return OrderedDict(sorted(fields.items()))
 
 
+def drop_unexported_fields(ht: hl.Table) -> hl.Table:
```
bpblanken, line 90: I opted to add a method here that just drops clinvar as the first thing when building the exports.  To me, this is simpler than continuing to support `clinvar` being part of the enum re-format.

bpblanken: eh, this was wrong.  It's causing more problems downstream, so I will remove it.

bpblanken: edit:  I misspelled `enums` as `nums` which caused the issue.  it's fine to stay.

hanars (approved):

PR #1136, opened 2025-08-05
Convert the v03 pipeline queue to a real queue

nvnieuwk:
This PR implements actual queuing into the v03 pipeline by expanding the current system. 

<b>Main changes</b>
1. Whenever a valid POST request is being sent to the `loading_pipeline_enqueue` route, a new queue file is created with a name structured as follows: `request_<uuid>.json` where `<uuid>` is a unique identifier to prevent any filename collisions
2. The pipeline worker will check if the queue directory is empty, when not empty, the worker will take the oldest queue file and run the job corresponding to that file. When the job is done, the file gets deleted and the next one is started


v03_pipeline/api/app_test.py
```diff
@@ -92,6 +92,7 @@ async def test_loading_pipeline_enqueue(self):
                     'projects_to_run': ['project_a'],
                     'reference_genome': 'GRCh38',
                     'sample_type': 'WGS',
+                    'skip_check_sex_and_relatedness': False,
```
bpblanken, line 95: Was this just a stray line?

v03_pipeline/lib/paths.py
```diff
@@ -1,6 +1,8 @@
 import hashlib
 import os
 
+from uuid import uuid1
```
bpblanken, line 4: I think it might actually be better to use the request_id (currently a timestamp) being assigned by the [pipeline_worker](https://github.com/broadinstitute/seqr-loading-pipelines/blob/main/v03_pipeline/bin/pipeline_worker.py#L37-L39) to instead be passed in to `loading_pipeline_queue_path()`.  That value is lexically sortable, for example:

```
20250805-123456
20250805-223000
20250806-001500
```

v03_pipeline/lib/paths.py
```diff
@@ -450,8 +452,18 @@ def loading_pipeline_queue_path() -> str:
     return os.path.join(
         LOCAL_DISK_MOUNT_PATH,
         'loading_pipeline_queue',
-        'request.json',
+        f'request_{uuid1().int}.json',
+    )
+
+def get_oldest_queue_path() -> str:
```
bpblanken, line 458: This method might actually fit nicely in [runs.py](https://github.com/broadinstitute/seqr-loading-pipelines/blob/40ae0ce35fdbf141ae69aba9a944d87c9f29516c/v03_pipeline/lib/misc/runs.py#L4).  You'll probably want to abstract the 

```
os.path.join(LOCAL_DISK_MOUNT_PATH,
         'loading_pipeline_queue')
```
into a helper in some way that would be shared between `loading_pipeline_queue_path(run_id: str)` and this method!

v03_pipeline/api/app.py
```diff
@@ -35,24 +35,6 @@ async def loading_pipeline_enqueue(request: web.Request) -> web.Response:
     except ValueError as e:
         raise web.HTTPBadRequest from e
 
-    try:
```
bpblanken, line 38: we don't need to do this on this pr, but a back pressure notion of "there are too many files in the queue" would make this more complete!  I can add a ticket.

v03_pipeline/api/app_test.py
nvnieuwk: The tests were failing because that line was missing. No idea why but it didn't seem related to my changes.

v03_pipeline/lib/paths.py
nvnieuwk: Good idea! I'll give it a go

v03_pipeline/api/app.py
nvnieuwk: Good idea, what would be 'too many files' in this case?

v03_pipeline/lib/paths.py
nvnieuwk: I've ultimately decided to do a combination of both methods. A queue file would look something like this: `request_20250805-123456_<uuid>.json`. I chose to keep the uuid in there to be prevent losses in jobs when multiple jobs are submitted within the same second

v03_pipeline/lib/paths.py
nvnieuwk: I moved the method to `runs.py` :)

v03_pipeline/api/app_test.py
bpblanken: Interesting, I will take a look into this as it is unexpected!

v03_pipeline/api/app.py
nvnieuwk: We would probably want to queue more in our case. Would it be bad if the limit is higher? (I'm thinking 1000 even)

v03_pipeline/lib/paths.py
```diff
@@ -445,14 +447,26 @@ def project_pedigree_path(
...
+def loading_pipeline_queue_path() -> str:
+    """
+    Returns a new path for a loading pipeline queue request file.
+    """
+    run_id = datetime.datetime.now(datetime.UTC).strftime(
+        '%Y%m%d-%H%M%S',
+    )
+    return os.path.join(
+        loading_pipeline_queue_dir(),
+        f'request_{run_id}_{uuid.uuid1().int}.json',
```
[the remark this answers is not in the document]
bpblanken: Ah, I see that you're not including it when you regex parse downstream.  I think if you shorten the randomness you should be fine to include it as part of the run id.

v03_pipeline/api/app.py
nvnieuwk: That's a good idea, thank you! We sadly can't do analysis on large cohorts at our lab due to several reasons, but having a way to configure that limit might be the best way to handle this.

Also really looking forward to the concurrent loads!

nvnieuwk: I added the `LOADING_QUEUE_LIMIT` environment variable with a default of `10` for this. The app will first check if the queue is full before adding a file to it and will return a `409` error in this case.

bpblanken (approved):
@nvnieuwk thanks for incorporating the feedback!  looks like I might need to move a few imports around to get the build passing but things looks fully functional and correct to me!

nvnieuwk:
Awesome! I can also take a look at it tomorrow if you haven't done it by then :)

PR #1139, opened 2025-08-06
feat: deduplicate with a deterministic algorithm 

bpblanken:


v03_pipeline/lib/tasks/exports/write_new_entries_parquet.py
```diff
@@ -71,6 +72,7 @@ def create_table(self) -> None:
                 ),
             )
             ht = deglobalize_ids(ht)
+            ht = deduplicate_by_most_non_ref_calls(ht)
```
bpblanken, line 75: I added this to both the production export and the migration.

bpblanken: Also noting that this'll run for ALL dataset types.  Not an issue for SNV_INDEL (except for the known AnVIL projects), but will impact GCNV/MITO too, likely in a positive way.

hanars (approved):

PR #1140, opened 2025-08-07
feat: modify annotations load to support lack of `annotations_disk`, improve performance of `LEFT ANTI JOIN`, remove staging dictionary as RELOAD is atomic.

bpblanken:


v03_pipeline/lib/misc/clickhouse.py
```diff
@@ -531,18 +519,32 @@ def direct_insert_new_keys(
...
+            SELECT {clickhouse_table.select_fields}
+            FROM {src_table} WHERE {clickhouse_table.key_field} IN {table_name_builder.staging_dst_prefix}/_tmp_loadable_keys`
+            """,
+        )
     logged_query(
         f"""
         INSERT INTO {dst_table}
-        SELECT {clickhouse_table.select_fields}
-        FROM {src_table} WHERE {clickhouse_table.key_field} IN {table_name_builder.staging_dst_prefix}/_tmp_loadable_keys`
+        SELECT {ClickHouseTable.ANNOTATIONS_MEMORY.select_fields}
```
bpblanken, line 546: the idea here is to use the `annotations_memory` table for the join that determines new variants for both `annotations_disk` and `annotations_memory` inserts.  We will always write to disk first, so in the case of the load failing halfway through, with some variants on disk but not in memory, we will use the safest and least-loaded table.

hanars (approved):

PR #1148, opened 2025-09-03
Benb/remove project pedigree paths arg

bpblanken:


v03_pipeline/lib/tasks/exports/write_new_entries_parquet_test.py
```diff
@@ -77,14 +80,27 @@ def setUp(self) -> None:
         )
 
     def test_write_new_entries_parquet(self):
+        copy_project_pedigree_to_mocked_dir(
```
bpblanken, line 83: this refactor was a bit more painful than necessary, pointing to a need to better align the test pedigrees with their actual function.

PR #1172, opened 2025-10-10
feat: add task to write annotations parquet

bpblanken:
Adds an export of the full annotations hail table to an annotations parquet.  We will need this to support field migrations on the rocskdb tables.


bpblanken:
Per convo, we should only do this a single time when as many fields as possible are migrated off rocksdbs.

PR #1176, opened 2025-10-20
feat: support alterate partitioning

bpblanken:


v03_pipeline/lib/misc/clickhouse.py
```diff
@@ -354,10 +406,25 @@ def delete_existing_families_from_staging_entries(
...
+        )
+    ]
+    common = [c for c in dst_cols if c in src_cols]
+    dst_list = ', '.join(common)
+    src_list = ', '.join(common)
     logged_query(
         f"""
-        INSERT INTO {table_name_builder.staging_dst_table(ClickHouseTable.ENTRIES)}
-        SELECT *
+        INSERT INTO {table_name_builder.staging_dst_table(ClickHouseTable.ENTRIES)} ({dst_list})
```
bpblanken: `INSERT INTO SELECT *` fails with columns misalignment.  We really want `INSERT EXCEPT (...)...` which isn't supported though `SELECT EXCEPT` is.

v03_pipeline/lib/misc/clickhouse_test.py
```diff
@@ -1152,3 +1160,56 @@ def test_rebuild_gt_stats(self):
             """,
         )
         self.assertCountEqual(gt_stats, [(1, 0)])
+
+    def test_repartitioned_entries_table(self):
```
bpblanken: the existing tests cover the existing partitioning structure.

v03_pipeline/lib/misc/clickhouse.py
```diff
@@ -289,6 +290,53 @@ def get_create_mv_statements(
     )[0]
 
 
+def normalize_partition(partition: str) -> tuple:
```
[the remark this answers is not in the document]
hanars: if you do `SELECT tuple(partition)` does that force it to be consistent, or no?

bpblanken:
this is actually safe to go in without the seqr side, it will just be a no-op.  I will release and go validate in dev.

PR #1182, opened 2025-10-28
bugfix: use materialized view in aggregation

bpblanken:
change the aggregation approach to handle OOM when inserting into the project_gt_stats.  Previously was using a select, this modifies to instead use the incremental materialized view in the staging environment.


bpblanken:
I just tested this.  Though it works, it's too slow (estimated 3 hours for our big project).  I think iterating over project partitions makes sense, so will incorporate that logic.

PR #1189, opened 2025-11-04
optimization: "parallel with" within optimize table

bpblanken:
This is intended to go out alongside [these setting changes](https://github.com/broadinstitute/tgg-helm/pull/215/files).  I noticed during testing that `OPTIMIZE TABLE` queues up the partitions sequentially, which is not ideal.  We want parallelism, but do not want to consume all cores.

### DIFF
--- a/v03_pipeline/lib/misc/clickhouse_test.py
+++ b/v03_pipeline/lib/misc/clickhouse_test.py
@@ -618,6 +618,7 @@ class ClickhouseTest(MockedDatarootTestCase):
         insert_new_entries(table_name_builder)
         optimize_entries(
             table_name_builder,
+            ['project_a', 'project_b', 'project_c'],
         )
         staged_project_gt_stats = client.execute(
             f"""


v03_pipeline/lib/misc/clickhouse.py
```diff
@@ -466,10 +467,22 @@ def optimize_entries(
...
+                    staging=True,
+                )
+                table_name = table_name_builder.staging_dst_table(
+                    ClickHouseTable.ENTRIES,
+                )
+                optimize_statements = [
+                    f'OPTIMIZE TABLE {table_name} PARTITION {partition} FINAL'
+                    for partition in partitions
+                ]
+                parallel_optimize_sql = '\nPARALLEL WITH\n'.join(optimize_statements)
```
bpblanken: [PARALLEL WITH](https://clickhouse.com/docs/sql-reference/statements/parallel_with)

PR #1200, opened 2025-11-26
feat: allow skipping individual validation steps

bpblanken:


v03_pipeline/lib/tasks/validate_callset.py
```diff
@@ -31,6 +28,22 @@
 
 @luigi.util.inherits(BaseLoadingRunParams)
 class ValidateCallsetTask(BaseUpdateTask):
+    @property
+    def validation_dependencies(self) -> dict[str, hl.Table]:
+        deps = {}
+        if (
+            ALL_VALIDATIONS not in self.validations_to_skip
```
bpblanken: There's a bit of weirdness here with checking ALL_VALIDATIONS.  I set up a conversion mechanism from `ALL_VALIDATIONS` -> the list of validations that runs _before_ we enter the luigi portion of the pipeline, so this additional check is only to handle the tests.

PR #1230, opened 2026-04-15
chore: use autozone

bpblanken:
us-central1-b is struggling, let gcp choose for us!


hanars:
All the buckets used by the loading pipeline are in us-central1 so my understanding it that running the pipeline in any other region makes it cost a lot more for ingress and egress - is that not the case?

PR #1235, opened 2026-05-06
feat: update to pytest, add CLAUDE.md, variant annotations task rename (just for appearances)

### LINKED ISSUE #600: Migrate from `nose` to a different unittest provider
the new v03 pipeline is using `nose` because that's what the v02 pipeline was using.  However, it's not being maintained: 

```
Nose has been in maintenance mode for the past several years and will likely cease without a new person/team to take over maintainership. New projects should consider using [Nose2](https://github.com/nose-devs/nose2), [py.test](http://pytest.org/), or just plain unittest/unittest2.
```

and fell over itself when pushing to `python 3.10`.

I've historically used `py.test` but a discussion and/or evaluation of other options is probably warranted.

bpblanken:
Resolves https://github.com/broadinstitute/seqr-loading-pipelines/issues/600

NOTE: I only migrated one test file away from `unittest.TestCase`, the unit test structure works as is.


hanars:
One high level concern I have here (that I did not know before reading this PR) is that nose actually has a much more similar test structure to django test cases than pytest. If the goal is to move this repo into the main seqr repo and unify test db setup with django than actually the migration to pytest in this PR will make more work for us to basically undo the pytest change to if we do want to incorporate the django test runner later

bpblanken:
I think we can leave the test structure as is if necessary; `pytest` can slide in as a runner over the existing `unitest` structure and we can take advantage of the nicer output and non-eol library.

hanars (approved):

PR #1236, opened 2026-05-08
run sample qc on loading families only

### LINKED ISSUE #1068: Run sample_qc only on loading families.
NOTE: this is not just filtering the sample_qc output run over the entire callset to loading samples, but running sample qc over a merged MT of all remapped_and_subsetted callsets.

bpblanken:
Resolve https://github.com/broadinstitute/seqr-loading-pipelines/issues/1068

### DIFF
--- a/loading_pipeline/lib/tasks/write_remapped_and_subsetted_callset.py
+++ b/loading_pipeline/lib/tasks/write_remapped_and_subsetted_callset.py
@@ -197,11 +197,6 @@ class WriteRemappedAndSubsettedCallsetTask(BaseWriteTask):
                 key='s',
             ),
         )
-        # Drop additional fields imported onto the intermediate callsets but
-        # not used when creating the downstream optimized tables.
-        for field in mt.row_value:
-            if field not in self.dataset_type.row_fields:
-                mt = mt.drop(field)
 
         if self.dataset_type.overwrite_male_non_par_calls:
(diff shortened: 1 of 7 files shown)


loading_pipeline/lib/tasks/write_remapped_and_subsetted_callset.py
```diff
@@ -209,11 +197,6 @@ def create_table(self) -> hl.MatrixTable:
                 key='s',
             ),
         )
-        # Drop additional fields imported onto the intermediate callsets but
```
bpblanken, line 212: Using the subsetted call sets for sample_qc means this snippet needs to be removed; `info.AF` is required there.  

Coming up with a solution for what to do without a callset annotated AF will be a blocker for running sample_qc on external call sets.

Generally, we probably shouldn't be using a callset AF anyway... so the future likely requires bridging the clickhouse allele frequencies with the call sets... which is doable but far in the future.

hanars: annotating a callset AF on an MT in hail is pretty trivially easy, if we did want to run sample qc on external data without needing a full rearchitecting. This also impacts our ability to load groups of gVCFs. Agrred that in th elong run we do not want to rely on callset AF though

hanars (approved):

PR #1239, opened 2026-06-26
Local install bugfix

hanars:
## Pull request overview

Fixes local-install edge cases by tightening task completion checks for Hail outputs and avoiding unnecessary ClickHouse rebuild work when datasets are empty.

**Changes:**
- Update `WriteRemappedAndSubsettedCallsetTask.complete()` to require non-empty `family_samples` globals in the output MatrixTable.
- Update `rebuild_gt_stats()` to short-circuit when `gt_stats` is empty by checking `max(key)` up front.

### DIFF
--- a/loading_pipeline/lib/misc/clickhouse.py
+++ b/loading_pipeline/lib/misc/clickhouse.py
@@ -1113,13 +1113,22 @@ def rebuild_gt_stats(
         msg = f'Skipping gt stats rebuild for {reference_genome.value}/{dataset_type.value} {project_guids[:10]}...'
         logger.info(msg)
         return
-    msg = f'Attempting rebuild gt stats for {reference_genome.value}/{dataset_type.value} {project_guids[:10]}...'
-    logger.info(msg)
     table_name_builder = TableNameBuilder(
         reference_genome,
         dataset_type,
         run_id,
     )
+    max_key = logged_query(
+        f"""
+        SELECT max(key) FROM {table_name_builder.dst_table(ClickHouseTable.GT_STATS)}
+        """,
+    )[0][0]
+    if not max_key:
+        msg = f'Skipping gt stats rebuild for empty dataset {reference_genome.value}/{dataset_type.value} {project_guids[:10]}...'
+        logger.info(msg)
+        return
+    msg = f'Attempting rebuild gt stats for {reference_genome.value}/{dataset_type.value} {project_guids[:10]}...'
+    logger.info(msg)
     drop_staging_db()
     create_staging_tables(
         table_name_builder,
@@ -1160,11 +1169,6 @@ def rebuild_gt_stats(
         table_name_builder.staging_dst_prefix,
     )
     # NB: encountered OOMs with large projects, necessitating sharding the insertion query.
-    max_key = logged_query(
-        f"""
-        SELECT max(key) FROM {table_name_builder.dst_table(ClickHouseTable.GT_STATS)}
-        """,
-    )[0][0]
     step = math.ceil(max_key / 5)
     for range_start in range(0, max_key, step):
         range_end = min(range_start + step, max_key + 1)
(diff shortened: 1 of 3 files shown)


loading_pipeline/lib/misc/clickhouse.py
```diff
@@ -1113,13 +1113,22 @@ def rebuild_gt_stats(
...
     )
+    max_key = logged_query(
+        f"""
+        SELECT max(key) FROM {table_name_builder.dst_table(ClickHouseTable.GT_STATS)}
+        """,
+    )[0][0]
+    if not max_key:
+        msg = f'Skipping gt stats rebuild for empty dataset {reference_genome.value}/{dataset_type.value} {project_guids[:10]}...'
+        logger.info(msg)
+        return
```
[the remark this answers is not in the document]
hanars: testing this is massively out of scope for this PR - it would require huge changes to how we set up the database in tests. This is a casualty of us not moving this code into the seqr repo with shared DB set up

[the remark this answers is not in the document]
hanars: AI seems really confused by the fact that clickhouse is 1 indexed and there is no legitimate case where there is and data in the table with a key of 0. While there may be badly constructed unit tests that corrupt the data, in the real world the first key will always be 1

loading_pipeline/lib/tasks/write_remapped_and_subsetted_callset.py
```diff
@@ -53,8 +53,13 @@ class WriteRemappedAndSubsettedCallsetTask(BaseWriteTask):
...
     def complete(self) -> luigi.Target:
-        return super().complete() and hl.eval(
-            hl.read_matrix_table(self.output().path).globals.remap_pedigree_hash
+        if not super().complete():
+            return False
+        mt = hl.read_matrix_table(self.output().path)
+        return bool(
+            hl.eval(mt.globals.family_samples),
+        ) and hl.eval(
+            mt.globals.remap_pedigree_hash
```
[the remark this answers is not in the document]
hanars: this is true, but all other approaches I took to this did not end up evaluating properly, so the slight performance hit is worth the trade off for accuracy

loading_pipeline/lib/misc/clickhouse.py
```diff
@@ -1113,13 +1113,22 @@ def rebuild_gt_stats(
...
         reference_genome,
         dataset_type,
         run_id,
     )
+    max_key = logged_query(
+        f"""
+        SELECT max(key) FROM {table_name_builder.dst_table(ClickHouseTable.GT_STATS)}
+        """,
+    )[0][0]
+    if not max_key:
```
larrybabb, line 1126: If the GT_STATS table has rows where max(key) legitimately equals 0, this would incorrectly skip the rebuild. Whether key can be 0 depends on the schema — worth verifying that key is always >= 1.

loading_pipeline/lib/tasks/write_remapped_and_subsetted_callset.py
```diff
@@ -53,8 +53,13 @@ class WriteRemappedAndSubsettedCallsetTask(BaseWriteTask):
     project_i = luigi.IntParameter()
 
     def complete(self) -> luigi.Target:
-        return super().complete() and hl.eval(
-            hl.read_matrix_table(self.output().path).globals.remap_pedigree_hash
+        if not super().complete():
+            return False
```
larrybabb, line 57: The function return type annotation says luigi.Target but returns bool. The existing code already had this issue (and chains return the falsy value or the last truthy value), but the new code makes it worse by explicitly returning `False`. The type annotation is misleading - unless there's some undocumented reason for this).  Minor, but this may be worth correcting since it's being touched.

larrybabb (changes requested):
2 items

loading_pipeline/lib/misc/clickhouse.py
hanars: I verified this previously

loading_pipeline/lib/tasks/write_remapped_and_subsetted_callset.py
hanars: Fixing the type here causes a cascade of other upstream type issues, and as you note it was already an issue here so I think fixing it is out of scope for this PR

larrybabb (approved):

PR #1240, opened 2026-08-06
bugfix: safety check for empty families list, maximum retries in the …

bpblanken:
…optimize


hanars (approved):

hanars:
This should fix https://github.com/broadinstitute/seqr/pull/5534 (although will probably not fix whatever underlying reasons we are ending up in corrupted family samples state in the first place)

hanars (approved):

PR #1242, opened 2026-08-06
typo

bpblanken:


loading_pipeline/lib/paths.py
```diff
@@ -128,7 +128,7 @@ def postprocessed_callset_path(
             reference_genome,
             dataset_type,
         ),
-        'posprocessed_callsets',
+        'postprocessed_callsets',
```
bpblanken, line 131: I just noticed this while looking at gcs 🤦.

hanars (approved):

